Issue #98, opened 2021-10-26
should StringValue nodes contain the quotes?


Geal:
## Description

I see that when parsing a string value, the node's span contains the double quotes. Example here: https://github.com/apollographql/apollo-rs/blob/main/crates/apollo-parser/src/parser/grammar/value.rs#L208

So when manipualting it, if I want to access the text, I need to remove them. It is not a big issue, but I was wondering if that is something that could be done at the API level (a method of StringValue removing the quotes) or at the parsing level (considering the quotes bound the string token but are not part of it).

Geal:
in which PR (and version) was it fixed? I'll check if there's still some code in the router remaining with quotes handling

lrlna:
oops, closed without submitting a comment! apologies

we implemented a `Into<String> for StringValue` a while back as part of #111 and published way before we started doing a changelog (so pre 0.2.0). you will already have this in the router.

here's how you'd use it:

``` rust
if let ast::Value::StringValue(val) = argument.value().expect("Cannot get argument value.") {
    let val: String = val.into(); // trims `"`
}
```

PR #111, opened 2021-11-15
Into<T> for StringValue, IntValue, FloatValue, BooleanValue

### LINKED ISSUE #110: implement Into<T> methods for VALUE's StringValue, FloatValue, IntValue, and Boolean
## Description

It would be useful for VALUE node to have `Into<T>` implementation to Rust's types for StringValue, FloatValue, IntValue and Boolean. This would make it easier for anything Eq/PartialEq operations we have to do, and (theoretically) not interfere with any semantic analysis work we will be completing in the future. 

I'd like to see VALUEs be used like this, for example:

```rust
let schema = r#"
enum Test @dir__one(string: "string value", int_value: -10, float_value: -1.123e+4, bool: false) {
  INVENTORY
} "#;

let parser = Parser::new(schema);
let ast = parser.parse();

assert!(ast.errors.is_empty());

let document = ast.document();
for definition in document.definitions() {
    if let ast::Definition::EnumTypeDefinition(enum_) = definition {
        for directive in enum_.directives().unwrap().directives() {
            for argument in directive.arguments().unwrap().arguments() {
                let val = argument.value().expect("Cannot get argument value.");
                match val {
                    ast::Value::StringValue(val) => {
                        let s: String = val.into();
                        assert_eq!(s, "string value".to_string());
                    }
                    ast::Value::IntValue(val) => {
                        let i: i64 = val.into();
                        assert_eq!(i, -10);
                    }
                    ast::Value::FloatValue(val) => {
                        let f: f64 = val.into();
                        assert_eq!(f, -1.123e+4);
                    }
                    ast::Value::BooleanValue(val) => {
                        let b: bool = val.into();
                        assert_eq!(b, false);
                    }
                    _ => unimplemented!(),
                }
            }
        }
    }
}
```

lrlna:
This helps converting some of VALUE's values into native rust types which should make them a lot easier to be accessed.

fixes #110 

perhaps also addresses #98

@Geal @o0Ignition0o what do you think about something like this to address some of the VALUE work you're doing in the router?


o0Ignition0o:
This looks great! I'll let you know how it goes.

Issue #59, opened 2021-07-20
cargo fuzz for apollo-parser


lrlna:
Once the test case generator in #114 is written we can use it to fuzz `apollo-parser`. `wasm-tools` also have a good [reference implementation](https://github.com/bytecodealliance/wasm-tools/tree/e67010a741b3e1686ec08ca80e517f8873fd9d4e/fuzz) that we could peek at when writing this. Writing a `fuzz_target` will be the smallest part of this task, the bulk of the work here would be to fix any bugs that we find when fuzzing.

garypen:
I enabled some simple fuzz testing on the parser. I used a very small corpus of 1 schema and limited the input to alphanumeric:

schema
```
type Query {
  getAllBlogs(): [BlogPost]
  getBlogPost(id: ID!): BlogPost
}

type Mutation {
  addBlogPost(title: String!, content: String!, authorID: ID!): BlogPost!
}

type BlogPost {
  id: ID!
  title: String!
  content: String!
  hasAuthor: Author!
}

type Author {
  id: ID!
  name: String!
  hasBlog: [BlogPost]
}
```

Tiny code change:

```
use apollo_parser::Parser;

 fuzz_target!(|data: &[u8]| {
    if let Ok(v) = std::str::from_utf8(data) {
        if v.chars().all(|x| x.is_alphanumeric()) {
            let _ = Parser::new(v);
        }
    }
 });
```
Fuzzing is limited to +nightly, but it seems to be working fine. So far, it's been running for about 30 minutes and no issues found!

Geal:
I think both the fuzzer and the test case generator are useful. The test case generator makes valid schemas and queries that we should handle correctly, while the fuzzer makes sure the parse is robust against the raw garbage that we can receive from network requests

lrlna:
> I think both the fuzzer and the test case generator are useful. The test case generator makes valid schemas and queries that we should handle correctly, while the fuzzer makes sure the parse is robust against the raw garbage that we can receive from network requests

yep, that's why we want a fuzzer :)

lrlna:
done with #168

PR #122, opened 2021-11-30
set up fuzzing for apollo-parser

Geal:
following up on https://github.com/apollographql/apollo-rs/pull/115#discussion_r759136264, I set up a fuzzing test (I see @garypen worked on it too in https://github.com/apollographql/apollo-rs/issues/59#issuecomment-960620361 ) and it is already triggering issues in the lexer, that should be easy to address

### DIFF
--- a/crates/apollo-parser/src/parser/grammar/extensions.rs
+++ b/crates/apollo-parser/src/parser/grammar/extensions.rs
@@ -14,6 +14,6 @@ pub(crate) fn extensions(p: &mut Parser) {
         Some("union") => union_::union_type_extension(p),
         Some("enum") => enum_::enum_type_extension(p),
         Some("input") => input::input_object_type_extension(p),
-        _ => p.err("A Type System Extension cannot be applied"),
+        _ => p.err_and_pop("A Type System Extension cannot be applied"),
     }
 }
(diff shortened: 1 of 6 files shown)


Geal:
the interesting thing is that it can find small issues in the lexer easily if we do not provide any existing query in the corpus. In those cases, a fuzzer just tests things and slowly learns how to make a valid file

Geal:
I added some of those issues and their fixes. The fuzzer will keep running into `unwrap()` calls in the lexer, and infinite loop in the parser. It will generate annoying bugs for a while, but it's an easy way to make the project more robust

crates/apollo-parser/src/parser/grammar/extensions.rs
```diff
@@ -14,6 +14,6 @@ pub(crate) fn extensions(p: &mut Parser) {
         Some("union") => union_::union_type_extension(p),
         Some("enum") => enum_::enum_type_extension(p),
         Some("input") => input::input_object_type_extension(p),
-        _ => p.err("A Type System Extension cannot be applied"),
+        _ => p.err_and_pop("A Type System Extension cannot be applied"),
```
lrlna, line 17: ohhh good catch!

lrlna (approved):
Thank you, this is great!

PR #124, opened 2021-12-01
chore: add tests for unterminated strings and invalid type system extensions

lrlna:
follows up on #122 and adds tests for the incorrectly lexed and parsed inputs that fuzzing discovered.

This commit also changes logic around having an "unexpected end of data" for `""` string. This now gets lexed into a `StringValue` token.


Geal:
nice :)

PR #127, opened 2021-12-03
fix(apollo-parser): properly create TYPE's NAMED_TYPE, LIST_TYPE, NON_NULL_TYPE

### LINKED ISSUE #125: cannot cast a SyntaxNode to a Type
cc @cecton who found the issue

with the following query: 
```
query Foo($bar:Int){name}
```

This syntax tree is generated:

```
- DOCUMENT@0..27
    - OPERATION_DEFINITION@0..27
        - OPERATION_TYPE@0..6
            - query_KW@0..5 "query"
            - WHITESPACE@5..6 " "
        - NAME@6..9
            - IDENT@6..9 "Foo"
        - VARIABLE_DEFINITIONS@9..21
            - L_PAREN@9..10 "("
            - VARIABLE_DEFINITION@10..20
                - VARIABLE@10..15
                    - DOLLAR@10..11 "$"
                    - NAME@11..15
                        - IDENT@11..14 "bar"
                        - WHITESPACE@14..15 " "
                - COLON@15..16 ":"
                - WHITESPACE@16..17 " "
                - TYPE@17..20
                    - NAMED_TYPE@17..20
                        - NAME@17..20
                            - IDENT@17..20 "Int"
            - R_PAREN@20..21 ")"
        - SELECTION_SET@21..27
            - L_CURLY@21..22 "{"
            - FIELD@22..26
                - NAME@22..26
                    - IDENT@22..26 "name"
            - R_CURLY@26..27 "}"
```

if we run this: 

```rust
  let document = tree.document();

    for def in document.definitions() {
        match def {
            Definition::OperationDefinition(op) => {
                println!("op: {:?}", op);
                for variable in op.variable_definitions().unwrap().variable_definitions() {
                    println!("variable: {:?}", variable.variable());
                    println!("ty: {:?}", variable.ty());
                }
            }
            _ => {}
        }
    }
```

we get the variable name correctly, but the `ty()` call returns `None`.

It is apparently caused by the `AstNode::cast` implementation for `Type`:

https://github.com/apollographql/apollo-rs/blob/f288903fbf753e2be960b0324c2819eebddf6d55/crates/apollo-parser/src/ast/generated/nodes.rs#L2272-L2280

it checks for the `NAMED_TYPE` kind (and others), but the node's kind is actually `TYPE` (the `NAMED_TYPE` node is right below)

lrlna:
fixes #125

whenever a NAMED_TYPED, LIST_TYPE, NON_NULL_TYPE are created, they are automatically get created as part of the TYPE node, so we do not need to start it manually. This fix makes it possible to once again do:

```rust
if let ast::Type::NamedType(name) = var.ty().unwrap() {
    assert_eq!(name.name().unwrap().text().as_ref(), "Int")
}
```

### DIFF
--- a/crates/apollo-parser/src/parser/grammar/ty.rs
+++ b/crates/apollo-parser/src/parser/grammar/ty.rs
@@ -57,7 +57,6 @@ pub(crate) fn ty(p: &mut Parser) {
     fn process(types: &mut VecDeque<(SyntaxKind, Token)>, p: &mut Parser) {
         match types.pop_front() {
             Some((kind @ S!['['], token)) => {
-                let _ty_g = p.start_node(SyntaxKind::TYPE);
                 let _list_g = p.start_node(SyntaxKind::LIST_TYPE);
                 p.push_ast(kind, token);
                 process(types, p);
@@ -66,7 +65,6 @@ pub(crate) fn ty(p: &mut Parser) {
                 }
             }
             Some((kind @ SyntaxKind::NON_NULL_TYPE, _)) => {
-                let _ty_g = p.start_node(SyntaxKind::TYPE);
                 let _non_null_g = p.start_node(kind);
                 process(types, p);
             }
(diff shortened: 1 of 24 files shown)


Geal:
Great, thanks 🙂

PR #129, opened 2021-12-04
Separate "enum StringValue" to "struct StringValue"s

Danue1:
I'm glad I found apollo-rs.
Looking at the source code, it seems that it is not yet complete.
I tried to correct the part that was difficult to understand the intention.

The purpose of PR is simple.
Four StringValues grouped by enum are separated into each struct.
The reason for the separation is that each StringValue is used in a different context.

### DIFF
--- a/crates/apollo-encoder/src/directive_def.rs
+++ b/crates/apollo-encoder/src/directive_def.rs
@@ -1,6 +1,6 @@
 use std::fmt;
 
-use crate::{InputValue, StringValue};
+use crate::{InputValue, TopStringValue};
 
 /// The `__Directive` type represents a Directive that a service supports.
 ///
@@ -35,7 +35,7 @@ pub struct Directive {
     // Name must return a String.
     name: String,
     // Description may return a String or null.
-    description: StringValue,
+    description: TopStringValue,
     // Args returns a Vector of __InputValue representing the arguments this
     // directive accepts.
     args: Vec<InputValue>,
@@ -49,7 +49,7 @@ impl Directive {
     pub fn new(name: String) -> Self {
         Self {
             name,
-            description: StringValue::Top { source: None },
+            description: Default::default(),
             args: Vec::new(),
             locations: Vec::new(),
         }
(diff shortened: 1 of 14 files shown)


Danue1:
I signed the CLA.

lrlna:
hi @Danue1, thank you for taking the time to submit a PR! I must say, this change does not make a lot of sense to me. Since `StringValue` is [a singular token](https://spec.graphql.org/October2021/#StringValue) in the GraphQL spec, it makes the most sense for it to be a single item in the encoder. It's written as an enum in this case to separate out the various ways that indentation is done when encoding different parts of the grammar. If this is what was unclear, it may be better to clarify the variants' names.

Danue1:
This change makes sense.
In this code, you could "accidentally" put `StringValue::Field { value: String }` where you should have `StringValue::Top { value: String }`.
Of course, there are tests to prevent this, but if you only look at the syntax and semantics, you can think like the above.
That's why I made this PR. I hope you understand.

If you do not like this method, there is a way to modify it as follows. (I don't like the way.)

```rust
mod string_value {
   pub struct Top {
     value: String,
   }

   // ... and 3 structs
}
```

Danue1:
@lrlna 
If you understand the purpose of this PR and decide it's worth accepting, I'll track it further in [this issue](https://github.com/apollographql/apollo-rs/issues/132).
Because the changes are conflicting.

lrlna:
> In this code, you could "accidentally" put StringValue::Top { value: String } where you should have StringValue::Input { value: String }.

By the same logic users can also accidentally use `ObjectDef::new()` instead of `InputObjectDef::new()`. What may be useful here is documentation to prevent this sort of thing from happening, changing it into separate structs won't eliminate the possibility that an incorrectly indented StringValue is used. 

Something to be said about StringValue is at the moment it is only used by other Structs in apollo-encoder, so the initial example you provided cannot be implemented by users of the library, the library does that implementation. To remove this confusion, it may actually be better to make the StringValue enum `pub(crate)` as there is no real case of setting just a StringValue on its own - at the moment it must always be part of a Definition.

I am going to close this PR as at the moment this change is not suitable.

PR #157, opened 2022-02-04
Simplify fuzzing targets

MidasLamb:
By passing &str, rather than [u8], the output is also nicer (i.e.
```bash
Output of  `std::fmt::Debug`:

        "\u{0}\u{0}\"\"\u{5}{ca*a,r#")
```

### DIFF
--- a/fuzz/fuzz_targets/lexer.rs
+++ b/fuzz/fuzz_targets/lexer.rs
@@ -2,11 +2,6 @@
 use apollo_parser::Parser;
 use libfuzzer_sys::fuzz_target;
 
-fuzz_target!(|data: &[u8]| {
-    let s = match std::str::from_utf8(data) {
-        Err(_) => return,
-        Ok(s) => s,
-    };
-
-    let _parser = Parser::new(s);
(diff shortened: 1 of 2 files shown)


bnjjj:
hi @MidasLamb Thanks for your contribution, unfortunately this code is already deprecated as I'm working on it on this branch https://github.com/apollographql/apollo-rs/tree/bnj/feat_apollo_smith feel free to take a look if you want, but it's still in progress.

MidasLamb:
@bnjjj , I just checked out that branch, seems very promising! But can I suggest that maybe instead of replacing the current one with input coming straight from the fuzzer, the `generate_valid_document` becomes another fuzz target? 
I think the value from the current one is that there is almost no presumption about what the input is, while if you start generating documents, you start to embed some sort of knowledge in there, which might cause you to never generate certain inputs which cause problems.

bnjjj:
Thanks @MidasLamb for your help. Will consider it in the future but for now we're very focus on our roadmap to release what we need for apollo-router and rover. Will let you know when we are considering external contributions. Have a nice day

PR #158, opened 2022-02-04
Replace `p.err` with `p.err_and_pop` to prevent infinite loop.

MidasLamb:
`operation_definition` is called in a loop where the we never progress
in which tokens we're looking at.
Pop while reporting the error such that progress can be made.


lrlna:
Thank you, this is great!

PR #162, opened 2022-02-07
apollo-encoder: align encoder closer to graphql specs #161

### LINKED ISSUE #161: apollo-encoder: align encoder closer to graphql spec
## Description

Currently `apollo-encoder` supports creation of any GraphQL Type System definitions. It was primarily written to allow Rover to encode introspection response into a valid SDL. To ensure that `apollo-encoder` can be used for any definition in the spec, we need to adjust and add a series of definitions. This will especially be useful to support normalisation work that Apollo Router is doing.

## Implementation
The following definitions need to be added:
- [x] Operation Definition
- [x] Fragment Definition
- [x] Type System Extensions
    - [x] Schema Extension
    - [x] Scalar Extension
    - [x] Interface Extension
    - [x] Union Extension
    - [x] Enum Extension
    - [x] Input Object Extension

The following definitions need to be adjusted
- [x] Input Field and Field are two different types in the GraphQL type system and they need to be separated
- [x] Input Value Definitions can accept any type of directive, not just `deprecated`
- [x] Schema Definition can accept any type of directive
- [x] Scalar Definition can accept any type of directive
- [x] Object Definition can accept any type of directive
- [x] Interface Definition can accept any type of directive
- [x] Union Definition can accept any type of directive
- [x] Enum Definition can accept any type of directive
- [x] Input Object Definition can accept any type of directive

bnjjj:
close  #161


crates/apollo-encoder/src/fragment.rs
```diff
@@ -0,0 +1,341 @@
...
+        write!(f, "...{}", self.name)?;
+        for directive in &self.directives {
+            write!(f, " {}", directive)?;
+        }
+
+        Ok(())
+    }
+}
+
+/// The __inlineFragment type represents an inline fragment that could be used as a field
```
lrlna, line 123:
```suggestion
/// The __inlineFragment type represents an inline fragment in a selection set that could be used as a field
```

crates/apollo-encoder/src/fragment.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    /// Set the inline fragment's type condition.
+    pub fn type_condition(&mut self, type_condition: Option<TypeCondition>) {
+        self.type_condition = type_condition;
+    }
+
+    /// Should be used everywhere in this crate isntead of the Display implementation
+    /// Display implementation is only useful as a public api
+    pub(crate) fn format_with_indent(&self, indent_level: usize) -> String {
+        let mut text = String::new();
```
lrlna, line 185:
```suggestion
        let mut text = String::from("...");
```

crates/apollo-encoder/src/input_value.rs
```diff
@@ -1,14 +1,14 @@
 use std::fmt;
 
-use crate::{StringValue, Type_};
+use crate::{Directive, StringValue, Type_};
 
 // NOTE(@lrlna): __InputValue is also meant to be used for InputFields on an
 // InputObject. We currently do not differentiate between InputFields and
 // Fields, so this is not applied directly to InputObjects. Once we are able to
 // walk an AST to encode a schema, we will want to make sure this struct is used
 // directly: InputObject --> InputField --> InputValue
```
lrlna, line 9: you've made fixes that make this note irrelevant! 🎉 
```suggestion
```

crates/apollo-encoder/src/interface_def.rs
```diff
@@ -134,7 +158,7 @@ impl fmt::Display for InterfaceDef {
 #[cfg(test)]
 mod tests {
```
lrlna, line 159: Let's also add a test with an extension here.

crates/apollo-encoder/src/input_object_def.rs
```diff
@@ -97,7 +122,7 @@ impl fmt::Display for InputObjectDef {
 #[cfg(test)]
 mod tests {
```
lrlna, line 123: Let's also add a test with an input object definition extension.

crates/apollo-encoder/src/enum_def.rs
```diff
@@ -84,6 +112,8 @@ impl fmt::Display for EnumDef {
 
 #[cfg(test)]
 mod tests {
+    use crate::{Argument, Value};
```
lrlna, line 115: let's add a test with an enum definition extension

crates/apollo-encoder/src/enum_value.rs
```diff
@@ -1,6 +1,6 @@
 use std::fmt;
 
-use crate::StringValue;
+use crate::{Directive, StringValue};
```
lrlna, line 3: nit: I saw you grouped Field Definition and Field in the same file, we can probably do the same for Enum Definition and Enum Value.

crates/apollo-encoder/src/operation.rs
```diff
@@ -0,0 +1,257 @@
...
+                query($variable_def: [Int]) @testDirective(first: "one") {
+                  first
+                  second
+                }
+            "#}
+        );
+    }
+
+    #[test]
+    fn it_encodes_a_deeper_query_operation() {
```
lrlna, line 197: these are great 🙏 !

crates/apollo-encoder/src/operation.rs
```diff
@@ -0,0 +1,257 @@
...
+    variable_definitions: Vec<VariableDef>,
+    directives: Vec<Directive>,
+    selection_set: SelectionSet,
+}
+
+impl OperationDef {
+    /// Create a new instance of OperationDef
+    pub fn new(operation_type: OperationType, selection_set: SelectionSet) -> Self {
+        Self {
+            operation_type,
```
lrlna, line 67: How do you handle query shorthand encoding that doesn't have the "query" keyword, but is indeed of `Query` operation type?

crates/apollo-encoder/src/schema.rs
```diff
@@ -33,6 +33,7 @@ use crate::{
 /// );
 /// ```
 #[derive(Debug)]
+// TODO delete in profit of Document
```
lrlna, line 36:
```suggestion
// TODO delete in favour of Document
```

lrlna: And yea, let's do this as a second pass after we merge this PR.

crates/apollo-encoder/src/schema_def.rs
```diff
@@ -127,4 +154,24 @@ mod tests {
         "#}
         );
     }
+
+    #[test]
```
lrlna, line 158: let's also add an extension test

crates/apollo-encoder/src/selection_set.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// Add a selection in the SelectionSet
+    pub fn selection(&mut self, selection: Selection) {
+        self.selections.push(selection);
+    }
+
+    /// Should be used everywhere in this crate isntead of the Display implementation
+    /// Display implementation is only useful as a public api
+    pub(crate) fn format_with_indent(&self, mut indent_level: usize) -> String {
+        let mut text = String::new();
+        text.push_str("{\n");
```
lrlna, line 60:
```suggestion
        let mut text = String::from("{\n");
```

crates/apollo-encoder/src/selection_set.rs
```diff
@@ -0,0 +1,205 @@
...
+        } else {
+            text.push_str(&format!("{}}}", "  ".repeat(indent_level - 1)));
+        }
+
+        text
+    }
+}
+
+// This impl is only useful when we generate only a SelectionSet
+// If it's used from a root element, we call `format_with_indent`
```
lrlna, line 81: What do you mean by `root` here? The top-level `Document`?

crates/apollo-encoder/src/union_def.rs
```diff
@@ -58,13 +73,24 @@ impl UnionDef {
...
-        write!(f, "union {} = ", self.name)?;
+        write!(f, " =")?;
 
         for (i, member) in self.members.iter().enumerate() {
             match i {
-                0 => write!(f, "{}", member)?,
+                0 => write!(f, " {}", member)?,
                 _ => write!(f, " | {}", member)?,
             }
         }
```
lrlna, line 96: let's have an extension test here as well

crates/apollo-encoder/src/value.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+impl fmt::Display for Value {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Variable(v) => write!(f, "${v}"),
+            Self::Int(i) => write!(f, "{i}"),
+            Self::Float(fl) => write!(f, "{fl}"),
+            Self::String(s) => {
+                if s.contains('"') {
```
lrlna, line 38: it's also a block string character if it contains `\n` or `\r`
```suggestion
                if s.contains('"') | s.contains('\n') | s.contains('\r') {
```

lrlna (changes requested):
This is looking super good! Excited to have this published soon. 

The biggest thing I don't see resolved yet is the using query shorthand for an operation definition. Maybe I missed the code for it?

crates/apollo-encoder/src/selection_set.rs
bnjjj: I should use `parent` instead

crates/apollo-encoder/src/enum_value.rs
bnjjj: We also have `field.rs` and `field_value.rs` I think it's good to keep the same. If you think it's weird then I can also delete the `field_value.rs` and put everything inside. As you wish

crates/apollo-encoder/src/operation.rs
bnjjj: It was not supported. It's fixed :)

crates/apollo-encoder/src/enum_value.rs
lrlna: ah! i missed the field.rs and field_value.rs differentiation. let's keep it separate then

crates/apollo-encoder/src/selection_set.rs
lrlna: yes `parent` makes more sense to me!

bnjjj:
I made every changes except the one with `enum_value.rs` check my answer on your comment. Also I used this https://github.com/apollographql/apollo-rs/pull/162/files#diff-c5727d7d12a744bf48d8d53af8c6d9aa98316eaa005f7e3fb4e0be813fbcdf66R1 to keep the readme up to date with our documentation and our code.

lrlna (approved):
woot woot! let's squash and merge it!

PR #182, opened 2022-02-25
fix(apollo-encoder): do not trim double quotes when needed, rename method input_object_ and type InputValueDef for consistency

bnjjj:
Bugs detected when implementing the new version of apollo encoder into Rover (https://github.com/apollographql/rover/pull/1017)


lrlna (approved):
Thank you!

PR #202, opened 2022-04-01
Introduce apollo-compiler crate: beginnings of a incremental, query-based compiler for graphql

### LINKED ISSUE #149: semantic analysis: validate operation definition in Document
## Description

Validate the operation definition and add related methods to the SourceDatabase.

Related Issues, #144, #147 

## Specification

<html>
<body>

  | Documentation | SEMANTIC PROCESSING | ERROR | WARNING | HINT | SUGGESTION | QUERIES TO BE ADDED TO SOURCEDATABASE
-- | -- | -- | -- | -- | -- | -- | --
| Operation | Defined in spec |   | <ul><li>shorthand can only be used if there are no variables or directives</li><li>shorthand cannot be used if there are multiple operations</li></ul> |  |<ul><li>show where the variable or directive is already defined </li><li>show other operations</li></ul>  |  | <ul><li>variables in an operation definition → `Vec<Variables>`</li><li>directives in operation definition → `Vec<Directives>`</li><li>operation_definitions → `Vec<OperationDefinitions>`</li></ul> |

</body>
</html>

### LINKED ISSUE #152: semantic analysis: setup source database with salsa-rs
## Description
This issue is to take care of setting up an instance of `[salsa](https://docs.rs/salsa/latest/salsa/)` in our codebase. We already know that the `SyntaxTree` returned by the parser will be the basis of the Database. Database's `queries` are for the moment informed by validation rules, and any specific data that Apollo Router needs (like specified in #109, for example).

There are a few things to figure out and tweak as we build this out:
- what will be a good way for us to set up interning with the existing `SyntaxTree` returned by the parser?
- what's the most user-friendly and ergonomic way to provide users access to the database's queries?
- what would make for a good workflow for `input` --> `parse` --> `analyze/validatge/give context`?

lrlna:
Ok, I want to merge this bit of work in so I am not sitting on a branch for weeks (well, let's be honest, months) on end. This is in no way what we are going to publish as the first version of this crate, but it will allow me to incrementally (heh get it, get it? :D) make changes a bit quicker, track issues in the actual issue tracker rather in my head and the myriads of `NOTE @lrlna`. 

A few things to finish up before merging:
- [x] fix CI
- [x] tidy up database queries
- [x] add / fix subscription validation to overarching validation
- [x] tidy up diagnostics enum a tad
- [x] a couple of tests

## Description
Beginnings of `apollo-compiler` crate, an incremental, query-based compiler built on top of [`salsa-rs`](docs.rs/salsa).

This PR uses `salsa`'s queries to create a memoised database of Executable Definitions, Fragments and Operations (for now).
Based on the context built while populating the database we will then be able to run any possible validations to produce diagnostics. This PR does some very basic variable and operation name validation. 

closes #149, closes #152


crates/apollo-compiler/src/queries/database.rs
```diff
@@ -0,0 +1,656 @@
...
+        .iter()
+        .filter_map(|sel| match sel {
+            Selection::Field(field) => Some(field.as_ref().clone()),
+            _ => None,
+        })
+        .collect();
+    Some(Arc::new(fields))
+}
+
+fn operation_inline_fragment_fields(db: &dyn SourceDatabase, id: Uuid) -> Option<Arc<Vec<Field>>> {
```
bnjjj, line 196: I wonder if instead of having this kind of long function name we could implement method like `fields(&self, &dyn SourceDatabase) -> Option<Arc<Vec<Field>>>`

BTW why should we put this into an `Arc` ? Maybe we should let the user choose, if (s)he needs to wrap the value to an `Arc` then the user could do it by (her/him)self.

Last thought, why not just returning a `Vec<Field>` instead of an `Option` ? I mean if the Vec is empty it means there is no fields ? I don't know, just some thoughts. Would be happy to discuss with you about that

lrlna: @bnjjj appreciate the effort in trying to review this! I think at this point of this work it's not super useful for me to have the API reviewed, however. It's definitely far from the final form of what users are going to be working with, or what we will be publishing. I wanted to have something rounded up and merged so I am not continuously working on the same branch. If you are keen to review, I can let you know when the crate is ready for a proper look over and the sorts of feedback I am looking for!

> I wonder if instead of having this kind of long function name we could implement method like fields(&self, &dyn SourceDatabase) -> Option<Arc<Vec<Field>>>

this function is not intended to be used directly, but is meant to be memoised in the db for various validation rules to use. and yep, there is a plan to have a .fields() function for `InlineFragment` value for better ergonomics

> BTW why should we put this into an Arc ? Maybe we should let the user choose, if they needs to wrap the value to an Arc then the user could do it by themselves.

There is a lot of things here with Arc, but the intent is for the users to never use this directly. These queries are simply here to be memoised for either our own compiler uses or for the wrapper API. This is very likely to change quite a bit next week as I'll be breaking down the database into smaller chunks, and fixing the wrapper API, so it's a bit early to discuss this fully. 

> Last thought, why not just returning a Vec<Field> instead of an Option ? I mean if the Vec is empty it means there is no fields?

Could probably return an empty vec \o/. I'll have to rewrite this function once the databases get split, so this will change also 😬

PR #207, opened 2022-04-21
Update description serialization for input value definitions

### LINKED ISSUE #205: apollo-encoder: serialisation of descriptions in InputValueDefinitions
## Description

We are currently a bit all over the place with `StringValue` and `BlockStringValue` serialisations in the encoder, especially for `InputValueDefinition` in `ArgumentsDefinition`. We should serialise similarly to [`graphl-js`](https://github.com/graphql/graphql-js/blob/b38429f4d46b17f23d15406a2ca0d1eb201864b6/src/language/printer.ts#L168-L178). Values get multilined and indented which will make the generated SDL a lot easier to read. 

I imagine that instead of inlining everything for this sort of FieldDefinition:
```graphql
picture("""size of a current image"""size: Url, visibility: Boolean): Url
```
we could instead multiline it:
```graphql
picture(
  """
  size of a current image
  """
  size: Url
  visibility: Boolean
): Url
```

lrlna:
This PR introduces clearer (more readable) serialization for descriptions used in input value definitions. If any of the input value definitions in a given field definition comes with a description, we will multiline all input value definitions. That is to say, instead of serializing arguments definition like this:
```graphql
type Foo {
  "This is a description of the \`one\` field."
  one("This is a description of the \`argument\` argument." argument: InputType!): Type
}
```
we serialize it as:
```graphql
type Foo {
  "This is a description of the \`one\` field."
  one(
    "This is a description of the \`argument\` argument."
    argument: InputType!
  ): Type
}
```
This makes it a lot more readable, especially for users with a large number of input value definitions with descriptions.

In the process this PR also introduces a few changes to make this work well:
1. all description setters for types that have a description now accepts `String` instead of `Option<String>`. It makes it a bit more straightforward from an API perspective for setters to set a non-option value. The original decision was made to accommodate Rover's use-case which is rather unique. This change allows to easier broader usage.
2. We previously used an incorrect type for FieldDefinitions' internal definition of arguments. This change aligns it closer to the GraphQL spec in that it is now represented as an ArgumentsDefinition type. As part of this addition ArgumentsDefinition struct has a few (**breaking**) changes to the API. There is now `ArgumentsDefinition::new()` that initiates with an empty vec of InputValueDefinitions; `ArgumentsDefinition::with_values(vec)` that initiates given a vector of InputValueDefinitions; as well as a setter for input values `arguments_definition.input_value(input_value)`.
3. updates `apollo-smith` to use the APIs that changed in `apollo-encoder`.

Reasons for this PR were originally raised in apollographql/rover#1088

closes #205

### DIFF
--- a/crates/apollo-encoder/README.md
+++ b/crates/apollo-encoder/README.md
@@ -103,7 +103,7 @@ let mut document = Document::new();
 
 // Create a Directive Definition.
 let mut directive_def = DirectiveDefinition::new("provideTreat".to_string());
-directive_def.description(Some("Ensures cats get treats.".to_string()));
+directive_def.description("Ensures cats get treats.".to_string());
 directive_def.location("OBJECT".to_string());
 directive_def.location("FIELD_DEFINITION".to_string());
 directive_def.location("INPUT_FIELD_DEFINITION".to_string());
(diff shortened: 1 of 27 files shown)


lrlna:
@bnjjj would love your feedback on the API change for description setters, and your thoughts on the multilining input value definitions!

PR #213, opened 2022-05-03
feat(apollo-smith): add interface definition to internal stack

bnjjj:
Added support of interface definition in the stack to fill an operation with correct fields.


lrlna (approved):
Thank you!

PR #228, opened 2022-05-17
fix(parser): add a fix for StringValue node when we start with a backslash

### LINKED ISSUE #227: lexer fails to lex string values with escaped characters
## Description

See apollographql/router#1065

### Steps to reproduce

It's difficult to reproduce in the router, but I wrote a program to demonstrate the issue in the parser.

```
fn main() {
    use apollo_parser::Parser;

    let input = r#"mutation {
            createStore(draft: {
              name: [{ locale: "en", value: "\"my store\"" }]
            }) {
              name(locale: "en")
            }
          }"#;
    let parser = Parser::new(input);
    let ast = parser.parse(); <- This will never return until memory exhausted.
}
```

### Expected result

If mutation is valid it should parse. If not, fail with error(s).

### Actual result

Never returns until memory exhausted.

### Environment

- Operating system and version: Mac OS X
- Shell (bash/zsh/powershell): zsh
- Crate version: 0.2.5

bnjjj:
close #227


crates/apollo-parser/src/parser/grammar/value.rs
```diff
@@ -294,4 +294,50 @@ query GraphQuery($graph_id: ID!, $variant: String) {
...
+        let input = r#"mutation {
+            createStore(draft: {
+              name: [{ locale: "en", value: "\"my store\"" }]
+            }) {
+              name(locale: "en")
+            }
+          }"#;
+        let parser = Parser::new(input);
+        let ast = parser.parse();
+        println!("ast ===\n{:?}", ast);
```
garypen, line 309: Remove println!()

crates/apollo-parser/src/lexer/mod.rs
```diff
@@ -414,4 +418,14 @@ mod test {
...
     }
+
+    #[test]
+    fn tests_escaped_char() {
+        let gql_1 = r#"
+        { name: "\"my store\"" }
+        "#;
+        let lexer_1 = Lexer::new(gql_1);
+        dbg!(lexer_1.tokens);
+        dbg!(lexer_1.errors);
```
garypen, line 429: Remove these `dbg!()` and make some kind of assertion....

Geal:
with this implementation, it will accept strings with a `\` and non escaped chars, right? like `"aaa \a bbb"`

crates/apollo-parser/src/parser/grammar/value.rs
```diff
@@ -339,4 +339,19 @@ query GraphQuery($graph_id: ID!, $variant: String) {
...
+            createStore(draft: {
+              name: [{ locale: "en", value: "my \a store" }]
+            }) {
+              name(locale: "en")
+            }
+          }"#;
+        let parser = Parser::new(input);
+        let ast = parser.parse();
+
+        assert!(ast.errors().next().is_none());
```
Geal, line 355: the parser should return an error because `\a` is not a valid character https://spec.graphql.org/June2018/#sec-String-Value

bnjjj: Also fixed in this PR

crates/apollo-parser/src/parser/grammar/value.rs
```diff
@@ -294,4 +294,65 @@ query GraphQuery($graph_id: ID!, $variant: String) {
...
+    }
+
+    // Commented because it's creating an infinite loop
```
bnjjj, line 343: Need to find a good solution for this problem. TLDR, when we have lexer issue we could have an infinite loop adding errors

lrlna: this test should error, since `\a` is not an allowed escaped character, correct? Does this test not hit the piece of code you wrote for an ["Unexpected escaped character"](https://github.com/apollographql/apollo-rs/pull/228/files#diff-f261e5826eed1b43cb5200eeea7df771eb9af5efdb47d49741232cba6db33225R143-R145)?

bnjjj: It fails, but the issue is that, when it fails in the lexer some assumptions that are made in the parser are wrong. For example sometimes in the parser if it encounters an error it doesn't consume the current token because this kind of error should happen if we miss a closing parenthesis for example. but when there is an error in the lexer it's too unpredictable and then the assumptions made by the parser seem wrong. For example in that case it will loop forever, add an error without consuming the current token and then create a memory overflow after a moment. It never ends. That's why I commented this example because when you already have an error in the lexer, IMHO we should early return

lrlna: the parser's design doesn't really allow for early returns, we try to parse as much as we can. 

the issue you're describing is a bit unclear to me, but the behaviour _sounds_ like we ought to be doing an `err_and_bump` instead of `err` somewhere. let me take a peek (:P) quickly in the debugger

lrlna: this is looking a lot better with this new change you just pushed @bnjjj:
```
                                    - L_BRACK@64..65 "["
                                    - OBJECT_VALUE@65..90
                                        - L_CURLY@65..66 "{"
                                        - WHITESPACE@66..67 " "
                                        - OBJECT_FIELD@67..81
                                            - NAME@67..73
                                                - IDENT@67..73 "locale"
                                            - COLON@73..74 ":"
                                            - WHITESPACE@74..75 " "
                                            - STRING_VALUE@75..81
                                                - STRING@75..79 "\"en\""
                                                - COMMA@79..80 ","
                                                - WHITESPACE@80..81 " "
                                        - OBJECT_FIELD@81..89
                                            - NAME@81..86
                                                - IDENT@81..86 "value"
                                            - COLON@86..87 ":"
                                            - WHITESPACE@87..88 " "
                                            - WHITESPACE@88..89 " "
                                        - R_CURLY@89..90 "}"
                                    - R_BRACK@90..91 "]"
                                    - WHITESPACE@91..104 "\n            "
```
and errors come out as 
```
- ERROR@88:101 "unexpected escaped character" "my \a store"
- ERROR@102:103 "expected a valid Value" }
```

There is still something up with the whitespace index around the error, but this is already heaps better!

lrlna: for those following this conversation: Benjamin and I made a decision to solve the index issue in a separate PR as it does not directly pertain to escaped characters bug this PR is solving.

PR #244, opened 2022-06-09
resource bound parsing execution (followup)

garypen:
This properly unifies the limits around recursion for both:
 - selection sets
 - fields

The tests are expanded and properly exercise the various possible outcomes with recursion limits.

Co-authored-by: Irina Shestak <shestak.irina@gmail.com>


lrlna (approved):
looking good!

PR #262, opened 2022-07-26
DNM failing test for missing types on fields in inline fragments

lennyburdette:
here's a failing test showing that field return type info isn't present inside inline fragments cc @lrlna


lrlna:
thanks so much for these, @lennyburdette! i will merge your tests with a fix for the inline fragment types

lrlna:
closing **this one** - test cases are part of the fix in #282  PR

PR #267, opened 2022-08-02
fix(lexer,parser): do not run in infinite loop when we're in an unclosed string value

### LINKED ISSUE #266: infinite loop when giving a bad input string value
## Description
When the subgraph schema is malformed or contains a typo rover crash with a timeout without any understandable errors.

### Steps to reproduce

use this schema:

```graphql
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.0",
        import: ["@key", "@external])

```

Look at the unclosed `"@external`

bnjjj:
I don't know if it's the right way to fix it, but we already have this kind of issues with value if we don't pop error so I added a parameter to pop the token if needed to not have an infinite loop again there.
close #266

### DIFF
--- a/crates/apollo-parser/src/parser/grammar/value.rs
+++ b/crates/apollo-parser/src/parser/grammar/value.rs
@@ -15,7 +15,7 @@ use crate::{
 ///     EnumValue
 ///     ListValue
 ///     ObjectValue
-pub(crate) fn value(p: &mut Parser) {
+pub(crate) fn value(p: &mut Parser, pop_on_error: bool) {
     match p.peek() {
         Some(T![$]) => variable::variable(p),
         Some(TokenKind::Int) => {
@@ -50,7 +50,22 @@ pub(crate) fn value(p: &mut Parser) {
         }
         Some(T!['[']) => list_value(p),
         Some(T!['{']) => object_value(p),
-        _ => p.err("expected a valid Value"),
+        Some(token) => {
+            let error_message = format!("expected a valid Value got {token:?}");
+            if pop_on_error {
+                p.err_and_pop(&error_message);
+            } else {
+                p.err(&error_message);
+            }
+        }
+        _ => {
+            let error_message = "expected a valid Value";
+            if pop_on_error {
+                p.err_and_pop(error_message);
+            } else {
+                p.err(error_message);
+            }
+        }
     }
 }
 /// See: https://spec.graphql.org/October2021/#EnumValue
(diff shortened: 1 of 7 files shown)


crates/apollo-parser/src/parser/grammar/value.rs
```diff
@@ -50,7 +50,22 @@ pub(crate) fn value(p: &mut Parser) {
...
         Some(T!['{']) => object_value(p),
-        _ => p.err("expected a valid Value"),
+        Some(token) => {
+            let error_message = format!("expected a valid Value got {token:?}");
+            if pop_on_error {
+                p.err_and_pop(&error_message);
+            } else {
+                p.err(&error_message);
+            }
+        }
```
lrlna, line 60: i am wondering if when we have an unclosed string `Value`, we should be getting two different errors: a lexical _and_ a syntactic error. That is: "you have an unclosed string" and "this is an unsupported value".

crates/apollo-parser/test_data/parser/err/0034_directive_with_unfinished_string_value.txt
```diff
@@ -0,0 +1,116 @@
...
+- ERROR@128:129 "expected a valid Value got LParen" (
+- ERROR@135:136 "expected a valid Value got Colon" :
+- ERROR@141:142 "expected a valid Value got RParen" )
+- ERROR@153:154 "expected }" !
+- ERROR@153:154 "expected a valid Value got Bang" !
+- ERROR@162:163 "expected a valid Value got Colon" :
+- ERROR@183:184 "expected a valid Value got Colon" :
+- ERROR@204:205 "expected a valid Value got Colon" :
+- ERROR@226:227 "expected a valid Value got Colon" :
+- ERROR@235:236 "expected a valid Value got RCurly" }
```
lrlna, line 113: something is off here, we should not be having this many "expected a valid Value" here. this should only happen for this bit of your graphql `import: ["@key", "@external]`.

crates/apollo-parser/test_data/parser/err/0034_directive_with_unfinished_string_value.txt
```diff
@@ -0,0 +1,116 @@
...
+- ERROR@237:237 "unexpected EOF" EOF
```
lrlna, line 114: we shouldn't be having this error with the input you have in test file 34.

lrlna (changes requested):
the lexer bit looks good, but we gotta think a bit more about the parser changes.

with the example schema you have, we should have two errors: `"@external` is unterminated string, and `"@external` is not a valid value. We are not technically "missing a value" in this case either - `import` already has an Value in its List, so technically we are not "missing a value" where they should have been one.

the rest of the schema is absolutely parseable, and we should not encounter any other errors.

lrlna (approved):
Thank you, this looks good now!

Issue #270, opened 2022-08-06
apollo-compiler: Expose whether an input definition has a default value available


allancalix:
## Description
Expose whether an input definition has a default value and can therefore be valid without a value set.

Add details like:
- Returns true if an input definition has a default available
- In order to perform validation against input definitions, you must know whether it's valid or not that you do not have a value set. Currently there's no way to get at that without digging into the AST itself.

Relevant PR: https://github.com/apollographql/apollo-rs/pull/271

PR #271, opened 2022-08-06
Expose whether an input definition defines a default

allancalix:
See https://github.com/apollographql/apollo-rs/issues/270 for motivation.

### DIFF
--- a/crates/apollo-compiler/src/queries/values.rs
+++ b/crates/apollo-compiler/src/queries/values.rs
@@ -1417,6 +1417,11 @@ impl InputValueDefinition {
         self.directives.as_ref()
     }
 
+    // Returns true if the input value definition defines a default value.
+    pub fn has_default_value(&self) -> bool {
+        self.default_value.is_some()
+    }
+
     // Get a reference to SyntaxNodePtr of the current HIR node.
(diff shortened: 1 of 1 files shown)


lrlna:
@allancalix thank you for submitting a PR!

Input Value Definition has a Default Value field, which is an Option. This means you can run `input_definition.default_value.is_some()` to check whether that default value is present (much like you've done in this PR). With Options my preference is to usually go through this route rather write a wrapper API around `is_some()`. This signals to the user they are working with an Option, rather than hiding this fact. Is there a particular reason for your to want this wrapper?

Additionally, providing a single wrapper for this particular Option also means that all other Options in the compiler's context must have this wrapper, which is rather unmaintainable and will result in an inconsistent API.

allancalix:
Thanks for the response. I don't have a strong preference for this over returning Option<&DefaultValue>. I'm happy to update the PR to expose the option directly.

lrlna:
@allancalix ah! So the getter was indeed missing as @bnjjj pointed out. I've added it today https://github.com/apollographql/apollo-rs/blob/main/crates/apollo-compiler/src/queries/values.rs#L1438-L1442. If having a getter solves this for you, I'll close this PR.

allancalix:
This change works just fine for what i'm doing, I appreciate you making the change! Closing this out.

PR #315, opened 2022-09-22
Expose directives from FragmentSpread types

allancalix:
There is currently no way to return the directives from a `FragmentSpread` definition, this PR adds one in the same manner that `InlineFragment` does.


lrlna (approved):
tysm!

Issue #308, opened 2022-09-21
validation: validate arguments


lrlna:
- [ ] validate argument names exist, [spec](https://spec.graphql.org/October2021/#sec-Argument-Names)
- [ ] validate arguments are unique, [spec](https://spec.graphql.org/October2021/#sec-Argument-Uniqueness)
- [ ] validate required arguments are provided, [spec](https://spec.graphql.org/October2021/#sec-Required-Arguments)

PR #317, opened 2022-09-28
Validate argument name uniqueness

goto-bus-stop:
One part of #308

It's an error to declare or provide multiple arguments by the same name, eg:
```graphql
type Query {
  things(offset: Int!, offset: Int!): [Thing]
  # ERR: duplicate argument definition: offset
}
```
```graphql
query GetThings {
  things(offset: 10, offset: 20) { id }
  # ERR: duplicate argument values: offset
}
```

I started with 1 specific case here, namely arguments in an object type field definition.
- [ ] the actual duplication check should be reused for directive arguments
- [ ] maybe also for `input` type definitions?
- [ ] similar check should apply when providing arguments to a field. as that's working with concrete values rather than types, it makes sense to have a separate check

some questions:
- `UniqueDefinition` is currently used for things that must be unique document-wide, so the error message is wrong for this case.
   image

   a `UniqueArgument` diagnostic may make sense, but UniqueDefinition is already used for a few different things, so maybe it can be made more flexible by adding another property?
- some `ast_node()` methods return a node reference, while others return an Option. what is the reason? it seems like we just always do `ast_node().unwrap()`, so it should actually not be None in practice?
- converting `text_range()` to a `SourceSpan` is a little verbose and could be done wrong (eg. by passing the wrong variable combination or calling `start()` twice instead of `start()`+`len()`). A From impl is infeasible as they are defined in different crates but a function could be nice, or an ext trait to add `.source_span()` to SyntaxNodes 😄

### DIFF
--- a/crates/apollo-compiler/src/validation/mod.rs
+++ b/crates/apollo-compiler/src/validation/mod.rs
@@ -15,6 +15,7 @@ mod object;
 // executable definitions
 mod operation;
 
+mod arguments;
 mod unused_variable;
 
 use apollo_parser::SyntaxNode;
@@ -38,6 +39,7 @@ pub trait ValidationDatabase:
     fn validate_input_object(&self) -> Vec<ApolloDiagnostic>;
     fn validate_object(&self) -> Vec<ApolloDiagnostic>;
     fn validate_operation(&self) -> Vec<ApolloDiagnostic>;
+    fn validate_arguments(&self) -> Vec<ApolloDiagnostic>;
     fn validate_unused_variable(&self) -> Vec<ApolloDiagnostic>;
 }
 
@@ -57,6 +59,7 @@ pub fn validate(db: &dyn ValidationDatabase) -> Vec<ApolloDiagnostic> {
     diagnostics.extend(db.validate_object());
     diagnostics.extend(db.validate_operation());
 
+    diagnostics.extend(db.validate_arguments());
     diagnostics.extend(db.validate_unused_variable());
 
     diagnostics
@@ -96,6 +99,10 @@ pub fn validate_object(db: &dyn ValidationDatabase) -> Vec<ApolloDiagnostic> {
 
 pub fn validate_operation(db: &dyn ValidationDatabase) -> Vec<ApolloDiagnostic> {
     operation::check(db)
+}
+
+pub fn validate_arguments(db: &dyn ValidationDatabase) -> Vec<ApolloDiagnostic> {
+    arguments::check(db)
 }
 
 pub fn validate_unused_variable(db: &dyn ValidationDatabase) -> Vec<ApolloDiagnostic> {
--- a/crates/apollo-compiler/src/validation/arguments.rs
+++ b/crates/apollo-compiler/src/validation/arguments.rs
@@ -49,7 +49,8 @@ mod test {
     fn it_fails_validation_with_duplicate_argument_names() {
         let input = r#"
 type Query {
-  method(arg: Boolean, arg: Boolean): Int
+  single(arg: Boolean): Int
+  duplicate(arg: Boolean, arg: Boolean): Int
 }
 "#;
         let ctx = ApolloCompiler::new(input);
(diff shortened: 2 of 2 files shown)


goto-bus-stop:
To check directive arguments, we basically need to walk the entire AST as directives can appear in many places 🤔 it doesn't seem great to manually look up all the places where it can appear inside the validation code. I thought maybe we could walk the Rowan tree but we need the HIR, not the AST nodes to do run the validation on. Perhaps a salsa query that does manually look up all the directive calls is the best solution for now.

goto-bus-stop:
eh I really dont want to manually find all directive calls, I'll definitely forget one spot that way. I think an AST traversal is useful anyways. I think it may need to be provided by apollo-parser, so it can walk the tree using rowan and create AST nodes on the fly?

goto-bus-stop:
^ so instead of that we can implement validation a bit differently in the compiler. There can be `validate_NODETYPE(hir_node)` queries that recurse manually. that way we essentially have a tree walk, and all validation can be cached in a granular way

goto-bus-stop:
As Salsa just works with owned values, we might want to wrap some stuff in the HIR tree in `Arc`s to reduce cloning with this new strategy. i'll not do that for now and just deep clone a bunch of things

crates/apollo-compiler/src/validation/mod.rs
```diff
@@ -39,6 +44,230 @@ pub trait ValidationDatabase:
...
+    fn check_input_object_type_definition(&self, object_type: hir::InputObjectTypeDefinition) -> Vec<ApolloDiagnostic>;
+    fn check_schema_definition(&self, object_type: hir::SchemaDefinition) -> Vec<ApolloDiagnostic>;
+    fn check_selection_set(&self, selection_set: hir::SelectionSet) -> Vec<ApolloDiagnostic>;
+    fn check_arguments_definition(&self, arguments_def: hir::ArgumentsDefinition) -> Vec<ApolloDiagnostic>;
+    fn check_field_definition(&self, field: hir::FieldDefinition) -> Vec<ApolloDiagnostic>;
+    fn check_input_values(&self, input_values: Arc<Vec<hir::InputValueDefinition>>) -> Vec<ApolloDiagnostic>;
+    fn check_db_definitions(&self, definitions: Arc<Vec<hir::Definition>>) -> Vec<ApolloDiagnostic>;
+    fn check_directive(&self, schema: hir::Directive) -> Vec<ApolloDiagnostic>;
+    fn check_arguments(&self, schema: Vec<hir::Argument>) -> Vec<ApolloDiagnostic>;
+    fn check_field(&self, field: Arc<hir::Field>) -> Vec<ApolloDiagnostic>;
```
lrlna, line 62: what would be the semantic difference of `validate_` and `check_`? Should we try merging them?

crates/apollo-compiler/src/validation/mod.rs
```diff
@@ -39,6 +44,230 @@ pub trait ValidationDatabase:
...
+
+pub fn check_schema_definition(db: &dyn ValidationDatabase, schema_def: hir::SchemaDefinition) -> Vec<ApolloDiagnostic> {
+    let mut diagnostics = Vec::new();
+
+    for directive in schema_def.directives() {
+        diagnostics.extend(db.check_directive(directive.clone()));
+    }
+
+    diagnostics
+}
```
lrlna, line 114: i like this, i think this should work! any individual parts (arguments, directives, fields) can be separated out and validated individually in the future.

lrlna:
>^ so instead of that we can implement validation a bit differently in the compiler. There can be validate_NODETYPE(hir_node) queries that recurse manually. that way we essentially have a tree walk, and all validation can be cached in a granular way

yesss definitely like this idea

crates/apollo-compiler/src/validation/mod.rs
goto-bus-stop: Ah, nothing, it's just to prevent naming collisions while migrating.

lrlna: Gotcha!

lrlna (approved):
\o/ merging this in

PR #323, opened 2022-10-07
Reduce token copying

goto-bus-stop:
A little low-hanging fruit.

Since the Lexer instance isn't needed anymore after the lexing is complete,
we can take ownership of the tokens and errors vectors and reverse them
in-place without making a copy. Big schemas can have 100K+ tokens so
it's actually quite a lot of work to copy them.

As it is, maybe it makes sense to have the lexer just be
```rust
fn lex(input: &str) -> (Vec<Token>, Vec<Error>)
```
unless we're planning to add more features to it in the future?
(personally I also like the idea of a streaming lexer like in #115!)

Also spotted a few other clones in the parser that are not (or no longer)
necessary. Likely only the `expect` one has a chance of making a difference though.

### DIFF
--- a/crates/apollo-parser/src/lexer/mod.rs
+++ b/crates/apollo-parser/src/lexer/mod.rs
@@ -83,6 +83,11 @@ impl Lexer {
     pub fn errors(&self) -> Iter<'_, Error> {
         self.errors.iter()
     }
+
+    /// Consume the lexer and return the tokens and errors.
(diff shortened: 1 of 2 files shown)


Geal:
now that the work to reverse the token list is done, it could be a good idea to revisit https://github.com/apollographql/apollo-rs/pull/115 or https://github.com/apollographql/apollo-rs/pull/322

lrlna:
yes! we were just talking about that on Monday. It's prioritised for the nearest of futures.

PR #324, opened 2022-10-16
fix(apollo-parser): handle unexpected tokens in document

JrSchild:
Unhandled tokens directly inside a document would break the loop in the parser. This resulted in the rest of the parsing to be skipped. An error would be expected here instead.

### DIFF
--- a/crates/apollo-parser/src/parser/grammar/document.rs
+++ b/crates/apollo-parser/src/parser/grammar/document.rs
@@ -28,7 +28,7 @@ pub(crate) fn document(p: &mut Parser) {
                 select_definition(def, p);
             }
             TokenKind::Eof => break,
-            _ => p.err_and_pop("expected a String, Name or OperationDefinition"),
+            _ => p.err_and_pop("expected a StringValue, Name or OperationDefinition"),
         }
     }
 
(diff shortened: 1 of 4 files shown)


JrSchild:
Thank you too @lrlna! I've picked up a lot of nice things from this project. This, and rust-analyzer is a delight to learn. I did find myself copying a lot though; One can dream that one day we have some these components generically available.

A possible bug I noticed was that the span's seem incorrect:
```
- DOCUMENT@0..81
    - WHITESPACE@0..1 "\n"
...
- ERROR@0:1 "expected a StringValue, Name or OperationDefinition" @
```
I would expect the whitespace token to have a different span than the error token. Maybe I'll take a look at it at some point if I have time. I think some tests would be nice that "re-assembles" the file based on the leaf-nodes and errors to assert that it produces the same output as the original input file, while verifying there is no overlap.

lrlna:
> Thank you too @lrlna! I've picked up a lot of nice things from this project. This, and rust-analyzer is a delight to learn. 

@JrSchild ah nice, I am glad!

>A possible bug I noticed was that the span's seem incorrect

Good catch! That seems to be a bug with the way ignored tokens are handled when popping an error, and not with the work you've done here. You can see a similar incorrect range for a missing definition that precedes a comment, an error which is handled in a different non terminal procedure:

```
cats
# comment
```
ast:
```
- DOCUMENT@0..10
    - WHITESPACE@0..1 "\n"
    - COMMENT@1..10 "# comment"
- ERROR@0:4 "expected definition" cats
```

PR #328, opened 2022-10-25
feat(compiler): pub compiler storage

lrlna:
While trying to come up with a nice example to show how the compiler's query_groups can be used in other contexts, I realised we are not exporting any of our internal storage structs. This allows for internal query_groups to be exported, and adds a specific example writing a linter.

It was a bit difficult to create lints with `name` fields in all of the hir nodes not having their own ast_nodes attached to them, so this also adds ast node pointers to all the names.


crates/apollo-compiler/src/database/def.rs
```diff
@@ -327,7 +327,10 @@ fn operation_definition(
     // check if there are already operations
     // if there are operations, they must have names
     // if there are no names, an error must be raised that all operations must have a name
-    let name = op_def.name().map(|name| name.text().to_string());
+    let name = op_def.name().map(|name| Name {
```
goto-bus-stop, line 330: many of these look like they could use the `name()` function instead of manually creating Name nodes each time. or is something preventing that?

lrlna: these are all a bit of special cases, the name() method we have takes an option of name and creates a node, some of these are option of named type, or something else. butttt i can just make a method for the hir node creation

goto-bus-stop (approved):
lgtm!

PR #329, opened 2022-10-26
feat(encoder): provide TryFrom<parser types> for encoder types

goto-bus-stop:
Provides `TryFrom` impls for the apollo-encoder AST types.

The conversion effectively assumes that the apollo-parser AST is valid and complete. Otherwise you get an Err, but not a very useful one, because the TryFrom impl doesn't know anything about its parent (so we can't show source code where the error originated for example).

I put it behind a feature flag so apollo-encoder doesn't *have* to depend on apollo-parser, not sure if that is truly worth it, so i also enabled it by default 😛 .

### DIFF
--- a/crates/apollo-encoder/Cargo.toml
+++ b/crates/apollo-encoder/Cargo.toml
@@ -24,7 +24,6 @@ apollo-parser = { version = "0.2.12", optional = true }
 thiserror = "1.0.37"
 
 [features]
-# TODO(@goto-bus-stop) remove before merging, just easier this way while writing : )
 default = ["from-parser"]
 from-parser = ["apollo-parser"]
 
--- a/crates/apollo-encoder/Cargo.toml
+++ b/crates/apollo-encoder/Cargo.toml
@@ -19,6 +19,15 @@ categories = [
 ]
 edition = "2021"
 
+[dependencies]
+apollo-parser = { version = "0.2.12", optional = true }
+thiserror = "1.0.37"
+
+[features]
+# TODO(@goto-bus-stop) remove before merging, just easier this way while writing : )
+default = ["from-parser"]
+from-parser = ["apollo-parser"]
+
 [dev-dependencies]
 pretty_assertions = "0.7.1"
 indoc = "1.0.3"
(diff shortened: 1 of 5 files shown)


crates/apollo-encoder/Cargo.toml
```diff
@@ -19,6 +19,14 @@ categories = [
 ]
 edition = "2021"
 
+[dependencies]
+apollo-parser = { version = "0.2.12", optional = true }
+thiserror = "1.0.37"
+
+[features]
+default = ["from-parser"]
+from-parser = ["apollo-parser"]
```
lrlna, line 28: should we just call this feature `apollo-parser`?

lrlna (approved):
yay! excited for this feature!!

crates/apollo-encoder/Cargo.toml
goto-bus-stop: i'm not sure what the general practice is on this in rustland 🤔 I used `from-parser` to describe what the feature actually does. but, if there are other features in the future that also rely on apollo-parser, it's fine if the TryFrom impls are made available anyways, so maybe that is actually counterproductive.

lrlna: I am leaning towards having this TryFrom be always available. It's part of the same ecosystem, so it makes sense.

goto-bus-stop: ok, I think then the current way is fine, as it's enabled by default but you can opt-out by setting `default-features=false` if you really only need the encoder and don't want the extra dependencies.

goto-bus-stop:
One thing I noticed while working on this, is that the `Display` impls on AST nodes make it easy to do the wrong thing. I was doing `.name().to_string()` all over the place but that actually returns the source text including possible whitespace, not only the name. The same goes for StringValue nodes, where `node.to_string()` includes quotes. I think accessing the source code is a bit of a specialised operation that doesn't need to be so easy. I think we should either remove the Display impl for parser nodes entirely or replace it by a named method like `.source_string()`.

crates/apollo-encoder/Cargo.toml
goto-bus-stop: decided to do tht^ in slack : )

PR #330, opened 2022-10-27
feat(parser): remove the impl Display for generated nodes

goto-bus-stop:
As mentioned in https://github.com/apollographql/apollo-rs/pull/329#issuecomment-1293416623

The `Display` impls for generated nodes returned the source text for that node. That's not a super common operation but it was very easy to access. It's also a very different operation from eg. `let content: String = node.string_value().into()` which returns the *content* of a string: `node.string_value().to_string()` returned the string as it was written in the source code, quotes and escapes and all.

Now `.to_string()` is replaced by a `.source_string()` method. It allocates a new String (just like `.to_string()` did). A syntax node can represent multiple slices (I think to support different structures like Ropes as input?), so slicing the original source isn't actually possible.


crates/apollo-smith/src/description.rs
```diff
@@ -51,6 +51,13 @@ pub enum StringValue {
     Line(String),
 }
 
+#[cfg(feature = "parser-impl")]
+impl From<apollo_parser::ast::StringValue> for StringValue {
+    fn from(val: apollo_parser::ast::StringValue) -> Self {
+        Self::from(Into::<String>::into(val))
```
goto-bus-stop, line 57: This `Into::<String>::into()` business is a bit awkward, i wonder if we could `impl From<StringValue> for String` instead? or provide a separate named method that returns the unquoted contents of a StringValue?

crates/apollo-smith/src/description.rs
```diff
@@ -26,7 +26,7 @@ impl From<Description> for String {
 #[cfg(feature = "parser-impl")]
 impl From<apollo_parser::ast::Description> for Description {
     fn from(desc: apollo_parser::ast::Description) -> Self {
-        Description(StringValue::from(desc.to_string()))
+        Description(desc.string_value().map(|s| s.into()).unwrap_or_else(|| StringValue::Line(Default::default())))
```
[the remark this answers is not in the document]
lrlna: ah, yea the original line should have been `desc.text().to_string()`, since `.text()` strips away the extra quotation marks around string values. So that's something to keep in mind with this PR - source_string() for string values that have quotes around them should have those quotes stripped.

crates/apollo-smith/src/description.rs
lrlna: We already have `impl Into<String> for ast::StringValue` [here](https://github.com/apollographql/apollo-rs/blob/e8a4540f88b5d504a36380d82a8dfd40787a4a50/crates/apollo-parser/src/ast/node_ext.rs).`From`s are usually nicer, but that requires us to create a syntax node with an unknown source, which perhaps is a bit too tricky? What do you think?

lrlna (commented):
I think we need a few tests to go along with this in the parser side of things. I mostly want to make sure we know if this is going to be a breaking change, for example #98 still works etc. i think a few tests for a some of node conversions, as well as making sure we can still properly display and have the source string of the whole tree (and it's equivalent to the original input). What do you think?

lrlna:
hmmm you mean the other way around? Our `Into<String>` does not work on references, and only works on owned values:
```rust
impl Into<String> for ast::StringValue {
    fn into(self) -> String {
      //
    }
}
```

In your test you can't use the `val` after you've converted it, since it moved.

(separate PR is good!)

lrlna (approved):

goto-bus-stop:
Yes i mean that's what it does, but the implementation of the function doesn't actually require ownership, it only takes ownership to satisfy the trait. The same feature could be implemented as a `fn contents(&self) -> String` on StringValue without having to consume the value.

PR #331, opened 2022-10-28
fix(compiler): fix apollo-parser dependency version

goto-bus-stop:
I assume this was a typo. Caused warnings like:
```
warning: crates/apollo-compiler/Cargo.toml: unused manifest key: dependencies.apollo-parser.apollo-parser
```


lrlna (approved):
weeeeiiiird how did this compile on all of the CI machines & my computers without any warnings 🤔 🤔

PR #333, opened 2022-10-28
Add subtype_map and is_subtype queries

### LINKED ISSUE #297: is_subtype query
This is one of the router's usecases.

Let's say we have a schema like this:

```graphql
type Query {
    me: String
}
interface Foo {
    me: String
}
interface Bar {
    me: String
}
interface Baz {
    me: String,
}

type ObjectType implements Foo & Bar {
  me: String
}

interface InterfaceType implements Foo & Bar {
  me: String
}

union UnionType = Foo | Bar | Baz
extend union UnionType = Baz
```

`is_subtype` function would be used to run validations like these:
- `Foo` is a subtype of `UnionType` (true),
- `Baz` is a subtype of `UnionType` (true) ,
- `ObjectType` is a subtype of `Foo` (true), 
- `InterfaceType` is a subtype of `Foo` (true),
- `Foo` is a subtype of `InterfaceType` (false). 

This function should live in one of the db's query groups (maybe `Document`) and make use of the already in use `find_*` queries.

SimonSapin:
closes #297


SimonSapin:
On further thought, is it worth caching results of "is_subtype" at all? It’s rather cheap once `subtype_map` is available/cached.

PR #338, opened 2022-10-31
chore(apollo-rs): run cargo fmt check in CI's lint command

lrlna:
What do we think? Is this going to be _very_ annoying?


goto-bus-stop:
I think it'll b e fine as we get the checks on PRs, where its not different from having clippy complain, but easier to fix 🤔

lrlna:
ok, i'll leaved this in, if it starts getting annoying, we can always remove

PR #344, opened 2022-11-03
Extend ast::*Value node conversions

SimonSapin:
See respective commit messages

### DIFF
--- a/crates/apollo-parser/CHANGELOG.md
+++ b/crates/apollo-parser/CHANGELOG.md
@@ -17,6 +17,36 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## Maintenance
 
 ## Documentation -->
+
+# [x.x.x] (unreleased) - 2021-mm-dd
+
+## Features
+
+- **extend `ast::*Value` node conversions - [SimonSapin], [pull/344]**
+
+  These node types implement conversion to standard types, extracting their value:
+
+  - `StringValue` → `String`
+  - `IntValue` → `i32`
+  - `FloatValue` → `f64`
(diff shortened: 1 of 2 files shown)


SimonSapin:
The CI failure seems to be unrelated but due to Clippy 1.65 having a new warning

goto-bus-stop:
the clippy warning should be addressed on main

goto-bus-stop (changes requested):
this is great and something I also wanted while working on parser → encoder node conversions 😇

the changelog got auto-formatted, can you remove all the changes and only add the new entry?

goto-bus-stop:
Would it make sense to *only* have the borrowed impls like `From<&StringValue> for String` etc? As a user you can just do `String::from(&ast_node)`, and with `ast_node.into()` it's even better. Or is there a compelling use case for the owned variants?

goto-bus-stop (approved):

lrlna (approved):
thank you!

SimonSapin:
Sorry I didn’t check the diff! :sweat_smile: I’ve reverted the changelog formatting.

Having only the borrowed impls would be sufficient in that you can accomplish the same things, but having both means you don’t *need* to add an explicit `&` borrow if the value you have happens to be owned and not used again later. Especially when using `Into` (because type inference has enough context), you’d need parenthesis: `(&node).into()`

goto-bus-stop:
Ahh OK I thought `node.into()` would automatically borrow it. makes sense to me!

SimonSapin:
Uhh now I’m not sure. It might in some cases, but inference might not figure it out as often maybe? I definitely remember `(&some).thing()` with lots of generics involved.

PR #346, opened 2022-11-04
docs(parser): example of modifying queries with parser + encoder

lrlna:
A few folks have been trying to modify/generate new schemas and queries using the AST provided with the parser. Given changes in #329, it's a bit easier generating SDL with the encoder using `ast_type.try_from()?`. 

This example outlines two possible modifications:
- merging two queries
- omitting certain fields in a query.

### DIFF
--- a/crates/apollo-parser/examples/modify_query_using_parser_and_encoder.rs
+++ b/crates/apollo-parser/examples/modify_query_using_parser_and_encoder.rs
@@ -1,4 +1,4 @@
-use apollo_encoder::{self, Document, Field, OperationDefinition, Selection, SelectionSet};
+use apollo_encoder;
 use apollo_parser::{
     ast::{self, AstNode},
     Parser,
@@ -6,7 +6,7 @@ use apollo_parser::{
 
 use anyhow::Result;
 
-fn merge_queries() -> Result<Document> {
+fn merge_queries() -> Result<apollo_encoder::Document> {
     let query = r"#
     query LaunchSite {
       launches {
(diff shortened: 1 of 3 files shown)


goto-bus-stop (commented):
if i do 
```
cargo run -p apollo-parser --example modify_query_using_parser_and_encoder
```
the `@omitted` fields are shown in the output 🤔 maybe there's a logic bug still?

crates/apollo-parser/examples/modify_query_using_parser_and_encoder.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    let mut new_query = apollo_encoder::Document::new();
+    for def in doc.definitions() {
+        if let ast::Definition::OperationDefinition(op) = def {
+            let mut selection_set = apollo_encoder::SelectionSet::new();
+            for selection in op.selection_set().unwrap().selections() {
+                if let ast::Selection::Field(field) = selection {
+                    if let Some(dir) = field.directives() {
+                        for dir in dir.directives() {
+                            if dir.name().unwrap().source_string() != "omitted" {
```
goto-bus-stop, line 88: Ah we need `.text()` here instead of `.source_string()`, as `.source_string()` includes surrounding whitespace tokens.
```suggestion
                            if dir.name().unwrap().text() != "omitted" {
```

goto-bus-stop: (I think this example might also do the wrong thing with multiple directives on a single field, as we're adding the field to the `new_query` inside a loop)

crates/apollo-parser/examples/modify_query_using_parser_and_encoder.rs
```diff
@@ -0,0 +1,122 @@
...
+                                selection_set.selection(apollo_encoder::Selection::Field(
+                                    field.clone().try_into()?,
+                                ))
+                            }
+                        }
```
goto-bus-stop, line 93: this should work I think:
```suggestion
                        let omit = dir.directives().any(|dir| dir.name().unwrap().text() == "omitted");
                        if !omit {
                            selection_set.selection(apollo_encoder::Selection::Field(
                                field.clone().try_into()?,
                            ))
                        }
```

lrlna: ah you're right! thank you so much :)

goto-bus-stop (approved):

PR #347, opened 2022-11-04
feat(parser): limit number of tokens to parse, closes #327

### LINKED ISSUE #327: apply recursion limit api to all tokens
To help with securing parser usage in server and/or router environments we should allow limiting the number of tokens that can parsed by the parser. This can expand on the existing recursion limit API we already have, but apply to not only selection sets, but all nodes.

### Prior art
GraphQL Java has [an implementation limiting number of parsed tokens](https://github.com/graphql-java/graphql-java/pull/2549) to help prevent DOS attacks in java graphql servers. And so does [graphql-js](https://github.com/graphql/graphql-js/pull/3702)

goto-bus-stop:
Still a few things to do:
- [x] document it
- [x] change the `let-else` usage as it's like … 2 days old. do we have an official MSRV? 🤔 
  - going to follow the router MSRV for now (1.63)
- [x] add tests with some cases where the token limit is reached in the middle of parsing a node


goto-bus-stop:
misclick :D 

I tried to do it entirely in the Parser but it's actually not as simple, because we do need to communicate the `index` into the original source string at the point where we abort parsing, and the Parser doesn't know about indices. So we'd have to add something to the lexer that lets you access the index at any time, which feels a bit ad hoc… I think it's nicer to keep it like I did here.

crates/apollo-parser/src/error.rs
```diff
@@ -34,10 +34,37 @@ use std::fmt;
...
+            Self::Text(text) => text.len(),
+        }
+    }
+}
+
+impl fmt::Display for ErrorData {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Eof => write!(f, "EOF"),
+            Self::LimitExceeded => Ok(()),
```
lrlna, line 57: What does `Ok(())` mean in the context of displaying limit errors?

crates/apollo-parser/src/error.rs
```diff
@@ -34,10 +34,37 @@ use std::fmt;
...
+            Self::Text(text) => write!(f, "{}", text),
+        }
+    }
+}
+
+#[derive(PartialEq, Eq, Clone, thiserror::Error)]
+#[error("ERROR@{index}:{} {message:?} {data}", .index + .data.len())]
```
lrlna, line 64: this reminds me, we should implement a better `Display` for just the parser errors. We didn't implement it earlier while still deciding on how diagnostics will look like in the compiler, but now those are pretty certain, so these can have a display too.

crates/apollo-parser/src/lexer/mod.rs
```diff
@@ -55,9 +56,16 @@ impl<'a> Lexer<'a> {
             input,
             index: 0,
             finished: false,
+            // An arbitrary default limit that fits most documents without risking memory exhaustion.
+            limit: LimitTracker::new(15_000),
```
lrlna, line 60: i am thinking some of our many-mb documents will totally have more than 15K (lexical) tokens. What do you think is a good way to stress test these limits? I guess I am trying to see where to draw a line between malicious vs just very very large schemas.

lrlna: let's actually maybe talk to the router folks and see what they think about this number in particular

crates/apollo-parser/src/parser/mod.rs
```diff
@@ -107,6 +107,15 @@ impl<'a> Parser<'a> {
...
 
+    /// Configure the limit on the number of tokens to parse. If an input document
+    /// is too big, parsing will be aborted.
+    ///
+    /// The default limit is 15 000.
+    pub fn with_token_limit(mut self, token_limit: usize) -> Self {
+        self.lexer = self.lexer.with_limit(token_limit);
+        self
+    }
+
```
lrlna, line 118: i feel like the router will probably want to do a recursion limit _and_ token limit. Should we write up an example on how users can set both of these limits in the parser?

lrlna (commented):
Stopping in the lexer when token limit is reached is fine, I think. I can see that it'll become a bit more convoluted if we just do it in the parser. One question I kind of just thought about reading through this - should we consider error tokens as tokens as well? Let's say we have a limit at 10, and there are exactly 10 tokens and 1 lexical error, do we return a limit exceeded error? What happens if we have 10 tokens and 1 _syntactic_ error, do we also return a limit exceeded error?


I'd like to see a few more tests in our test_dir thought:
- a document with a few lexical errors and limit exceeded error
- a document with a few syntactic errors and limit exceeded error (i.e. what happens if limit is exceeded, but we find parsing errors after limit is exceeded)
- a document that passes the token limit but has a recursion limit error
- a document that passes recursion limit but has a token limit error
- if we are choosing to count errors as part of our token limit, let's have something to test a error token limits as well (both lexical and parsing)

(i've got mine what can go wrong hat on haha)

crates/apollo-parser/src/lexer/mod.rs
goto-bus-stop: yeah. 15K feels quite low to me, it's what the java impl went with, but it's probably not a good limit for a schema. it's probably fine for a query?

crates/apollo-parser/src/parser/mod.rs
goto-bus-stop: was wondering if i should change the `with_recursion_limit` to a builder method here, instead of a second constructor?
```rust
Parser::new(input)
  .with_recursion_limit(1_000)
  .with_token_limit(70_000)
```

crates/apollo-parser/src/error.rs
goto-bus-stop: doesn't display any data for the error. so it results in something like `ERROR@245460:245460 "token limit reached, aborting parsing" `.

crates/apollo-parser/src/lexer/mod.rs
lrlna: i've got a few samples of queries that are totally larger than that. i ran a few that hit a recursion limit of 26K, which means they'll have even more tokens 😅 😅 😅

crates/apollo-parser/src/parser/mod.rs
lrlna: i like a builder as an api for this! maybe even simpler?

```rust
Parser::new(input)
  .recursion_limit(1_000)
  .token_limit(70_000)
```

lrlna (approved):
let's do it! i think the `with_recursion_limit` rename to `recursion_limit` and introduction of the "limit builder pattern" means this will have to released as `apollo-parser@0.4.0`. We should also make sure we modify that API in the router's code after we release.

PR #352, opened 2022-11-08
apollo-smith@0.2.0

### LINKED ISSUE #351: apollo-smith 0.1.5 and 0.1.4 are semver-incompatible with 0.1.3
## Description

apollo-router CI started failing in the job that compiles the projects with `Cargo.lock` (so equivalent to running `cargo update` and getting latest versions of everything that Cargo considers compatible).

https://app.circleci.com/pipelines/github/apollographql/router/7731/workflows/e3460290-3355-42bf-9c8b-685aaf419830/jobs/58134?invite=true#step-108-472

### Steps to reproduce

* Have dependencies like this:
  ```toml
  apollo-smith = { version = "0.1.3", features = ["parser-impl"] }
  apollo-parser = "0.2.12"
  ```
* Construct a value of type `apollo_parser::ast::Document`
* Pass it to `apollo_smith::Document::from`
 
https://github.com/apollographql/router/blob/62d7452373f4eeb5343415fb08da2d33101434f4/fuzz/src/lib.rs#L31

### Actual result

apollo-smith 0.1.3 has a corresponding `From` impl so this compiles. However 0.1.4 and 0.1.5 update their apollo-parser dependency to 0.3, so there is no impl for the 0.2 `Document` anymore.

Note that [in Cargo dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) the default is to specify "anything semantic-versioning-compatible with this version", not "exactly this version": `version = "0.1.3"` is equivalent to `version = "^0.1.3"` which is equivalent to `version = ">=0.1.3, <0.2.0"`. (Cargo deviates from semver.org for 0.y.z.)

### Expected result

It’d be nicer for users if apollo-rs follows the Rust ecosystem’s conventions for semantic versioning. In this case, since `apollo-parser` is part of the public API of `apollo-smith` (at least through that `From` impl), this means that updating `apollo-smith` to a semver-incompatible version of `apollo-parser` should be considered a breaking change and be published as 0.2.0 instead of 0.1.4.

A work-around is to update other uses of `apollo-parser` to 0.3, but that may required other code changes: https://github.com/apollographql/router/pull/2051

In case other users might be affected, one option is to yank 0.1.4 and 0.1.5 from crates.io and republish as 0.2.0

goto-bus-stop:
[Rendered changelog](https://github.com/apollographql/apollo-rs/blob/apollo-smith@0.2.0/crates/apollo-smith/CHANGELOG.md#020---2022-11-08)

closes #351

### DIFF
--- a/crates/apollo-smith/CHANGELOG.md
+++ b/crates/apollo-smith/CHANGELOG.md
@@ -18,7 +18,24 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## Maintenance
 
 ## Documentation -->
-# [0.1.5](https://crates.io/crates/apollo-smith/0.1.5) - 2022-11-04
+# [0.2.0](https://crates.io/crates/apollo-smith/0.2.0) - 2022-11-08
+
+## BREAKING
+
+- **update apollo-parser@0.3.x - [lrlna], [pull/340], [pull/348]**
+
+  This change was first released in the apollo-smith@0.1.4 patch release.
+  It should have been a breaking change, as the update to the new version
+  requires users to also update apollo-parser to 0.3.0 at the same time.
+
+  This version is identical to 0.1.5 except for the version number.
+  apollo-smith versions 0.1.4 and 0.1.5 have been yanked.
+
+  [lrlna]: https://github.com/lrlna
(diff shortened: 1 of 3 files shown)


lrlna (approved):
tysm!!

PR #357, opened 2022-11-10
fix(parser): fix lexing escaped and unicode characters in block strings

### LINKED ISSUE #341: lexing fails to tokenize unicode in block string values
This type with a description containing unicode fails to lex correctly:

```graphql
"""unicode in block string 🤷"""
input Filter {
    title: String
}
```

### LINKED ISSUE #342: lexing fails with escaped block string quotes in block string value
This description does not get lexed as a block string token:

```graphql
"""
\""" a/b \"""
"""
input Filter {
    title: String
}
```

### LINKED ISSUE #343: lexing fails with an escaped backslash in string value
The following string value does not lex into a string value token:

```graphql
type Query {
    format: String = "Y-m-d\\TH:i:sP" 
}
```

lrlna:
fixes #341 fixes #342 fixes #343


goto-bus-stop:
do we have a test for `\n`, `\t` and the like in strings?

lrlna:
@goto-bus-stop we do, but I added some extras for block string values.

PR #364, opened 2022-11-15
chore(compiler): Avoid double lookup in `SchemaDefinition::query/mutation/subscription`

SimonSapin:
The previous pattern of `db.find_object_type(op.object_type_id(db)?)` would first find an `ObjectTypeDefinition` by its name, extract its UUID, then find the same `ObjectTypeDefinition` again by its UUID. The last step is redundant.

This adds a `RootOperationTypeDefinition::object_type` public API. It could be private, but it seems reasonable to make public since `object_type_id` is already public.


goto-bus-stop (approved):
lgtm!

PR #366, opened 2022-11-16
Fix injected `@include` directive locations to match spec

allancalix:
Directive locations for the `@include` directive should include FragmentSpread as a location and not FragmentDefinition. Updating injected directive to match expectation.


lrlna (approved):
tysm!

PR #368, opened 2022-11-17
feat(compiler): multi file support

lrlna:


crates/apollo-compiler/src/database/ast.rs
```diff
@@ -1,20 +1,26 @@
...
+    /// SyntaxNodePtr to an actual SyntaxNode.
+    fn document(&self, file_id: FileId) -> GreenNode;
 
+    /// Get syntax errors found in the compiler's manifest.
     fn syntax_errors(&self) -> Vec<ApolloDiagnostic>;
 }
 
-fn ast(db: &dyn AstDatabase) -> SyntaxTree {
-    let input = db.input();
+fn ast(db: &dyn AstDatabase, file_id: FileId) -> SyntaxTree {
```
lrlna, line 22: The AST will always be computed in the same way for schemas or queries, since they are part of the same GraphQL Document grammar. But to ensure that the incoming queries (especially for the router) do not rewrite the schema context, we need to think of a way to separate _what_ can be added as HIR nodes from a query (only Operation and Fragment definitions). We should likely have a change to the HIR structure where `Definition` is indeed an enum of two types `ExecutableDefinition` and `TypeSystemDefinitionOrExtension`. Maybe then this function can distinguish between the two?

goto-bus-stop:
With different input types to salsa, we were discussing how to distinguish them, and when to use which input query. We talked about trying the different input types in turn, so if we were eg. looking up source code for a file, we could do (pseudocode):
```rust
db.input_document(id)
  .or_else(|| db.input_schema(id))
```
but salsa panics if an input cannot be resolved, so that wont actually work without further changes.

It's possible to do that if we store the source type (schema/query/document) in the source manifest. that could still be a good idea but I think this way is simpler: the compiler now uses 1 input type, which is a structure contaniing the source type AND the source strirng together, and everything can just use that directly and only check the source type if it's relevant. the API for typical uses is still the same.

crates/apollo-compiler/src/lib.rs
```diff
@@ -63,22 +72,55 @@ pub struct ApolloCompiler {
...
+        let id = self.source_manifest.add_source(path);
+        self.db.set_input_query(id, input.to_string());
+
+        id
+    }
+
+    /// Compile with currently set sources.
+    pub fn compile(&mut self) {
+        self.db.set_sources(self.source_manifest.clone());
     }
```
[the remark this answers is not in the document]
goto-bus-stop: It does look like we need something like this, as salsa doesnt have a way to list all known inputs. I think we could do it implicitly by storing the SourceManifest in the DB and updating it when we add a new source 🤔

goto-bus-stop:
@SimonSapin the PR still needs some work, for example the FileIds are still UUIDs here. I wanted to see if we can use salsa's interning for that, or else do the hashing strategy. But I think it's ready for an initial review and I could really use some feedback 😇

crates/apollo-compiler/src/database/inputs.rs
```diff
@@ -1,50 +1,34 @@
...
 
+    // NOTE(@lrlna): in the context of an environment where a schema represents
+    // the current state that must not be modified from incoming queries, we
+    // want to make sure there are two separate points of input: schemas and queries.
     #[salsa::input]
-    fn recursion_limit(&self) -> Option<usize>;
+    fn input_schema(&self, file_id: FileId) -> String;
+
+    #[salsa::input]
+    fn input_query(&self, file_id: FileId) -> String;
```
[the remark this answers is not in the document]
goto-bus-stop: I ended up making this use a single `input` query type that contains source code + an enum for the file type: https://github.com/apollographql/apollo-rs/pull/368#issuecomment-1325385532

SimonSapin (commented):
I haven’t read through `src/validation` yet. Are there "meaningful" changes there or is it all plumbing file IDs and updating calls for signature changes?

goto-bus-stop:
OK, now the `.schema()`/`.executable()`/`.document()` methods use salsa's interning to turn the file name into a file ID, and if you use a path that was already known to the database, it updates the file isntead of adding a new one. I think it's readyish to go.

the SourceManifest is basically replaced by having a `Vec` of all known file IDs stored in the database, and a `.input(file_id) -> source` query that includes the source code + filename in the output. you can use that to do `db.source_code(file_id)` for each file or something if necessary.

salsa interning would be really useful for the HIR too if we were to fully rely on it … but that would come in the way of the initialize-from-precomputed-HIR idea.

at the moment this doesn't come with a way to remove a file.

crates/apollo-compiler/src/validation/interface.rs
```diff
@@ -17,18 +17,20 @@ pub fn check(db: &dyn ValidationDatabase) -> Vec<ApolloDiagnostic> {
     //
     // Return a Unique Definition error in case of a duplicate name.
     let mut seen: HashMap<&str, &InterfaceTypeDefinition> = HashMap::new();
+    dbg!(db.interfaces());
```
goto-bus-stop, line 20: @goto-bus-stop remove

goto-bus-stop:
> I haven’t read through src/validation yet. Are there "meaningful" changes there or is it all plumbing file IDs and updating calls for signature changes?

It's basically all plumbing. `validation_db` has one meaningful change which is making a global `validate()` call run through _all_ the registered executable documents, but that's it.

SimonSapin:
> OK, now the `.schema()`/`.executable()`/`.document()` methods use salsa's interning to turn the file name into a file ID, and if you use a path that was already known to the database, it updates the file isntead of adding a new one. I think it's readyish to go.

Sorry I missed this earlier. A mandatory `PathBuf` is slightly awkward for input that might not come from a file at all, but even more so if it has to be unique. As long as the Router only deals with one executable document at a time per compiler instance it can hard-code a dummy string, but if that changes it’d need to generate random strings which feels like busywork

goto-bus-stop:
Hmm yes. I wasn't sure if this would be ideal long-term, but didn't want to change it if there was no reason to.

The alternative I had in mind was having something like `(&mut db).create_file_id()` which returns a unique file ID that you have to hold onto yourself. Then there would be no filename, or it could have an associated non-unique filename for diagnostics.
(This could also happen a bit more behind the scenes if we have `db.create_query("source code")` create a new unique file ID, and require use of a separate method like `db.update_query(id, "new code")` to change an existing input.)
Would that be better for a router-like case?

SimonSapin:
It looks like nothing uses the filename? It only serves to identify a previous source to change its text? In that case yes I’d prefer to remove filenames and have separate methods for updating an existing source (by `FileId`) v.s. creating a new one (returning a new `FileId`, which could be assigned outside of Salsa with a private `next_file_id: usize` field)

crates/apollo-compiler/src/validation/directive.rs
```diff
@@ -16,36 +16,16 @@ pub fn check(db: &dyn ValidationDatabase) -> Vec<ApolloDiagnostic> {
...
-                    .into();
+                let current_offset: usize = dir_def.loc().unwrap().offset();
+                let current_node_len: usize = dir_def.loc().unwrap().node_len();
                 errors.push(ApolloDiagnostic::UniqueDefinition(UniqueDefinition {
                     ty: "directive".into(),
                     name: name.into(),
-                    src: db.input(),
+                    src: db.source_code(prev_def.loc().unwrap().file_id()),
                     original_definition: (prev_offset, prev_node_len).into(),
                     redefined_definition: (current_offset, current_node_len).into(),
```
SimonSapin, line 30: Here (and multiple other places) we create a diagnostic that contains two `miette::SourceSpan` which are byte ranges into a `miette::SourceCode` string stored next two them. But with multi-input the two span now could be coming from different sources.

I looked at bit at Miette’s docs and I don’t know if it’s possible to associate  two `SourceCode` strings to a single diagnostic.

SimonSapin: It looks like this limitation is by design :( https://github.com/zkat/miette/issues/193

goto-bus-stop: yes I noticed this too. I was hoping to be able to address it later. miette doesn't support multi-file diagnostics but you can add multiple diagnostics across different files and mark them as related, so this could be an option.

(we could also propose a re-addition to miette if we can come up with something simple enough. all we really would need is a way to pass some value T (FileId) into a SourceSpan and get that propagated to the SourceCode trait, and the rest could be implemented in apollo-rs.)

goto-bus-stop:
Kept the filename as is, as it's marginally more work to remove it and we will need it eventually…

SimonSapin (approved):
Looks great, let’s get this merged!

PR #369, opened 2022-11-17
Add extension definitions to the HIR and subtype queries

SimonSapin:
Validation is still missing


lrlna (approved):
Thank you so much for doing all of these! super helpful

PR #371, opened 2022-11-21
feat(parser,smith): make the conversions to native types fallible

### LINKED ISSUE #358: Non-panicky conversions from AST nodes to Rust types
In apollo-parser we have a few `From` implementations to get Rust types out of AST nodes, like an `i32` from an `IntValue`. But these can sometimes panic, which is not expected for `From`, especially since it's dealing with user input that can error in a totally normal run of a program. These should be changed to `TryFrom`.

Also, we can add a conversion from `IntValue` to `f64`, which is supported by the GraphQL spec.

goto-bus-stop:
This PR changes the `From<$ParserNode> for $RustType` implementations to `TryFrom`, so that user input cannot cause panics as easily. For example, `i32::from(node)` could panic if `node` contains a very large integer.

This PR also adds an `impl TryFrom<IntValue> for f64` implementation, as you are allowed to write integers when a schema expects a float.

closes #358

These are still todos but could be done later:
- [ ] compiler is still panicky, we would have to make it store `Result`s
- [ ] now that `apollo-smith` uses `TryFrom`, it could use the `FromError::MissingNode` type just like apollo-encoder, instead of panicking on missing nodes.

### DIFF
--- a/crates/apollo-compiler/src/database/hir_db.rs
+++ b/crates/apollo-compiler/src/database/hir_db.rs
@@ -940,9 +940,10 @@ fn value(val: ast::Value) -> Value {
             ast_ptr: SyntaxNodePtr::new(var.syntax()),
         }),
         ast::Value::StringValue(string_val) => Value::String(string_val.into()),
-        ast::Value::FloatValue(float) => Value::Float(Float::new(float.into())),
-        ast::Value::IntValue(int) => Value::Int(int.into()),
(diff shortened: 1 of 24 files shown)


lrlna (approved):
This is definitely going to be a breaking change for the router team. I am wondering if before we merge this in, we should run it from a git commit in the router?

goto-bus-stop:
i prepared a patch to upgrade router to the new APIs, and their tests pass

lrlna:
oh yay, am glad!

PR #372, opened 2022-11-21
AST int → HIR float  when expected by type system (best effort)

SimonSapin:
The syntax for an integer value is also valid for a float value, but the latter has a wider range of accepted values (~53 bits v.s. 32 bits). This adds support for such "extra" values during HIR creation by returning a `hir::Value::Float` for an `ast::IntValue` when type system definitions indicate that a float is expected.

In some cases, type system definitions would have enough information but this does not yet implement all lookups necessary.

In other cases it truly does not have sufficient information. Either way, `hir::Value::Int` is used as a fallback as before.

### DIFF
--- a/crates/apollo-compiler/src/database/hir_db.rs
+++ b/crates/apollo-compiler/src/database/hir_db.rs
@@ -726,7 +726,7 @@ fn input_value_definitions(
             let description = description(input.description());
             let name = name(input.name());
             let ty = ty(input.ty().expect("Input Definition must have a type"));
-            let default_value = default_value(input.default_value());
+            let default_value = default_value(input.default_value(), Some(&ty));
             let directives = directives(input.directives());
             let ast_ptr = SyntaxNodePtr::new(input.syntax());
 
@@ -743,8 +743,16 @@ fn input_value_definitions(
     Arc::new(input_values)
 }
 
-fn default_value(default_value: Option<ast::DefaultValue>) -> Option<DefaultValue> {
-    default_value.map(|val| value(val.value().expect("Default Value must have a value token")))
+fn default_value(
+    default_value: Option<ast::DefaultValue>,
+    expected_type: Option<&Type>,
+) -> Option<DefaultValue> {
+    default_value.map(|val| {
+        value(
+            val.value().expect("Default Value must have a value token"),
+            expected_type,
+        )
+    })
 }
 
 fn root_operation_type_definition(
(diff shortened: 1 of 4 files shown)


crates/apollo-compiler/src/database/hir_db.rs
```diff
@@ -929,7 +942,12 @@ fn argument(argument: ast::Argument) -> Argument {
     }
 }
 
-fn value(val: ast::Value) -> Value {
+/// `expected_type` is the type of value that is expected to be returned,
+/// as far as known based on type system definitions.
+/// This affects values with integer syntax that should be interpreted as floats.
+///
+/// `None` means unknown, or type inference not fully implemented yet.
```
lrlna, line 949: This is sort of related to the conversation @goto-bus-stop and I were having. It feels like it's becoming a bit more important for us to be able to run extra compile-time checks, such as coercion or the type inference you're mentioning here. We are expecting the HIR to be this perfectly correct representation of the world, but in fact it can still be missing information or have incorrect types (like this argument value and its expected type). What it really means is that this hir_db is a bit overloaded.

I am wondering if in the long run we will have to create a further layer to perform some of these extra compile-time checks after some form of a HIR is created. This could either be done before or after the validation step.

lrlna (approved):
I think this works given our current setup. I'll make an issue for us to track handling inference and type coercions a bit better. The real solution is definitely a bit more of a structural change, and needs a bit more thinkingggg

crates/apollo-compiler/src/database/hir_db.rs
```diff
@@ -941,20 +959,34 @@ fn value(val: ast::Value) -> Value {
         }),
         ast::Value::StringValue(string_val) => Value::String(string_val.into()),
         ast::Value::FloatValue(float) => Value::Float(Float::new(float.into())),
-        ast::Value::IntValue(int) => Value::Int(int.into()),
+        ast::Value::IntValue(int) => {
+            if let Some(Type::Named { name, .. }) = expected_type {
+                if name == "Float" {
+                    return Value::Float(Float::new(int.as_float()));
```
goto-bus-stop, line 965: After #371 you can do `f64::try_from(int).unwrap()` here. (we will still panic for now, as not doing that is a bigger refactor 😬 )

SimonSapin:
Since `f64` losslessly represent all `i32` values, should there even be both `Float` and `Int` as separate variants of `hir::Value`? What if it was a single `Number { value: f64, syntactically_int: bool }`? That way HIR construction would not require type inference

lrlna:
@SimonSapin that... could work! how would you then run validation and response matching on top of that? store another kv to represent what the type provided was?

goto-bus-stop:
I also thought about that yesterday! I think it's a good solution, as you can delay picking a specific number type to the point where you actually need a native number, and there you probably already know the type that was declared in the schema (or you can decide you dont care and go f64 all the time)

SimonSapin:
I don’t know the details of validation and matching but the idea of `syntactically_int: bool` is that having it false where an integer is expected would cause an error

SimonSapin:
> iirc looking up the type definition at this exact moment here is going to cause a cycle, i.e. it's not possible ://

Since this HIR-construction-time type inference approach cannot be complete, closing in favor of the `hir::Value::Number` idea

PR #375, opened 2022-11-28
Add coercion of hir::Value::Int to Float input value without losing precision

SimonSapin:
Previously the HIR stored an `i32` for values with integer syntax, restricting the supported range if the value ended up used as a `Float`.

Now it stores `f64` for all numeric literals (though still with `Int` and `Float` variants). Coercion to `Int` is a separate fallible method.

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -941,6 +941,8 @@ impl TryFrom<&'_ Value> for f64 {
     }
 }
 
+#[derive(thiserror::Error, Debug)]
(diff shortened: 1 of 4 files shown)


goto-bus-stop (approved):
oh it actually works pretty nicely with just `Int(Float)` instead of the `Number(Float, SyntacticType)` thing. thanks 😄

SimonSapin:
Oh yeah I forgot to explain: I started with that and it felt silly to have a boolean field when we’re already have enum variants

SimonSapin:
`Int(Float)` looks unexpected at first but I think it’s ok with the doc-comment.

Issue #256, opened 2022-07-15
`apollo-compiler` panics when i do weird things with the schema


EverlastingBugstopper:
if you place characters after a `[Type]`, like `[Type]oops`, `apollo-compiler` will panic.

example valid schema:

```graphql
type Person {
  id: ID!
  name: String
  appearedIn: [Film]
  directed: [Film]
}

type Film {
  id: ID!
  title: String
  actors: [Person]
  director: Person
}

type Query {
  person(id: ID!): Person
  people: [Person]
  film(id: ID!): Film!
  films: [Film]
}

```

example panic schema:

```graphql
type Person {
  id: ID!
  name: String
  appearedIn: [Film]s
  directed: [Film]
}

type Film {
  id: ID!
  title: String
  actors: [Person]
  director: Person
}

type Query {
  person(id: ID!): Person
  people: [Person]
  film(id: ID!): Film!
  films: [Film]
}
```

abbreviated panic

```console
$ rover schema lint --schema ./schema.out
Validating that --schema ./schema.out conforms to the GraphQL specification.
thread 'main' panicked at 'Field must have a type', /home/a/.cargo/git/checkouts/apollo-rs-c2f4fdeb51021351/2008021/crates/apollo-compiler/src/queries/database.rs:963:28
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

panic with full backtrace

```console
$ rover schema lint --schema ./schema.out
Validating that --schema ./schema.out conforms to the GraphQL specification.
thread 'main' panicked at 'Field must have a type', /home/a/.cargo/git/checkouts/apollo-rs-c2f4fdeb51021351/2008021/crates/apollo-compiler/src/queries/database.rs:963:28
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:143:14
   2: core::panicking::panic_display
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:73:5
   3: core::panicking::panic_str
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:56:5
   4: core::option::expect_failed
[...]

PR #381, opened 2022-12-08
fix(compiler,parser): do not panic during HIR creation

goto-bus-stop:
Ref #256, probably not a *full* fix yet, but that issue is quite broad.

I set out to simply change every panic in HIR creation to returning a `None`, but while writing tests I triggered panics in the parser, so I tried to address those too.

- This removes some panics from type name parsing. For example, `field: []` does not panic anymore. Instead it produces a syntax error and an incomplete List type.
- Fixes the parser completely bailing out of parsing a list of field definitions when encountering a syntax error in one field.
   ```graphql
   type A {
      fieldA: [] # ← has error, missing item type
      fieldB: Int
      fieldC: Int
   }
   ```
   Previously fieldB and fieldC would not be parsed, now they are.
- Replaces panics when creating the HIR with `Option`s. This means that AST nodes with errors are *ignored* by the HIR. *Most* cases where AST nodes are missing already cause syntax errors in the parser, so users can still tell that something is wrong, and the validation etc. should not be expected to be perfect if there are syntax errors anyways.

   There are some cases where we ignore more of the tree than we need to. eg. if an object type definition is missing a name, we ignore the entire thing, even though we could still do useful validation on the directives and fields. We could add that stuff in the future by allowing for certain optional or invalid values in the HIR. As it is, silently ignoring those cases is still an improvement over panicking, I think (+ it's not actually silent because you do get a parse error already).

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -2359,4 +2359,19 @@ mod tests {
         assert_eq!(default_values[2], "98765432109876540000");
         assert_eq!(default_values[3], "98765432109876540000");
     }
+
+    #[test]
+    fn syntax_errors() {
+        let compiler = ApolloCompiler::new(
+            "type Person {
+                id: ID!
+                name: String
+                appearedIn: [Film]s
+                directed: [Film]
+            }",
+        );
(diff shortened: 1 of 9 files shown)


SimonSapin (approved):
Looks great!

PR #382, opened 2022-12-09
Add some accessors or convenience public APIs

SimonSapin:
* `DocumentDatabase::find_enum_by_name`, similar to existing `find_union_by_name`
* `Directive::argument_by_name`
* `impl Copy for hir::Float`
* `Float::get`
* `Float::to_i32_checked` (naming pattern from std’s to_int_unchecked)
* `ScalarTypeDefiniton::is_built_in`

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -678,6 +678,16 @@ impl Directive {
         self.arguments.as_ref()
     }
 
+    /// Get a reference to the value of the directive argument with the given name, if it exists.
+    pub fn argument_by_name(&self, name: &str) -> Option<&Value> {
+        Some(
+            self.arguments
+                .iter()
+                .find(|arg| arg.name() == name)?
+                .value(),
+        )
+    }
+
     // Get directive definition of the currently used directive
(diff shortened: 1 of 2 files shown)


crates/apollo-compiler/src/database/hir.rs
```diff
@@ -1326,6 +1333,20 @@ impl Float {
             inner: OrderedFloat(float),
         }
     }
+
+    pub fn get(self) -> f64 {
+        self.inner.0
+    }
+
+    /// Convert the value to the nearest i32, if it is finite and in the i32 range.
```
goto-bus-stop, line 1341: Does it round to the nearest i32 or does it truncate? I thought `as` truncates

goto-bus-stop (approved):
Looks nice, one doc question 😇 thanks!

crates/apollo-compiler/src/database/hir.rs
SimonSapin: I should have checked rather than write this from memory :sweat_smile: https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast says float `as` int rounds towards zero. I’ll fix the doc-comment.

(This public API could be used in cases where this makes a difference, but it’s mostly useful when matching on `hir::Value::Int` where the conversion will either fail or be exact.)

Issue #386, opened 2022-12-15
validation: name uniqueness should apply across different definition types


goto-bus-stop:
For example:

```graphql
type SomeIdentifier {}
scalar SomeIdentifier
input SomeIdentifier {}
```
is not valid, but currently allowed(?)

PR #387, opened 2022-12-16
feat(compiler): Name-indexed maps + inlined extensions

SimonSapin:
* For all kinds of definitions that have a mandatory name, keep them in source-order-preserving `IndexMap` instead of a `Vec` to speed up name lookup.

* Remove type extensions as independent HIR items, instead store them in a new `extensions` struct field on the respective definitions they extend.

* Wrap more HIR components in Arc, to reduce cost of Salsa’s cloning.

### DIFF
--- a/crates/apollo-compiler/src/database/hir_db.rs
+++ b/crates/apollo-compiler/src/database/hir_db.rs
@@ -7,6 +7,7 @@ use apollo_parser::{
 use uuid::Uuid;
 
 use crate::{database::FileId, hir::*, AstDatabase, InputDatabase};
+use indexmap::IndexMap;
 
 // HIR creators *ignore* missing data entirely. *Only* missing data
 // as a result of parser errors should be ignored.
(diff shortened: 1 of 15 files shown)


crates/apollo-compiler/src/database/hir_db.rs
```diff
@@ -253,201 +201,80 @@ fn schema(db: &dyn HirDatabase) -> Arc<SchemaDefinition> {
...
-        .collect();
-    let scalars = built_in_scalars(scalars);
-
-    Arc::new(scalars)
+macro_rules! by_name {
+    ($db: ident, $convert: expr) => {{
+        let mut map = IndexMap::new();
+        for def in type_definitions($db, $convert) {
+            let name = def.name().to_owned();
+            map.entry(name).or_insert_with(|| Arc::new(def));
```
goto-bus-stop, line 209: we should also record duplicates here for validation. I can work on that since validation isn't the most pressing thing for the router, but then a todo comment would be nice 😇 . probably we could have a struct that contains both the IndexMap and a Vec of errored definitions / extensions.

crates/apollo-compiler/src/database/hir_db.rs
```diff
@@ -85,141 +73,105 @@ fn db_definitions(db: &dyn HirDatabase) -> Arc<Vec<Definition>> {
...
+    }
+    Arc::new(map)
+}
+
+/// Takes a fallible conversion from a specific AST type to an HIR type,
+/// finds matching top-level AST nodes in type definition files,
+/// and returns an iterator of successful conversion results.
+///
+/// Failed conversions are ignored.
+fn type_definitions<'db, AstType, TryConvert, HirType>(
```
goto-bus-stop, line 157: nifty!

goto-bus-stop (commented):
i really like the changes here! the `extensions`  field is a nice way to address what we were talking about with merged/unmerged HIR extension nodes 👍🏻

crates/apollo-compiler/src/database/hir_db.rs
```diff
@@ -253,201 +201,80 @@ fn schema(db: &dyn HirDatabase) -> Arc<SchemaDefinition> {
...
-        .filter_map(|(id, definition)| match definition {
-            ast::Definition::ScalarTypeDefinition(scalar_def) => scalar_definition(scalar_def, id),
-            _ => None,
-        })
-        .collect();
-    let scalars = built_in_scalars(scalars);
-
-    Arc::new(scalars)
+macro_rules! by_name {
+    ($db: ident, $convert: expr) => {{
```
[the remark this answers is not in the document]
SimonSapin: It’s a macro because it calls `.name()` inherent methods of different types. The most straightforward way to make this not a macro would be to add a trait for that method, but that seems heavyweight since there aren’t other uses.

However, thinking about it again now, we could have the `convert` callback return `Option<(String, HirType)>` with the name instead of `Option<HirType>`. I’ll give it a go

crates/apollo-compiler/src/database/hir_db.rs
SimonSapin: There was already validation code looking for duplicate names. In this PR I’ve changed it to be based on AST definitions where duplicates are still present despite this HIR change.

(BTW it looks like we should check for name collisions of type definitions regardless of their kind, not just collisions among object types, and separately collisions among enums, etc. I can make another PR for that.)

goto-bus-stop: Oh I actually read that AST-based validation and then forgot about it when I read this. that's OK, I filed https://github.com/apollographql/apollo-rs/issues/386 for different-kinded collisions, I'd be happy to do it too.

crates/apollo-compiler/src/database/hir_db.rs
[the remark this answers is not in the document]
SimonSapin: Ah but the next macro also pushes to `extensions` fields of different structs, and I can’t seem to find a way to express that with generics that feels nicer than macros.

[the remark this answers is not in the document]
goto-bus-stop: Gotcha, yeah I guess we'd have to have a trait and implement it for each relevent struct, which would be easiest to do with a macro anyways 😝 this is the better solution!

PR #389, opened 2022-12-19
Move named types to a separate HIR enum and separate namespacing

SimonSapin:
* Split `hir::Definition` enum variants that define types into a new `NamedTypeDefinition` enum.

* Add Salsa queries for finding a type by its name, (excluding non-type named definitions), and use them where appropriate. This changes some diagnostics from "wrong definition kind" to "no such type".

* Remove `find_type_system_definition_by_name` and `find_definition_by_name` queries since they mix up the four namespaces of a GraphQL document that should really be separate. Instead its callers should probably use one of:

  * `find_directive_definition_by_name`
  * `find_operation_by_name`
  * `find_fragment_by_name`
  * `find_type_definition_by_name`

### DIFF
--- a/crates/apollo-compiler/src/database/document.rs
+++ b/crates/apollo-compiler/src/database/document.rs
@@ -4,6 +4,7 @@ use std::{
 };
 
 use crate::{hir::*, AstDatabase, FileId, HirDatabase, InputDatabase};
+use indexmap::IndexMap;
 
 #[salsa::query_group(DocumentStorage)]
 pub trait DocumentDatabase: InputDatabase + AstDatabase + HirDatabase {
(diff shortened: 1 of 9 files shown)


lrlna (commented):
`NamedType` has meaning that is not equivalent to what you're describing in this PR ([spec](https://spec.graphql.org/October2021/#NamedType)). The definitions you're grouping here are referred to as `TypeDefinition` ([spec](https://spec.graphql.org/October2021/#TypeDefinition)), and we should probably stick to the spec-defined vocab as much as possible to avoid confusion.

SimonSapin:
I've renamed `NamedTypeDefinition` to `TypeDefinition`

SimonSapin:
(The two commits make sense separately, so this could be rebased instead of squashed.)

lrlna (approved):
tysm!

Issue #388, opened 2022-12-19
Compile errors with apollo-smith/parser-impl feature


SimonSapin:
## Description

I’m trying to make a local script to run roughly the same things as CI before submitting PRs. I see that `.circleci/config.yml` contains `cargo clippy --all-targets --all-features -- -D warnings`, but when I run the same on `main` I get compilation errors.

[CI is green for the same commit](https://app.circleci.com/pipelines/github/apollographql/apollo-rs/760/workflows/05fcefe8-a124-419c-ace7-b3ebabf9c934/jobs/4886?invite=true#step-107-2), at this point I don’t understand what difference in my local environment leads to different results.

### Steps to reproduce

The above can be reduced to:

```rust
cargo check -p apollo-smith --features parser-impl
```

### Expected result

Compilation succeeds

### Actual result

31 errors, all similar to this:

```rust
    Checking apollo-smith v0.3.1 (/Users/simon/projects/apollo-rs/crates/apollo-smith)
error[E0107]: this type alias takes 1 generic argument but 2 generic arguments were supplied
  --> crates/apollo-smith/src/argument.rs:37:71
   |
37 |     fn try_from(args_def: apollo_parser::ast::ArgumentsDefinition) -> Result<Self, Self::Error> {
   |                                                                       ^^^^^^       ----------- help: remove this generic argument
   |                                                                       |
   |                                                                       expected 1 generic argument
   |
note: type alias defined here, with 1 generic parameter: `T`
  --> /Users/simon/.cargo/registry/src/github.com-1ecc6299db9ec823/arbitrary-1.1.0/src/error.rs:40:10
   |
40 | pub type Result<T> = std::result::Result<T, Error>;
   |          ^^^^^^ -
```

### Environment

On current `main` branch: commit 5dd19f3

SimonSapin:
For now I’ll keep doing local testing without `--all-features`

goto-bus-stop:
It works with arbitrary-1.2.0. Though i think it's a mistake that we are using `arbitrary::Result` there in the first place.

PR #390, opened 2022-12-19
chore(smith): avoid using arbitrary::Result where std Result is intended

goto-bus-stop:
fixes https://github.com/apollographql/apollo-rs/issues/388

I think renaming or writing out `arbitrary::Result` is better here than aliasing `arbitrary::Result` to the name thats already in the rust prelude. I accidentally used the wrong `Result` type because of the name collision.


lrlna (approved):
seems heaps more clear to me!

PR #391, opened 2022-12-19
Remove UUIDs from the HIR

SimonSapin:
Refering to a particular definition can be done by its name, which is expected to be unique.


lrlna (approved):
ahhhh very nice 😌

PR #394, opened 2022-12-20
Merge `DocumentDatabase` into `HirDatabase`

SimonSapin:
This allows HIR construction for e.g. selection sets to use higher-level queries such as finding a type definition by name (which is now backed by a memoized map instead of linear search)

Relevant query implementations are left in `document.rs` because `hir_db.rs` is already 1.5k lines.


lrlna (approved):
nice nice nice

PR #395, opened 2022-12-20
Replace `hir::Definition` enum with `TypeSystemDefinitions` struct

SimonSapin:
This struct contains one field per kind of definition. Where a name is mandatory, that field is a map keyed by name.

The `type_system_definitions` query now returns this struct instead of `Vec<Definition>`.

The `db_definitions` query has been removed. Instead its callers should use some combination of `type_system_definitions`, `all_operations`, and/or `all_fragments`.

`find_definition_by_name` and `find_type_system_definition_by_name` have also been removed. Instead its callers should used other `*_by_name` queries for more specific kinds of definitions.


lrlna (approved):
thanks!!

Issue #293, opened 2022-09-09
Query parse times increase rapidly with query size


allancalix:
## Description
I noticed that parse times increase with query times more than I'd expect. I suspect that this has to do with the number of allocations happening during lexing/parsing as switching to jemalloc improved things ~15%. I'm not sure if it's something specific to this query or if any query of this size that sees the increase.

### 10 aliased fields
```
parser_peek_n           time:   [21.519 µs 21.535 µs 21.562 µs]                           
                        change: [-97.894% -97.891% -97.887%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### 100 aliased fields
```
Gnuplot not found, using plotters backend
parser_peek_n           time:   [197.55 µs 197.71 µs 197.91 µs]                          
                        change: [+816.92% +818.31% +819.65%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

### 1000 aliased fields
```
Gnuplot not found, using plotters backend
parser_peek_n           time:   [2.0601 ms 2.0618 ms 2.0636 ms]                           
                        change: [+940.25% +941.65% +943.08%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

### Steps to reproduce
1. Parse a large query [see example](https://github.com/allancalix/apollo-rs/blob/parser-fix/crates/apollo-parser/benches/queries/large.graphql)

### Expected result
Parse time for large queries should grow more slowly.

### Actual result
Every 10x increase in number of selections increases parse times by > 800%.

### Environment

- Linux / Darwin
- `apollo-rs` apollo-parser `revision = HEAD`

lrlna:
Interesting! This seems to be an issue specific to the particular Document you're benchmarking. We use apollo-rs on Documents that are MBs in size (as suppose to 57KB in the example you provided), and we are yet to observe such a large spike in parse times. 

I'll need to investigate this a bit more.

allancalix:
It looks like this issue is common for graphql parsers [gqlparser](https://github.com/vektah/gqlparser), async-graphql-parser, and graphql-parser. While I wasn't able to make the scaling of these queries better overall, [reducing the number of allocations in the lexer](https://github.com/apollographql/apollo-rs/compare/main...allancalix:apollo-rs:finite-state-lexer) does improve the performance overall and seems to make it the fastest parsing option of the bunch.

[...]

PR #397, opened 2022-12-21
Adds benchmark for nested aliased query type

allancalix:
https://github.com/apollographql/apollo-rs/issues/293#issue-1367277482

Adds a benchmark for a query identified to parse more slowly than expected given the overall size of the query. Extracted from https://github.com/apollographql/apollo-rs/pull/322 which attempts to improve this benchmark by eliminating string allocations during lexing.

This query is based on a pattern seen in production, I suspect there are GraphQL clients that generate queries with many aliases as a form of aggregation or batching.


lrlna (approved):
thanks for this!!

PR #403, opened 2023-01-05
fix(compiler): use #[salsa::transparent] for all find_ queries

### LINKED ISSUE #373: use #[salsa::transparent] for db.find_ derived queries
It's nice to have the db.find_ queries as part of the same db API, but these don't really need to be cached, so we can attach the `#[salsa::transparent]` attribute to them.

lrlna:
We don't need to store these queries in salsa's cache, so annotate them with `#[salsa::transparent]` and compute them when necessary.

Closes #373

### DIFF
--- a/crates/apollo-compiler/src/database/hir_db.rs
+++ b/crates/apollo-compiler/src/database/hir_db.rs
@@ -61,8 +61,6 @@ pub trait HirDatabase: InputDatabase + AstDatabase {
 
     // Derived from above queries:
     /// Return an operation definition corresponding to the name and file id.
-    /// Result of this query is not cached internally.
-    #[salsa::transparent]
     fn find_operation_by_name(
         &self,
         file_id: FileId,
--- a/crates/apollo-compiler/src/database/hir_db.rs
+++ b/crates/apollo-compiler/src/database/hir_db.rs
@@ -172,8 +172,8 @@ pub trait HirDatabase: InputDatabase + AstDatabase {
     /// - `ObjectType` and `InterfaceType` are subtypes of `Baz`.
     fn subtype_map(&self) -> Arc<HashMap<String, HashSet<String>>>;
 
-    /// Return `true` if the provided `abstract_type` is a subtype of the
-    /// corresponding `maybe_subtype`.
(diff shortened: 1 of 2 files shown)


crates/apollo-compiler/src/database/hir_db.rs
```diff
@@ -34,75 +35,172 @@ pub trait HirDatabase: InputDatabase + AstDatabase {
...
+    /// interface InterfaceType implements Baz { me: String }
+    /// ```
+    /// we can say that:
+    ///
+    /// - `Foo` and `Bar` are a subtypes of `UnionType`.
+    /// - `ObjectType` and `InterfaceType` are subtypes of `Baz`.
     fn subtype_map(&self) -> Arc<HashMap<String, HashSet<String>>>;
 
+    /// Return `true` if the provided `abstract_type` is a subtype of the
+    /// corresponding `maybe_subtype`.
```
SimonSapin, line 176: This sentence seems backwards?

```suggestion
    /// Return `true` if the provided `maybe_subtype` is a subtype of the
    /// corresponding `abstract_type`.
```

SimonSapin (approved):

crates/apollo-compiler/src/database/hir_db.rs
```diff
@@ -34,75 +35,172 @@ pub trait HirDatabase: InputDatabase + AstDatabase {
...
     fn directive_definitions(&self) -> ByName<DirectiveDefinition>;
 
+    /// Return all input object type definitions defined in the compiler.
     fn input_objects(&self) -> ByName<InputObjectTypeDefinition>;
 
     // Derived from above queries:
-
+    /// Return an operation definition corresponding to the name and file id.
+    /// Result of this query is not cached internally.
+    #[salsa::transparent]
```
goto-bus-stop, line 65: @SimonSapin brought up that this is still scanning a Vec internally: https://apollograph.slack.com/archives/C02H6SKKP6Y/p1672929069207239

so we could keep this cached for now?

lrlna: yea sounds good!

goto-bus-stop (approved):

PR #405, opened 2023-01-05
feat(compiler): make `FileId`s unique per-process

SimonSapin:
When we add an alternative kind of compiler input that contains already-parsed HIR nodes, this will allows `FileId` in these nodes to stay meaningful across multiple compiler instances in the same address space.

### DIFF
--- a/crates/apollo-compiler/src/database/sources.rs
+++ b/crates/apollo-compiler/src/database/sources.rs
@@ -1,6 +1,6 @@
 use std::{
     path::{Path, PathBuf},
-    sync::Arc,
+    sync::{atomic, Arc},
 };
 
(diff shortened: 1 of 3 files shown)


crates/apollo-compiler/src/database/sources.rs
```diff
@@ -63,4 +63,17 @@ impl Source {
 }
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
-pub struct FileId(pub u32);
+pub struct FileId {
+    id: u64,
+}
+
+impl FileId {
+    // Returning a different value every time does not sound like good `impl Default`
```
goto-bus-stop, line 71: yep 👍🏻

goto-bus-stop (approved):

lrlna (approved):

PR #410, opened 2023-01-09
feat(compiler): add type alias for `compiler.snapshot()`

SimonSapin:
This allows naming the return type of `ApolloCompiler::snapshot` without importing `salsa` separately.

### DIFF
--- a/crates/apollo-compiler/src/lib.rs
+++ b/crates/apollo-compiler/src/lib.rs
@@ -19,7 +19,7 @@ pub struct ApolloCompiler {
 }
 
 /// A read-only, `Sync` snapshot of the database.
-type Snapshot = salsa::Snapshot<RootDatabase>;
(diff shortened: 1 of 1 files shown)


crates/apollo-compiler/src/lib.rs
```diff
@@ -18,6 +18,9 @@ pub struct ApolloCompiler {
     pub db: RootDatabase,
 }
 
+/// A read-only, `Sync` snapshot of the database.
+type Snapshot = salsa::Snapshot<RootDatabase>;
```
goto-bus-stop, line 22: should this be pub then?
```suggestion
pub type Snapshot = salsa::Snapshot<RootDatabase>;
```

SimonSapin: 🤦 

thanks!

goto-bus-stop (approved):
thanks!

PR #414, opened 2023-01-10
feat(compiler): rework diagnostics to support multiple files

goto-bus-stop:
`miette` doesn't support referencing multiple files in one error, so we sadly can't use it directly. Most errors in executable documents will reference another document.

Without miette's derives, we need some other way to represent shared diagnostic properties like labels and help text. For now this PR uses:
```rust
pub struct ApolloDiagnostic {
    pub location: HirNodeLocation, // ← may be removed as the first Label will usually have the same location
    pub labels: Vec<Label>,
    pub help: Option<String>,
    pub data: DiagnosticData,
}
```
`DiagnosticData` is an enum with the different diagnostic types and specific information about them.
With that structure we need to format labels wherever we create an error.

I converted a part of the validation so far and explored using `ariadne` for displaying errors. I'll continue to work on `ariadne` output first in case the diagnostic data structure needs to change. Then some API polish, maybe we can reduce the boilerplatyness of this implementation a bit.

Todos:
- [x] We need an `ariadne` release (using a fork)
- [x] Tests now rely on the order they are run in to work, due to the global FileId counter -- they should be checked some other way. Maybe `ApolloDiagnostic` Debug impl can censor them consistently or we just make `FileId` display as a constant value … or we change the test output to use source strings wherever a `DiagnosticLocation` occurs … or something else
    #437


goto-bus-stop:
Something still wrong in printing the line/column offsets, but this is an example of a multi-file diagnostic:
image

lrlna:
ooooohh that is already looking real good @goto-bus-stop !

goto-bus-stop:
OK, we're really getting somewhere:
image

Time to resolve some conflicts ;p

goto-bus-stop:
This can be reviewed but it still has some blockers, will maintain checklist in OP

crates/apollo-compiler/src/diagnostics.rs
```diff
@@ -1,400 +1,265 @@
...
+        // Else just remove as the labeling is done separately.
+        fields: usize,
+        subscription: DiagnosticLocation,
+    },
+    #[error("{ty} root operation type is not defined")]
+    UnsupportedOperation {
+        // current operation type: subscription, mutation, query
+        ty: &'static str,
+    },
+    #[error("Cannot query `{field}` field")]
```
lrlna, line 159:
```suggestion
    #[error("cannot query `{field}` field")]
```

crates/apollo-compiler/src/diagnostics.rs
```diff
@@ -1,400 +1,265 @@
...
+        original_definition: DiagnosticLocation,
+        redefined_definition: DiagnosticLocation,
+    },
+    #[error("the argument `{name}` is defined multiple times")]
+    UniqueArgument {
+        name: String,
+        original_definition: DiagnosticLocation,
+        redefined_definition: DiagnosticLocation,
+    },
+    #[error("Subscriptions operations can only have one root field")]
```
lrlna, line 147:
```suggestion
    #[error("subscription operations can only have one root field")]
```

crates/apollo-compiler/src/diagnostics.rs
```diff
@@ -1,400 +1,265 @@
...
+        // field name
+        name: String,
+        // field type
+        ty: &'static str,
+    },
+    #[error(
+        "Custom scalars should provide a scalar specification URL via the @specifiedBy directive"
+    )]
+    ScalarSpecificationURL,
+    #[error("Missing query root operation type in schema definition")]
```
lrlna, line 205:
```suggestion
    #[error("missing query root operation type in schema definition")]
```

crates/apollo-compiler/src/diagnostics.rs
```diff
@@ -1,400 +1,265 @@
...
+    QueryRootOperationType,
+    #[error("Built-in scalars must be omitted for brevity")]
```
lrlna, line 207:
```suggestion
    #[error("built-in scalars must be omitted for brevity")]
```

lrlna (commented):
some error standardisation here

goto-bus-stop:
so at this stage it works and everything, but we have less context in error messages.

lrlna:
@goto-bus-stop ok, let's merge this in, and in our continued validation process work on adding more context to ariadne

lrlna (approved):

PR #415, opened 2023-01-10
breaking(compiler): unifying ty, type_def, kind naming in the HIR

### LINKED ISSUE #393: HIR APIs named `ty` refer to different things
For example:

* `FieldDefinition::ty` returns a GraphQL type like `[Foo!]`
* `Field::ty` returns a GraphQL type definition like `interface Foo { … }`
* `Definition::ty` returns the name of a Rust type for a kind of GraphQL definition like  `FragmentDefinition` (not necessarily a GraphQL type)

I think it’d be worth finding all instances of `fn ty` a reconsider naming. The cases above could be  `ty`, `type_def` and `kind` respectively

lrlna:
fixes #393 

Clarifies our `ty` names in the HIR, otherwise known as adding all the breaking changes to `0.6.0`.

- for `Type` struct impl: `ty()` --> `type_def()`, as it returns the `TypeDefinition` the `Type` is referencing.
- for `TypeDefinition` struct impl: `ty()` --> `kind()`, as it returns a Type Definition kind.
- for `FragmentDefinition` struct impl: `ty()` --> `type_def()`, as it returns the `TypeDefinition` the `FragmentDefinition` is referencing
- for `OperationDefinition` struct: `operation_ty` field --> `ty` field. "operation" is redundant, as the struct is already named "operation"
- for `RootOperationTypeDefinition` struct: `operation_type` field --> `ty` field. "operation" is redundant, as the struct already contains "operation"

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -188,7 +188,7 @@ pub struct OperationDefinition {
 }
 
 impl OperationDefinition {
-    /// Get a reference to the operation definition's ty.
+    /// Get the type of the operation: Query, Mutation, or Subscription
     pub fn ty(&self) -> &OperationType {
         &self.ty
     }
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -179,7 +179,7 @@ impl FragmentDefinition {
 
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub struct OperationDefinition {
-    pub(crate) operation_ty: OperationType,
+    pub(crate) ty: OperationType,
     pub(crate) name: Option<Name>,
     pub(crate) variables: Arc<Vec<VariableDefinition>>,
     pub(crate) directives: Arc<Vec<Directive>>,
@@ -189,8 +189,8 @@ pub struct OperationDefinition {
 
 impl OperationDefinition {
     /// Get a reference to the operation definition's ty.
-    pub fn operation_ty(&self) -> &OperationType {
-        &self.operation_ty
+    pub fn ty(&self) -> &OperationType {
+        &self.ty
     }
 
     /// Get a mutable reference to the operation definition's name.
@@ -205,7 +205,7 @@ impl OperationDefinition {
 
     /// Get operation's definition object type.
     pub fn object_type(&self, db: &dyn HirDatabase) -> Option<Arc<ObjectTypeDefinition>> {
-        match self.operation_ty {
+        match self.ty {
             OperationType::Query => db.schema().query(db),
             OperationType::Mutation => db.schema().mutation(db),
             OperationType::Subscription => db.schema().subscription(db),
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -411,7 +411,7 @@ impl Type {
     /// [`EnumTypeDefinition`]: Definition::EnumTypeDefinition
     #[must_use]
     pub fn is_output_type(&self, db: &dyn HirDatabase) -> bool {
-        if let Some(ty) = self.ty(db) {
+        if let Some(ty) = self.type_def(db) {
             ty.is_output_definition()
         } else {
             false
@@ -426,7 +426,7 @@ impl Type {
     /// [`InputObjectTypeDefinition`]: Definition::ObjectTypeDefinition
     #[must_use]
     pub fn is_input_type(&self, db: &dyn HirDatabase) -> bool {
-        if let Some(ty) = self.ty(db) {
+        if let Some(ty) = self.type_def(db) {
             ty.is_input_definition()
         } else {
             false
@@ -442,7 +442,7 @@ impl Type {
         }
     }
 
-    pub fn ty(&self, db: &dyn HirDatabase) -> Option<TypeDefinition> {
+    pub fn type_def(&self, db: &dyn HirDatabase) -> Option<TypeDefinition> {
         db.find_type_definition_by_name(self.name())
     }
 
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -54,7 +54,7 @@ impl TypeDefinition {
             Self::InputObjectTypeDefinition(def) => def.name_src(),
         }
     }
-    pub fn ty(&self) -> &'static str {
+    pub fn kind(&self) -> &'static str {
         match self {
             Self::ScalarTypeDefinition(_) => "ScalarTypeDefinition",
             Self::ObjectTypeDefinition(_) => "ObjectTypeDefinition",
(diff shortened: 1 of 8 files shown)


crates/apollo-compiler/src/database/hir.rs
```diff
@@ -189,8 +189,8 @@ pub struct OperationDefinition {
 
 impl OperationDefinition {
     /// Get a reference to the operation definition's ty.
```
goto-bus-stop, line 191: a clearer doc comment could also work to separate this type of `ty` from `named_type`
```suggestion
    /// Get the type of the operation: Query, Mutation, or Subscription
```

goto-bus-stop (commented):
I think `operation_ty(pe)` could be worth keeping, as it is something different from most `ty` methods, and it more clearly disambiguates it from the `named_type()` methods also available on operations. i feel like the best name would actually be operation _kind_ but that's not the language the spec uses sadly 🙈 

I dont think that's a blocker, just want to float it as an idea. but if we use the `ty()` name then we should document precisely what type of type (heh) it returns

lrlna:
> I think operation_ty(pe) could be worth keeping, as it is something different from most ty methods, and it more clearly disambiguates it from the named_type() methods also available on operations. i feel like the best name would actually be operation kind but that's not the language the spec uses sadly 🙈

@goto-bus-stop I am less on the fence about `operation_ty`, can keep it as is

SimonSapin:
> i feel like the best name would actually be operation _kind_ but that's not the language the spec uses sadly :see_no_evil: 

Except it is?

https://spec.graphql.org/October2021/#sel-GAHTRBBBFBTzuP
> A schema defines the initial ***root operation type*** for ***each kind of operation*** it supports: query, mutation, and subscription […]
>
> The `query` root operation type must be provided and must be an Object type.

(Emphasis mine)

My reading is that this:

```rust
enum ??? {
    Query,
    Mutation,
    Subscription,
}
```

… is called "operation kind" by the spec, whereas a "root operation type" refers to a specific *object* type.

---

By the way, the `named_type: Type` field of `RootOperationTypeDefinition` seems to allow things like `[SomeInterface!]` while the spec doesn’t. I think it should be `object_type_name: String` instead

goto-bus-stop:
Iiiiinteresting! I was looking at this: https://spec.graphql.org/October2021/#OperationDefinition it seems like they use both terms for that.

SimonSapin:
Ideally I think the spec should be changed to use `OperationKind` in the grammar and "kind" in prose, keeping "type" for things in the type system

> * for `Type` struct impl: `ty()` --> `type_def()`, as it returns the `TypeDefinition` the `Type` is referencing.
> * for `TypeDefinition` struct impl: `ty()` --> `kind()`, as it returns a Type Definition kind.
> * for `FragmentDefinition` struct impl: `ty()` --> `type_def()`, as it returns the `TypeDefinition` the `FragmentDefinition` is referencing

these lgtm

lrlna:
The official name in the grammar for `query`, `mutation` and `subscription` is `OperationType`. This is how [`RootOperationTypeDefinition`](https://spec.graphql.org/October2021/#RootOperationTypeDefinition) and [`OperationDefinition`](https://spec.graphql.org/October2021/#OperationDefinition) define their grammar. For us rust people it's definitely semantically a `kind` and should probably be changed in the spec >.<

> By the way, the named_type: Type field of RootOperationTypeDefinition seems to allow things like [SomeInterface!] while the spec doesn’t. I think it should be object_type_name: String instead

no, the node name in the spec is [`named_type`](https://spec.graphql.org/October2021/#sel-EAHTFEBAAAFCCAAJu6X), but I ought to specify `Type::Named` for its return type to clarify this, you're right. `RootOperationTypeDefinition` cannot have `[SomeInterface!]`, as that would be of `Type::List`, which is not allowed by the spec for this definition.

crates/apollo-compiler/src/database/hir.rs
```diff
@@ -188,9 +188,9 @@ pub struct OperationDefinition {
 }
 
 impl OperationDefinition {
-    /// Get a reference to the operation definition's ty.
-    pub fn operation_ty(&self) -> &OperationType {
-        &self.operation_ty
+    /// Get the type of the operation: Query, Mutation, or Subscription
+    pub fn ty(&self) -> &OperationType {
+        &self.ty
```
goto-bus-stop, line 193: tangentially:
```suggestion
    pub fn ty(&self) -> OperationType {
        self.ty
```

lrlna: should be done

SimonSapin:
It looks like `schema { query: [SomeInterface!] }` is allowed by the grammar, but then validation comes in: 

> The query root operation type must be provided and must be an Object type.

So IMO it comes down to: do we want states accepted by the grammar but otherwise invalid to be representable in HIR?

---

> ah well enum variant types aren't a thing (yet)!

I’m not sure they’ll ever be (or that they should). This leaves storing the fields of that variants, in this case `name: String, loc: Option<HirNodeLocation>`. Arguably `loc` is not useful (since the `RootOperationTypeDefinition` as a whole already stores a location), which leaves `object_type_name: String`

lrlna:
> It looks like `schema { query: [SomeInterface!] }` is allowed by the grammar, but then validation comes in:
> 
> > The query root operation type must be provided and must be an Object type.
> 
> So IMO it comes down to: do we want states accepted by the grammar but otherwise invalid to be representable in HIR?

that type system will actually panic at the moment - we look for a `node.name()` when creating root operation definition type, and a ListType doesn't have it. This particular discussions is largely part of #256 - we currently don't properly handle absence of a given node that resulted in a syntax error from the parsing step. We've definitely had a few discussions of perhaps having another level of intermediate representation that could perhaps come in after a validation step is complete that only valid syntax. This is a discussion for a different PR, however.

Validation for root operation type being an object type is reflected in #419

SimonSapin:
> This is a discussion for a different PR, however.

Yes, sorry for derailing!

crates/apollo-compiler/src/database/hir.rs
```diff
@@ -188,9 +188,9 @@ pub struct OperationDefinition {
 }
 
 impl OperationDefinition {
-    /// Get a reference to the operation definition's ty.
-    pub fn operation_ty(&self) -> &OperationType {
-        &self.operation_ty
+    /// Get the type of the operation: Query, Mutation, or Subscription
```
lrlna, line 191:
```suggestion
    /// Get the kind of the operation: `query`, `mutation`, or `subscription`
```

crates/apollo-compiler/src/database/hir.rs
```diff
@@ -1184,10 +1186,11 @@ impl RootOperationTypeDefinition {
     }
 
     /// Get the root operation type definition's operation type.
```
lrlna, line 1188:
```suggestion
    /// Get the kind of the root operation type definition: `query`, `mutation`, or `subscription`
```

SimonSapin (approved):

Issue #246, opened 2022-06-09
validation: directive location usage


lrlna:
Let's say a directive is defined on field definition and argument definition:
```
directive @example on FIELD_DEFINITION | ARGUMENT_DEFINITION
```

```[tasklist]
### Tasks
- [x] validation should ensure that whenever `@example` directive is used, it's actually being used in a field and argument definitions only. [spec](https://spec.graphql.org/October2021/#sec-Directives-Are-In-Valid-Locations)
- [ ] https://github.com/apollographql/apollo-rs/pull/488
  directives must be unique per location, [spec](https://spec.graphql.org/October2021/#sec-Directives-Are-Unique-Per-Location) 
```

PR #417, opened 2023-01-11
feat(compiler): validate directive locations + undefined directives

lrlna:
partial of #246

This looks over each directive usage and checks whether it's in a correct location as specified by its original DirectiveDefinition. In case of incorrect usage `UnsupportedLocation` error is returned. Since we are already looking for the original DirectiveDefinition, this PR also adds a `UndefinedDefinition` error for undefined directives in case that DirectiveDefinition is not found.

While working on this validation rule, I realised we need to tidy up our ValidationDatabase before we implement any follow up rules. Primarily we are using two different paths for validating: `check_*` and `validate_*`, which should be unified in some way. 
We are also creating a lot of `let diagnostics = Vec::new()` for pretty much each validation step, and it would be good to not keep allocating new vecs for each node as we add diagnostics.

### DIFF
--- a/crates/apollo-compiler/src/validation/validation_db.rs
+++ b/crates/apollo-compiler/src/validation/validation_db.rs
@@ -8,7 +8,7 @@ use miette::SourceSpan;
 use crate::{
     database::db::Upcast,
     diagnostics::{UndefinedDefinition, UniqueArgument, UnsupportedLocation},
-    hir::{self, DirectiveLocation},
+    hir,
     validation::{
         directive, enum_, input_object, interface, object, operation, scalar, schema, union_,
         unused_variable,
@@ -63,6 +63,7 @@ pub trait ValidationDatabase:
         &self,
         def: Arc<hir::EnumTypeDefinition>,
     ) -> Vec<ApolloDiagnostic>;
+    fn check_enum_value(&self, enum_val: hir::EnumValueDefinition) -> Vec<ApolloDiagnostic>;
     fn check_input_object_type_definition(
         &self,
         def: Arc<hir::InputObjectTypeDefinition>,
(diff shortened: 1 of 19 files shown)


lrlna:
sorry for 5000+ line additions! i swear most of these are tests 😅

crates/apollo-compiler/src/validation/validation_db.rs
```diff
@@ -135,50 +156,82 @@ pub fn check_union_type_definition(
...
+    input_obj: Arc<hir::InputObjectTypeDefinition>,
 ) -> Vec<ApolloDiagnostic> {
     let mut diagnostics = Vec::new();
 
-    // TODO: validate extensions
-    for directive in schema_def.directives() {
-        diagnostics.extend(db.check_directive(directive.clone()));
-    }
+    diagnostics.extend(db.check_input_values(
+        (*input_obj).clone().input_fields_definition,
```
goto-bus-stop, line 188: only need to clone the inner thing
```suggestion
        input_obj.input_fields_definition.clone(),
```

goto-bus-stop (approved):
> We are also creating a lot of `let diagnostics = Vec::new()` for pretty much each validation step, and it would be good to not keep allocating new vecs for each node as we add diagnostics.

I think in this tree-walk design, it's almost certainly going to be faster with standalone functions that can borrow hir nodes and push into one shared diagnostics vec, even if it has to rerun the entire validation any time something changes. (compared to having the many cacheable queries with owned values everywhere). maaaybe it wouldn't make as much of a difference if we wrap everything in Arcs but i don't really have a good idea of how expensive atomic operations are.

this looks good tho 😄

lrlna:
> > We are also creating a lot of `let diagnostics = Vec::new()` for pretty much each validation step, and it would be good to not keep allocating new vecs for each node as we add diagnostics.
> 
> I think in this tree-walk design, it's almost certainly going to be faster with standalone functions that can borrow hir nodes and push into one shared diagnostics vec, even if it has to rerun the entire validation any time something changes. (compared to having the many cacheable queries with owned values everywhere). maaaybe it wouldn't make as much of a difference if we wrap everything in Arcs but i don't really have a good idea of how expensive atomic operations are.
> 
> this looks good tho 😄

yea definitely something for us to think about once we simplify this validation a little bit!

Issue #312, opened 2022-09-21
validation: variables


lrlna:
- [x] variables in operation definitions must be unique, [spec](https://spec.graphql.org/October2021/#sec-Variable-Uniqueness)
- [x] variables must be of input type, [spec](https://spec.graphql.org/October2021/#sec-Variables-Are-Input-Types)
- [ ] all varaible usages are allowed, [spec](https://spec.graphql.org/October2021/#sec-All-Variable-Usages-are-Allowed)

PR #422, opened 2023-01-16
feat(compiler): validate variable definition uniqueness

lrlna:
trying out the new diagnostics changes from #414 !

@goto-bus-stop so far so good! process seems to be just as straightforward as before 🙏 

I am now noticing the thing you pointed out this morning about fileids in diagnostics struct - basically all the tests get updated whenever you add a new file, which means there are a ton of diffs.


lrlna:
gonna close this and do all the variable validations (#312) all together in one PR (and not only because i don't want to deal with all these conflicts >.<)

Issue #426, opened 2023-01-17
Infinite loop involving `apollo_smith::DocumentBuilder::limited_string`


SimonSapin:
## Description

For reasonable-looking inputs, apollo-smith seems to get into an infinite loop.

### Steps to reproduce

Add `crates/apollo-smith/tests/with_document.rs` with:

```rust
#[cfg(feature = "parser-impl")]
#[test]
fn test_with_document() {
    let mut u = arbitrary::Unstructured::new(&[1]);
    let schema = "schema { query: Query } type Query { id: ID! }";
    let schema = apollo_parser::Parser::new(schema)
        .parse()
        .document()
        .try_into()
        .unwrap();
    apollo_smith::DocumentBuilder::with_document(&mut u, schema)
        .unwrap()
        .operation_definition()
        .unwrap();
}
```

Run `cargo test -p apollo-smith --all-features`.

### Expected result

The test ends in a short time and passes.

### Actual result

The test does not end before I get bored and stop it. A few seconds of it in a sampling profiling shows that a lot of time is spent in `apollo_smith::DocumentBuilder::limited_string`:

<img width="1133" alt="Screenshot 2023-01-17 at 13 59 53" src="https://user-images.githubusercontent.com/291359/212905444-8bcd4041-9e46-43c2-bd52-eb30ceb9066a.png">

### Environment

- `apollo-rs` crate: apollo-smith
- Crate version: current `main`: 3ad63370eadf9e01c233d0891da3215bca015e84

SimonSapin:
Indeed, `int_in_range` starts with zero and exits its own loop when running out of input:

https://github.com/rust-fuzz/arbitrary/blob/d38b015af522ca1b43c14d4765f7eeb27545c00a/src/unstructured.rs#L335-L342

SimonSapin:
This behavior appears to be deliberate:

Code change: https://github.com/rust-fuzz/arbitrary/commit/38192741b9a1f14424106545f4b3b1e56fb0a3d3
Reasoning: https://github.com/rust-fuzz/arbitrary/commit/1d2f65360f634968b5525f4d6430a88a376277f0

PR #427, opened 2023-01-17
Fix an infinite loop in apollo-smith

SimonSapin:
It can happen very easily when running out of unstructured input.

Fixes https://github.com/apollographql/apollo-rs/issues/426


crates/apollo-smith/src/name.rs
```diff
@@ -90,7 +90,7 @@ impl<'a> DocumentBuilder<'a> {
     // Mirror what happens in `Arbitrary for String`, but do so with a clamped size.
     pub(crate) fn limited_string(&mut self, max_size: usize) -> ArbitraryResult<String> {
         loop {
-            let size = self.u.int_in_range(0..=max_size)?;
+            let size = self.u.int_in_range(1..=max_size)?;
```
SimonSapin, line 93: An empty result string is rejected below and causes this loop to start over, so let’s ask for a non-zero size

lrlna (approved):

PR #430, opened 2023-01-17
feat(smith): add DocumentBuilder::input_exhausted

SimonSapin:
This indicates we might have run into https://github.com/rust-fuzz/arbitrary/issues/139

### DIFF
--- a/crates/apollo-smith/src/lib.rs
+++ b/crates/apollo-smith/src/lib.rs
@@ -207,6 +207,11 @@ impl<'a> DocumentBuilder<'a> {
         Ok(builder)
     }
 
+    /// Returns whether the provided `Unstructured` is now empty
+    pub fn input_exhausted(&self) -> bool {
+        self.u.is_empty()
+    }
+
     /// Convert a `DocumentBuilder` into a GraphQL `Document`
     pub fn finish(self) -> Document {
         Document {


SimonSapin:
My current attempt at working around https://github.com/rust-fuzz/arbitrary/issues/139 is to double the size of the input until it’s large enough to not be exhausted, but currently this causes `DocumentBuilder::operation_definition` to overflow the stack, I suspect through generating very deeply nested selection sets.

SimonSapin:
By the way, at first this PR seems unnecessary because `DocumentBuilder` only borrows `&mut Unstructured<'_>` so after it is dropped the called should be able to call `Unstructured::is_empty` themselves. However because `DocumentBuilder` only has one lifetime parameter and borrows `&'a mut Unstructured<'a>` with the same `'a` in two places. Because something something variance, `Unstructured` is still considered borrowed when we try to use it. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=193bdb3f0ffae931f060751c39ef9087 is a reduced test case.

The "proper" fix would be:

```rust
struct DocumentBuilder<'a, 'b> {
    u: &'a mut Unstructured<'b>,
```

… but that changes the signature of the type which would need to be updated at every use, including each impl

lrlna (approved):

PR #435, opened 2023-01-20
chore(compiler): use Display impls for OperationType, DirectiveLocation

goto-bus-stop:
Noticed a small improvement opportunity while merging `main` into #414…

`OperationType` already has a `Display` impl, so we don't actually need a separate `Into<String>`: users can do `ty.to_string()` and it will use `Display`.

`DirectiveLocation` also has a `Into<String>` impl. If we use a `Display` impl instead, we can print `DirectiveLocation`s directly in format strings. Then we can also just stick the `DirectiveLocation` value into our diagnostics, instead of stringifying them first, which makes the diagnostic more programmatically useful.

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::{HashMap, HashSet},
+    fmt,
     sync::Arc,
 };
 
@@ -307,24 +308,12 @@ impl OperationType {
     }
 }
 
-impl std::fmt::Display for OperationType {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+impl fmt::Display for OperationType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             OperationType::Query => write!(f, "Query"),
             OperationType::Mutation => write!(f, "Mutation"),
             OperationType::Subscription => write!(f, "Subscription"),
-        }
-    }
-}
-
-impl From<OperationType> for String {
-    fn from(op_type: OperationType) -> Self {
-        if op_type.is_subscription() {
-            "Subscription".to_string()
-        } else if op_type.is_mutation() {
-            "Mutation".to_string()
-        } else {
-            "Query".to_string()
         }
     }
 }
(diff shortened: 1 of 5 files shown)


crates/apollo-compiler/src/diagnostics.rs
```diff
@@ -410,12 +411,12 @@ pub struct UnsupportedLocation {
...
 
     // current location where the directive is used
-    pub dir_loc: String,
+    pub dir_loc: DirectiveLocation,
 
     #[source_code]
     pub src: Arc<str>,
 
-    #[label("{} is not a valid location", self.dir_loc)]
+    #[label("{} is not a valid location for this directive", self.dir_loc)]
```
goto-bus-stop, line 419: didn't catch this in review earlier but I think this message could be improved a bit … maybe something like "cannot use directive `@xyz` here"

crates/apollo-compiler/test_data/diagnostics/0050_directives_in_invalid_locations.txt
```diff
@@ -2,7 +2,7 @@
     UnsupportedLocation(
         UnsupportedLocation {
             ty: "skip",
-            dir_loc: "QUERY",
+            dir_loc: Query,
```
goto-bus-stop, line 5: this changed only because it uses the Debug impl. The actual error messages displayed to the user are still all-caps-with-underscores

crates/apollo-compiler/src/diagnostics.rs
goto-bus-stop: I guess the thing is that the label is more visually prominent than the "main" error message, despite our "main" message being the thing that contains the important information
image

lrlna: oh yea this is nicer ty!

crates/apollo-compiler/src/database/hir.rs
```diff
@@ -317,18 +318,6 @@ impl std::fmt::Display for OperationType {
...
-        if op_type.is_subscription() {
-            "Subscription".to_string()
-        } else if op_type.is_mutation() {
-            "Mutation".to_string()
-        } else {
-            "Query".to_string()
-        }
-    }
-}
-
```
lrlna, line 331: i think removing this is technically a breaking change eh?

lrlna (approved):
looks good, thank you!

crates/apollo-compiler/src/database/hir.rs
goto-bus-stop: Yes you can no longer do `op_type.into()`, you have to do `op_type.to_string()` instead. So very minor but breaking nonetheless

PR #436, opened 2023-01-20
chore(compiler): simplify validation database

lrlna:
This change simplifies the validation database to (try) to walk the tree as much as it's possible. We are able to take advantage of salsa's caching for a lot of the queries, so when the input to a validation rule doesn't change, we don't have to re-validate.


lrlna:
@goto-bus-stop i think it's at a good stage for feedback! Would love to hear what you think about this approach so far.

I wanted to have as much validation as it's currently possible in salsa queries which should be helpful for re-running validation on updated inputs. I'd love to still have a bit more caching for all the queries that current iterate on plurals: operations, objects, interfaces etc, but we gotta make our `ByName` IndexMap a `Hash` first. Figured we could do this at a later stage.

There are still quite a few things wrong that I have to fix, especially with object, interface and operation validations. And I think I should probably deal with all the new vecs all over the place now rather than later

goto-bus-stop:
I like the overall direction, I think if we also Arc a lot of stuff here to reduce the cost of cloning, we'll be good to go. Though we may want to just merge it asap as everything else will conflict with this and address that in followup PRs.

TBH making `IndexMap` impl `Hash` sounds wrong, is it because we want to use it as an input somewhere? can we not instead retrieve the `ByName` maps with db queries wherever needed?

PR #421, opened 2023-01-15
feat(compiler): Add is_introspection helper method

jregistr:
Add is_introspection helper method to to OperationDefinition.

    Closes #396

Example usage:
```rust
let query_input = r#"
query Op {
  __type(name: "User") {
    name
  }
}
"#;

let mut compiler = ApolloCompiler::new();
let query_id = compiler.add_executable(query_input, "query.graphql");
let db = compiler.db;
let introspect_query_op: Arc<OperationDefinition> = db
            .find_operation_by_name(query_id, String::from("TypeIntrospect"))
            .expect("TypeIntrospect operation does not exist");

let is_introspection = introspect_query_op.is_introspection(&db);
```

PR #438, opened 2023-01-24
feat(compiler): Add validation for introspection fields in subscription operations

### LINKED ISSUE #305: validation: subscription operation must not have an introspection field in selection set
check that fields in subscription selection set are not introspection fields. [spec](https://spec.graphql.org/October2021/#sel-FALPHDGDDLAAADFBA04c)

gocamille:
### Description:
This PR adds validation for introspection fields in subscription operations in the apollo-compiler library. The code checks that the root field of a subscription selection is not an introspection field, and returns validation errors if the root field is an introspection field.

Fixes #305 .


crates/apollo-compiler/src/validation/subscription.rs
```diff
@@ -0,0 +1,57 @@
...
+                }
+            }
+        }
+    }
+
+    diagnostics
+}
+
+#[cfg(test)]
+
```
gocamille, line 33: @lrlna I kept this module-level test in the module, but should it be removed for the sake of convention and consistency?

crates/apollo-compiler/test_data/diagnostics/0050_subscription_operation_with_root_introspection_field.graphql
```diff
@@ -0,0 +1,3 @@
+subscription sub {
+  __typename
+}
```
lrlna, line 3: Let's add a subscription type to this document so we don't also get a `UnsupportedOperation` diagnostic in our output

```suggestion
subscription sub {
  __typename
}

type Subscription {
  __typename: String
}
```

crates/apollo-compiler/src/validation/subscription.rs
lrlna: yes, let's keep it these sort of tests in `test_data` directory.

crates/apollo-compiler/src/validation/subscription.rs
```diff
@@ -0,0 +1,57 @@
...
+    let mut diagnostics = Vec::new();
+    let subscription_operations = db.upcast().subscription_operations(file_id);
+
+    // Subscription fields must not have an introspection field in the selection set.
+    //
+    // Return an Introspection Field error in case of an introspection field existing as the root field in the set.
+    for op in subscription_operations.iter() {
+        for selection in op.selection_set().selection() {
+            if let Selection::Field(field) = selection {
+                if field.name().starts_with("__") {
```
lrlna, line 15: it's only a convention to keep `__` for introspection fields only, so theoretically someone can just have a field that starts with `__` in their schema. It'd be best to check for `__typename`, `__schema` and `__type` directly. 

If you rebase from main, we've just merged #421, so you should be able to use `field.is_introspection()` here.

crates/apollo-compiler/src/diagnostics.rs
```diff
@@ -425,3 +428,20 @@ pub struct UnsupportedLocation {
...
+
+#[derive(Diagnostic, Debug, Error, Clone, Hash, PartialEq, Eq)]
+#[error("Subscription operations can not have an introspection field as a root field")]
+#[diagnostic(code("apollo-compiler validation error"))]
+pub struct IntrospectionField {
+    // current definition
+    pub field: String,
+
+    #[help]
+    pub help: Option<String>,
```
lrlna, line 440: i think `help` field is not necessary here. The `error` above describes the problem, and the label for definition describes what's wrong with this field.

crates/apollo-compiler/src/diagnostics.rs
```diff
@@ -425,3 +428,20 @@ pub struct UnsupportedLocation {
     #[help]
     pub help: Option<String>,
 }
+
+#[derive(Diagnostic, Debug, Error, Clone, Hash, PartialEq, Eq)]
+#[error("Subscription operations can not have an introspection field as a root field")]
```
lrlna, line 433: going to try to stick with Rust's error messaging conventions, and keep things lower cased 😅 

```suggestion
#[error("subscription operations can not have an introspection field as a root field")]
```

lrlna (commented):
This is looking good!! I'd rebase from main, as we just merged #421, which will be helpful for this particular validation rule.

crates/apollo-compiler/src/validation/subscription.rs
gocamille: Ah, good to know! I'll update this and use that great `field.is_introspection()` method 👍

Issue #392, opened 2022-12-20
validation: apollo compiler validation ignore non existing fields


yanns:
## Description

I'm trying to use apollo-compiler to validate queries against schema like in https://github.com/apollographql/router/issues/1116#issuecomment-1357342184

The validation does not fail on non-existing fields.

### Steps to reproduce

Considering the following:
```rust
        let input = indoc! {r#"
            type Review {
              id: Int!
            }

            type Query {
              reviews: [Review!]!
            }

            {
              reviews {
                id
                key
              }
            }
        "#};

        let ctx = ApolloCompiler::new(input);
        let diagnostics = ctx.validate();
        for diagnostic in &diagnostics {
            println!("{}", diagnostic);
        }
        assert!(diagnostics.is_empty());
```

### Expected result

`diagnostics` should not be empty.

### Actual result

`diagnostics` is empty.

### Environment

- `apollo-rs` crate: `apollo-compiler`
- Crate version: `0.4.1`

lrlna:
@erikwrede would you like to make a PR to the compiler for this?

erikwrede:
@lrlna sure! Might have some time to work on this next week.
I want to try out using rust's `traverse` crate first though because it uses a stack-based approach and evaluate if there's any benefits from eliminating recursion for longer/more nested queries. If not, I'll keep the recursive approach for now (which is what most of the reference implementations do). 

Would probably limit the scope to ObjectTypes without InlineFragments for now so we can evaluate the approach first.

lrlna:
@erikwrede - no pressure at all! You've already put together a gist, I figured it would be good to merge that in :)

erikwrede:
@lrlna pushed a draft to #441

yanns:
Missing:
- https://github.com/apollographql/apollo-rs/issues/306
- https://github.com/apollographql/apollo-rs/issues/307

For the last one, https://github.com/apollographql/apollo-rs/pull/465

lrlna:
this is now published as part of the new releases -- `cargo add apollo-compiler@0.7.1`.

Issue #439, opened 2023-01-24
meta fields in compiler's context


lrlna:
For field validation, especially for validation involving introspection field types, it'd be useful to figure out how we should add implicit fields to compiler's context. So far, implicit directives and scalars are added to the HIR as the HIR gets created. These type don't have an AST associated with them, so in validation we must check whether or not a directive or scalar node has a location before proceeding to create a diagnostic. Extending the same logic to other meta fields like introspection, means fields (`__typename`, `__type`, `__schema`) and types (`__Field`, `__InputValue` etc) now also exist in the context and have no AST or location information associated with them. This will make validation rule implementation a bit verbose. It would be good to figure out a nice, ergonomic way of dealing with this virtual/meta nodes.

I had this idea of having a virtual file that could have all introspection types, implicit directives and scalars, but I am not sure if this will end up being a lot more work than `Option<HirNodeLocation>`

SimonSapin:
I ran into this trying to use `hir::Field::ty` instead of apollo-router’s own algorithm for resolving types in selections. A test has this query:

```graphql
      {
        __type(name: \"Bar\") {
          name
          fields {
            name
            type {
              name
            }
          }
        }
      }
```

Because apollo-compiler is not aware of `__type`, for its sub-fields `hir::Field::parent_obj` is `None`.

yanns:
Let me know if it's related:
when using `ApolloCompiler` with an introspection query, `compiler.validate()` returns an error:

```
Error: cannot query `__schema` field
    ╭─[query.graphql:3:7]
    │
  3 │ ╭─▶       __schema {
    ┆ ┆   
 20 │ ├─▶       }
    │ │             
    │ ╰───────────── `__schema` field is not defined
    │     
    │     Help: `__schema` is not defined on `Query` type
────╯
```
I guess this is because those introspection fields are not known by `ApolloCompiler`.

lrlna:
@yanns yep, that's the same issue!

yanns:
@lrlna do you have a concrete idea on how to tackle this?
I'd be ready to try this out.

lrlna:
@yanns yes, we've got a path forward in #489.

Issue #307, opened 2022-09-21
validation: leaf field selection


lrlna:
implement leaf field selection [spec](https://spec.graphql.org/October2021/#sec-Leaf-Field-Selections)

yanns:
I've opened a PR to partially implement this validation: https://github.com/apollographql/apollo-rs/pull/464

yanns:
I've opened a PR for this: https://github.com/apollographql/apollo-rs/pull/465

lrlna:
thanks @yanns! I'll take a look

lrlna:
closed by #465

Issue #306, opened 2022-09-21
validation: field selection merging


lrlna:
implement field selection merging, [spec](https://spec.graphql.org/October2021/#sec-Field-Selection-Merging).

in the compiler's db, fields should likely still be separated, but have an api that provides access to the merged field selection which is to be used in validation.

yanns:
The validation should check if the fields can be merged or not?
Or should we also merge the fields? For which validation?

yanns:
Maybe it's worth taking inspiration from https://github.com/sangria-graphql-org/sangria/pull/12

PR #441, opened 2023-01-28
feat(compiler): validate nested fields for definition on their types

erikwrede:
Following up on #436 and #392, I extended the validation to nested fields. To achieve that, I changed the `validate_selection_set` `validate_selection` and `validate_field` functions to accept the `TypeDef` alongside the relevant value. 
Please let me know what you think of this approach. If this is favorable, I will fix the linting and add extensive unit testing. 
Please also see my tagged TODO items. Some of them don't match the scope of this PR, but I'm happy to address them in follow-up PRs. The tests are still failing because I need to fix some undefined field errors in the graphql documents folder.


erikwrede:
After checking, it seems like I need to refactor almost all of the doc tests because they contain nested undefined fields / fields only available on implementations of the interfaces. I believe the tests could benefit from some streamlining so they are reduced to exactly the case that causes an error. Are there any open tasks regarding these that I could take into account? I'd prefer to do the refactor in a separate PR.

crates/apollo-compiler/src/validation/field.rs
```diff
@@ -10,11 +11,41 @@ use crate::{
...
+    let field_type = field.ty(db.upcast());
+    if let Some(field_type) = field_type {
+        // Get the type system definition for the type of the field - is there a better way to do this?
+        let field_type_def = field_type.type_def(db.upcast());
+
+        if let Some(field_type_def) = field_type_def {
+            diagnostics
+                .extend(db.validate_selection_set(field.selection_set().clone(), field_type_def));
+        } else {
+            // TODO what should we do if field_type_def is None although field_type is Some? Is that a case we are expecting?
```
lrlna, line 29: this would be impossible as `field.ty()`, the function that's called here to get `field_type` runs the same function as field_type_def - `db.find_type_definition_by_name()`. so if `field_type` is Some, field_type_def is always going to be Some

crates/apollo-compiler/src/validation/field.rs
```diff
@@ -10,11 +11,41 @@ use crate::{
...
     db: &dyn ValidationDatabase,
     field: Arc<hir::Field>,
+    type_def: hir::TypeDefinition,
 ) -> Vec<ApolloDiagnostic> {
     let mut diagnostics =
         db.validate_directives(field.directives().to_vec(), hir::DirectiveLocation::Field);
     diagnostics.extend(db.validate_arguments(field.arguments().to_vec()));
 
+    let field_type = field.ty(db.upcast());
+    if let Some(field_type) = field_type {
```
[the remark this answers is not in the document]
lrlna: I think introspection fields can be dealt with as a follow up. Scalar meta fields will work here, but we haven't added any of the introspection types. I am currently trying to figure out a path forward with #439 and can add the introspection bits here afterwards. 
I am not a fan of modifying the type signature of `field.ty`, as there are a lot of analysis that people do with the `Type` that function returns. returning a `String` will complicate thing and create ambiguity.

crates/apollo-compiler/src/validation/operation.rs
```diff
@@ -83,38 +89,63 @@ pub fn validate_operation_definitions(
...
+        let field_type_def = field.ty(db.upcast()).unwrap().type_def(db.upcast());
+
+        // TODO Handle Interface here - should be the same handling as object type
+        if let Some(TypeDefinition::ObjectTypeDefinition(field_type)) = field_type_def {
+            check_selection_set(db, field.selection_set(), field_type.clone(), diagnostics);
+        }
+    }
+}
+*/
 pub fn validate_subscription_operations(
```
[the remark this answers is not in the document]
lrlna: yes, i expect that some of the tests wouldn't pass this validation rule. thanks for fixing them!

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -34,6 +39,13 @@ pub fn validate_selection(
 pub fn validate_selection_set(
     db: &dyn ValidationDatabase,
     selection_set: hir::SelectionSet,
+    type_def: hir::TypeDefinition,
 ) -> Vec<ApolloDiagnostic> {
-    db.validate_selection(selection_set.selection)
+    let mut diagnostics = Vec::new();
+
+    // TODO handle Enums, Scalar (no selection set content allowed), Unions
```
lrlna, line 46: we can do this in a follow up PR

crates/apollo-compiler/src/validation/operation.rs
```diff
@@ -19,7 +19,13 @@ pub fn validate_operation_definitions(
         diagnostics
             .extend(db.validate_directives(def.directives().to_vec(), def.operation_ty().into()));
         diagnostics.extend(db.validate_variable_definitions(def.variables.as_ref().clone()));
-        diagnostics.extend(db.validate_selection_set(def.selection_set().clone()));
+        // TODO move this somewhere below the root operation type check
```
lrlna, line 22: ordering shouldn't really matter in validation 🤔 why does this need to be after root operation type check?

crates/apollo-compiler/src/validation/field.rs
```diff
@@ -10,11 +11,41 @@ use crate::{
 pub fn validate_field(
     db: &dyn ValidationDatabase,
     field: Arc<hir::Field>,
+    type_def: hir::TypeDefinition,
```
lrlna, line 14: I am wondering if it's necessary to have a type_def param? We can always get a field's type and it's type definition:
 
```rust
let ty = field.ty(db.upcast());
let type_def = field_type.type_def(db.upcast());
```

or is there something else you would like to do with this `TypeDefinition`?

lrlna (commented):
This is looking really good! Thank you so much for taking the time to put it together.

I don't think I necessarily follow the idea behind `type_def` argument. Could you elaborate a bit? I feel like that information can be available on `hir::Field` already.

Agree that we don't want to increase the scope of this PR. Field validation is quite large and complex, and a bunch of it we will handle as part of #306 and #307.

crates/apollo-compiler/src/validation/field.rs
erikwrede: The validation error for a missing field has a message along the lines of: `Field name is not defined on Type Person`. 

For this, I need the type def or name of the field's _parent type_ (Person). `field.ty` would return the name of the _field type_ (`None`, or `String` in case the field exists).
`field.parent_obj` would be the perfect choice, but it is optional. 
I just double-checked and it seems like we can still use it since it should be `Some` in case the field is defined on a valid parent object. This is ensured by performing the valid root operation type check first, to verify that the parent exists (explanation your review comment above)

> ordering shouldn't really matter in validation 🤔 why does this need to be after root operation type check?

If the paren't doesn't exist, we wouldn't continue validation and simply output the diagnostic.

erikwrede:
Thanks for the review, I will address your comments this weekend 🙂

crates/apollo-compiler/src/validation/field.rs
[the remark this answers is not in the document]
erikwrede: I was referring to the graphql scalar type `String` in this case, not changing the signature to always return a `String`. This was an Idea to not break validation for queries containing `__typename`, which is very common.

crates/apollo-compiler/src/validation/field.rs
lrlna: >For this, I need the type def or name of the field's parent type (Person). field.ty would return the name of the field type (None, or String in case the field exists).
ah, gotcha! let's call it `parent_ty` to disambiguate then. But I see you found a different way about it, which I think works well too.

> This is ensured by performing the valid root operation type check first, to verify that the parent exists (explanation your review comment above)

Ah! I am sorry, I didn't quite follow the first time around.

lrlna (approved):
Thank you so much for putting this together!

yanns:
It'd be great to have a release with that new feature.

lrlna:
hey @yanns we will do a release ~next week. We want to finish #306 and #307 to complete the field validation story. If we release now, you might run into some trouble still with operation fields!

Issue #362, opened 2022-11-15
Wrong order of ignored tokens around types


goto-bus-stop:
Probably related to #325 but not identical?

### Steps to reproduce

```rust
use apollo_parser::Parser;
let  ast = Parser::new(r#"
type Query {
    field: Int # comment
}
"#.trim()).parse();
dbg!(ast.document());
```

### Expected result

We want a tree like:
- NAMED_TYPE
  - NAME (`Int`)
- WHITESPACE (` `)
- COMMENT (`# comment`)
- WHITESPACE (`\n`)

### Actual result

We get a tree like:
- NAMED_TYPE
  - COMMENT (`# comment`)
  - WHITESPACE (`\n`)
  - NAME (`Int`)

### Etc

I think this happens because the custom type parsing has this order of operations:
- Pop the token for the type (`Int`) into a local variable
   https://github.com/apollographql/apollo-rs/blob/d77e358819c320ad0e57ab21b9198f072ff18169/crates/apollo-parser/src/parser/grammar/ty.rs#L45
- Start the `NAMED_TYPE` node. This consumes ignored tokens, adding the comment and whitespace tokens to the parser AST
  https://github.com/apollographql/apollo-rs/blob/d77e358819c320ad0e57ab21b9198f072ff18169/crates/apollo-parser/src/parser/grammar/ty.rs#L161-L162
- Add the type token (`Int`) to the parser AST
  https://github.com/apollographql/apollo-rs/blob/d77e358819c320ad0e57ab21b9198f072ff18169/crates/apollo-parser/src/parser/grammar/ty.rs#L164

So trailing whitespace and comments get attached to the AST first. The spans are also wrong because of this.

Some potential fixes:
- the type parsing could use `.peek()` instead for the initial token and then `.pop()` at the right time
- maybe `.start_node()` should not eat ignored tokens, but force us to do it manually, to have more control over the order of tokens in the AST.

lrlna:
ohhh this is _purely_ related to how different `ty.rs` does parsing, so it is, in fact, different to #325. I have previously considered to have `start_node()` to eat ignored tokens, but it adds a whole lot of overhead to all existing parsing rules? Perhaps worth it? Would be good to perhaps figure out how to do it accurately with just `peek()` and `pop()`.

Freyert:
👋 I heard about this project on a podcast I believe, and it seemed like an interesting place to work on some of my interests (but weak points: rust, and everything in the compiler space).

[...]

PR #443, opened 2023-01-31
fix(parser): Fix token order in parse tree around types

goto-bus-stop:
Closes https://github.com/apollographql/apollo-rs/issues/362

This uses rowan's checkpoint feature. We can set a checkpoint before parsing a type node, parse the type normally, and then if we see a `!` at the end, we retroactively wrap the node in a NonNullType node. Then we add the `!` token to this NonNullType node.

Compared to the previous two-step parsing strategy that we used for this, we can no longer run into issues with tokens being added to the parse tree out of order.

Ignored tokens like commas, whitespace, and comments are not considered part of the type node, but part of the parent node. eg in this:
```graphql
type Object {
  field: Ty # comment
}
```
`# comment` is part of the FieldDefinition instead of part of the ObjectTypeDefinition. I think *ideally* we could hoist it up even further, so the FieldDefinition doesnt include trailing whitespace and comments either. that would improve our error display too.

We now also support whitespace inside type nodes better, eg `[ [Type !] ] !` while previously there could be no whitespace before the `!`. Because I just used `.bump_ignored()`, you can also do some weird things like `[ Type , , , ! ]` which looks wrong, however graphql-js also allows this, so maybe it's intentional 🤷🏻‍♀️  .

### DIFF
--- a/crates/apollo-compiler/src/tests.rs
+++ b/crates/apollo-compiler/src/tests.rs
@@ -39,6 +39,9 @@ fn compiler_tests() {
         compiler.add_document(text, path.file_name().unwrap());
 
         let diagnostics = compiler.validate();
+        for diagnostic in &diagnostics {
+            println!("{diagnostic}");
+        }
         assert_diagnostics_are_present(&diagnostics, path);
         format!("{diagnostics:#?}")
     });
--- a/crates/apollo-compiler/test_data/ok/0003_schema_definition_with_custom_operation_types.txt
+++ b/crates/apollo-compiler/test_data/ok/0003_schema_definition_with_custom_operation_types.txt
@@ -104,13 +104,13 @@
                             - IDENT@255..259 "name"
                         - COLON@259..260 ":"
                         - WHITESPACE@260..261 " "
-                        - NON_NULL_TYPE@261..270
+                        - NON_NULL_TYPE@261..268
                             - NAMED_TYPE@261..267
                                 - NAME@261..267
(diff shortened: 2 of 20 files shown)


crates/apollo-compiler/src/tests.rs
```diff
@@ -39,6 +39,9 @@ fn compiler_tests() {
         compiler.add_document(text, path.file_name().unwrap());
 
         let diagnostics = compiler.validate();
+        for diagnostic in &diagnostics {
+            println!("{diagnostic}");
```
goto-bus-stop, line 43: helps debugging with `cargo t -p apollo-compiler -- --nocapture`

lrlna (approved):
This makes `ty` parsing so much more readable! 🙏 Thank you so much!

> We now also support whitespace inside type nodes better, eg `[ [Type !] ] !` while previously there could be no whitespace before the `!`. Because I just used `.bump_ignored()`, you can also do some weird things like `[ Type , , , ! ]` which looks wrong, however graphql-js also allows this, so maybe it's intentional 🤷🏻‍♀️ .

_very_ excited for this bit in particular >.<!! can we add a test for all these extra wobbly types?

goto-bus-stop:
This test does a few weird things with whitespace & commas:
https://github.com/apollographql/apollo-rs/blob/f16665616becadb58d884790d41cea3917d43886/crates/apollo-parser/test_data/parser/ok/0040_type_token_order.graphql

PR #445, opened 2023-02-01
fix(parser): Fix ignored token positioning in the AST

goto-bus-stop:
This makes spans for all nodes more specific, not including ignored tokens *after* the node. When ignored tokens are consumed, they are first stored separately, and then added to the AST just before the next node is started. This way ignored tokens are always inside the outermost possible node, and therefore all individual nodes will have spans that only contain that node and not more.

Todo:
- [x] Currently this has one regression with some preceding whitespace ending up at inside a node, while it should be *before* the start of the node


lrlna (approved):
Getting all the tokens in a row is bringing me a lot of joy ^.^

> Currently this has one regression with some preceding whitespace ending up at inside a node, while it should be before the start of the node

What input gives results in this (lots of tests changed, not going to look through them all 😅 )

goto-bus-stop:
Oh, i found it already. it happens in the new NonNullType parsing code We create a rowan Checkpoint first, and then when we call `start_node`, we process the queued ignored tokens. Then when we reach the `!` token, we create the wrapping NonNullType node, but the checkpoint includes the preceding ignored tokens. I think handling ignored tokens before we set a checkpoint should do the trick

goto-bus-stop:
Noticeable improvement in diagnostics: only the field name itself is now highlighted, the newline after it is not included. the span for the subscription type points to the correct ending line too.

image

this is still definitely one of the more confusing diagnostics, we still need more context lines for that

lrlna:
> Oh, i found it already. it happens in the new NonNullType parsing code We create a rowan Checkpoint first, and then when we call `start_node`, we process the queued ignored tokens. Then when we reach the `!` token, we create the wrapping NonNullType node, but the checkpoint includes the preceding ignored tokens. I think handling ignored tokens before we set a checkpoint should do the trick

excellent!! really happy to have this sorted 🙏 and the diagnostics are a lot more neat, even though it was just a couple of ignored tokens at the end

hoping we can work on that additional context as we finish up validation, perfect timing for it.

PR #446, opened 2023-02-03
fix(parser): Fix spans after parsing unexpected tokens

### LINKED ISSUE #325: erroneous spans when an error is created with err_and_pop
Our error spans are inconsistent when an error is created with `err_and_pop`. 

A few examples.
GraphQL:
```graphql
cats
# comment
```
Document:
```
- DOCUMENT@0..10
    - WHITESPACE@0..1 "\n"
    - COMMENT@1..10 "# comment"
- ERROR@0:4 "expected definition" cats
```

GraphQL:
```graphql
cats

type Query {
    name: String
}
```
Document:
```
- DOCUMENT@0..33
    - WHITESPACE@0..2 "\n\n"
    - OBJECT_TYPE_DEFINITION@2..33
        - type_KW@2..6 "type"
        - WHITESPACE@6..7 " "
        - NAME@7..13
            - IDENT@7..12 "Query"
            - WHITESPACE@12..13 " "
        - FIELDS_DEFINITION@13..33
            - L_CURLY@13..14 "{"
            - WHITESPACE@14..19 "\n    "
            - FIELD_DEFINITION@19..32
                - NAME@19..23
                    - IDENT@19..23 "name"
                - COLON@23..24 ":"
                - WHITESPACE@24..25 " "
                - NAMED_TYPE@25..31
                    - NAME@25..31
                        - IDENT@25..31 "String"
                - WHITESPACE@31..32 "\n"
            - R_CURLY@32..33 "}"
- ERROR@0:4 "expected definition" cats
```

This no longer becomes the case when the error precedes a valid statement, so if we reverse the two statements above:

```graphql
type Query {
    name: String
}

cats
```
Our Document's spans are then correct:
```
- DOCUMENT@0..33
    - OBJECT_TYPE_DEFINITION@0..33
        - type_KW@0..4 "type"
        - WHITESPACE@4..5 " "
        - NAME@5..11
            - IDENT@5..10 "Query"
            - WHITESPACE@10..11 " "
        - FIELDS_DEFINITION@11..33
            - L_CURLY@11..12 "{"
            - WHITESPACE@12..17 "\n    "
            - FIELD_DEFINITION@17..30
                - NAME@17..21
                    - IDENT@17..21 "name"
                - COLON@21..22 ":"
                - WHITESPACE@22..23 " "
                - NAMED_TYPE@23..29
                    - NAME@23..29
                        - IDENT@23..29 "String"
                - WHITESPACE@29..30 "\n"
            - R_CURLY@30..31 "}"
            - WHITESPACE@31..33 "\n\n"
- ERROR@33:37 "expected definition" cats
```

Initially found in #324

goto-bus-stop:
When there is a parse error, we ignore the token that caused the error. Then we continue parsing. The offsets associated with parse nodes increase linearly as tokens are added, so if we skip any token, the spans for all future nodes will be wrong. This adds an `ERROR` syntax token kind so we can add error tokens to nodes, which advances the spans, and fixes the issue. The tree is also fully lossless now.

There could be other similar cases, where we pop a token and never add it to the tree, but I think we can address those cases as we find them

I'm not sure if the addition of a new type of token to the parse tree could have issues downstream, I don't think it should (as we just ignnore them when building the typed tree), but maybe you have a better idea @lrlna ?

Ref #325

### DIFF
--- a/crates/apollo-parser/src/lexer/token.rs
+++ b/crates/apollo-parser/src/lexer/token.rs
@@ -92,6 +92,9 @@ impl fmt::Debug for Token {
             TokenKind::Pipe => {
                 write!(f, "PIPE@{}:{} {:?}", start, end, self.data)
             }
+            TokenKind::Error => {
+                write!(f, "ERROR@{}:{} {:?}", start, end, self.data)
+            }
             TokenKind::Eof => {
                 write!(f, "EOF@{start}:{start}")
             }
--- a/crates/apollo-parser/src/lexer/token.rs
+++ b/crates/apollo-parser/src/lexer/token.rs
@@ -92,9 +92,6 @@ impl fmt::Debug for Token {
             TokenKind::Pipe => {
                 write!(f, "PIPE@{}:{} {:?}", start, end, self.data)
             }
-            TokenKind::Error => {
-                write!(f, "ERROR@{}:{} {:?}", start, end, self.data)
-            }
             TokenKind::Eof => {
                 write!(f, "EOF@{start}:{start}")
             }
--- a/crates/apollo-parser/src/lexer/token_kind.rs
+++ b/crates/apollo-parser/src/lexer/token_kind.rs
@@ -32,7 +32,6 @@ pub enum TokenKind {
     LCurly,     // {
     RCurly,     // }
     Pipe,       // |
-    Error,
     Eof,
 
     // composite nodes
--- a/crates/apollo-parser/src/lexer/token_kind.rs
+++ b/crates/apollo-parser/src/lexer/token_kind.rs
@@ -32,6 +32,7 @@ pub enum TokenKind {
     LCurly,     // {
     RCurly,     // }
     Pipe,       // |
+    Error,
     Eof,
 
     // composite nodes
--- a/crates/apollo-parser/src/parser/generated/syntax_kind.rs
+++ b/crates/apollo-parser/src/parser/generated/syntax_kind.rs
@@ -76,6 +76,7 @@ pub enum SyntaxKind {
     IDENT,
     WHITESPACE,
     COMMENT,
+    ERROR,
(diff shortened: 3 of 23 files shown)


lrlna (approved):
ohhhh that's an interesting idea! i think it should work!

let's add a test or two inside parser's src to check if we can still access various bits of an incorrect ast, even though the error token is now part of the document.

goto-bus-stop:
This PR handles parse failures with valid tokens, but there's another similar issue for lexing errors which don't produce tokens in the first place. i'll try to address that here too 🤔

goto-bus-stop:
Hmmm. I think the stuff i've pushed now doesn't really have the effect I want. It probably needs a little bit of design thought. I'll revert those changes so we can get the partial / easier fix in, and then I can follow up with a more comprehensive approach soon.

goto-bus-stop:
OK there is one more issue with this. If there is an error inside some definition, the parser bails out of the definition completely, so something like this:
```graphql
type Object {
  a: Int
  :This_is_an_error|
  b: Int
}
```
results in a parse tree that represents `type Object { a: Int }` only. Ideally we would be able to skip the broken part and still pick up that `b: Int` is also meant to be a field. But i'm not sure how to approach that, it would probably have to be done differently than this, and it might be a separate issue.

I think these are 3 separate things:
- [ ] Lexer errors are skipped so the spans after a lexer error are wrong.
- [x] Parse errors are skipped so the spans after a parse error are wrong (addressed by this PR)
- [ ] Parse errors can cause us to bail out too hard

I'd say we go ahead with this fix, though it's not as complete as I hoped. then we can address the remainder after we finish the higher-priority validation work.

lrlna:
@goto-bus-stop let's merge this in as is, and create separate issues for the other two discrepancies you found here.

PR #447, opened 2023-02-03
breaking(compiler): unify find_operation

lrlna:
Mimics spec's [`getOperation`](https://spec.graphql.org/October2021/#GetOperation()) functionality and returns the anonymous operation if `None` is specified for operation name.


goto-bus-stop (approved):
nice this does read much better 😄

crates/apollo-compiler/src/database/document.rs
```diff
@@ -38,24 +38,14 @@ pub(crate) fn find_type_definition_by_name(
...
-pub(crate) fn find_anonymous_operation(
+pub(crate) fn find_operation(
     db: &dyn HirDatabase,
     file_id: FileId,
+    name: Option<String>,
 ) -> Option<Arc<OperationDefinition>> {
     db.operations(file_id)
         .iter()
-        .find(|def| def.name().is_none())
+        .find(|def| def.name() == name.as_deref())
```
SimonSapin, line 48: When the `name` parameter is `None` is looks for an anonymous operation. This deviates from [the spec’s algorithm](https://spec.graphql.org/October2021/#GetOperation()), which instead does something like:


```rust
let operations = db.operations(file_id);
if let Some(name) = name {
    operations.iter().find(|def| def.name() == Some(&*name))
} else if operations.len() == 1 {
    operations[0].clone()
} else {
    None
}

SimonSapin (changes requested):

crates/apollo-compiler/src/database/document.rs
lrlna: oh, good catch, i misread that entirely!

lrlna: we don't (yet, or maybe even never?) return Result from these kinds of queries, so if `None` is provided and there are _several_ operation definitions, this function will return `None`. Does that sound ok to everyone? @SimonSapin @goto-bus-stop?

https://github.com/apollographql/apollo-rs/blob/cd590648ef946333e8cf1c2deec464823d641395/crates/apollo-compiler/src/database/hir_db.rs#L1626-L1630

SimonSapin: Yes, sounds good. Depending on whether the parameter is `None` (which the caller knows of course) in each case there’s only one error case, so it’s fine IMO to have a `None` return value represent it.

SimonSapin (commented):
Sorry, I think this still doesn’t quite match the spec. The path checking `len() == 1` should only be taken when the `name` parameter is None. The path searching for a given `def.name()` should only be taken when the parameter is Some.

Or expressed as test cases:

* With `query A { b }`, `find_operation(Some("C"))` should return None because we ask for a specific name that isn’t there
* With `query D { e } query { f }`, `find_operation(None)` should return None because not specifying a name is ambiguous when there are multiple definitions.

SimonSapin:
By the way, as far as I can tell that second document does not raise a validation error for Operation Name Uniqueness nor for Lone Anonymous Operation, but that anonymous operation is impossible to execute!

SimonSapin (approved):
:+1:

lrlna:
well this has been a bad week!!

> By the way, as far as I can tell that second document does not raise a validation error for Operation Name Uniqueness nor for Lone Anonymous Operation, but that anonymous operation is impossible to execute!

I don't think so. It's an `update_executable` function, which means the executables get overwritten, so there is no name conflicts.

Issue #444, opened 2023-01-31
Context around error diagnostics


goto-bus-stop:
We moved from miette to ariadne so we can create diagnostics that reference multiple files at once. This caused one major regression in our diagnostic display: ariadne often prints way less context.

|Before|After|
|-|-|
|![image](https://user-images.githubusercontent.com/1006268/215773375-ce9dd12e-a974-45d0-8f90-f648b1320d7f.png)|![image](https://user-images.githubusercontent.com/1006268/215773404-8932328c-c95a-4924-97d4-4f8e5ea0e5f1.png)|

In the *After* case it's very hard to see where exactly this is happening, you basically have to jump to the given line number in your editor to make sense of it. Having a few lines displayed around the start and end of each labeled span would help a lot. I think this is not (yet) possible with ariadne, so we should contribute that feature if the author wants it.

PR #452, opened 2023-02-09
Validate correct arguments are provided

### LINKED ISSUE #308: validation: validate arguments

goto-bus-stop:
This validates that all required arguments are provided, and that only known arguments are provided. It does not do type-checking.

Fixes #308

Some existing tests had to be updated to provide required arguments 😇 

Example output:
image
(ref #444, it would be nice if we could also point to the type that the field was defined in)

### DIFF
--- a/crates/apollo-compiler/src/validation/field.rs
+++ b/crates/apollo-compiler/src/validation/field.rs
@@ -29,7 +29,7 @@ pub fn validate_field(
                 let diagnostic = ApolloDiagnostic::new(
                     db,
                     arg.loc.into(),
-                    DiagnosticData::MissingArgument {
+                    DiagnosticData::UndefinedArgument {
                         name: arg.name().into(),
                     },
                 )
@@ -57,7 +57,7 @@ pub fn validate_field(
                 let mut diagnostic = ApolloDiagnostic::new(
                     db,
                     field.loc.into(),
-                    DiagnosticData::MissingArgument {
+                    DiagnosticData::RequiredArgument {
                         name: arg_def.name().into(),
                     },
                 );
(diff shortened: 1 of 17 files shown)


crates/apollo-compiler/src/validation/field.rs
```diff
@@ -12,7 +12,67 @@ pub fn validate_field(
...
+                    DiagnosticData::UndefinedArgument {
+                        name: arg.name().into(),
+                    },
+                )
+                .label(Label::new(arg.loc, "argument by this name not found"))
+                .label(Label::new(field_definition.loc, "field declared here"));
+
+                diagnostics.push(diagnostic);
+            }
+        }
```
lrlna, line 41: should we add this to `validate_arguments` instead? i feel like that could be called from both `field` and `directive`.

lrlna (commented):
nice ✔️

crates/apollo-compiler/src/validation/field.rs
goto-bus-stop: Ah yes. i started with that but to build the diagnostics, we have to refer to different types (FieldDefinition and DirectiveDefinition respectively), so it's not trivial. I think duplicating the code is easier than trying to introduce some abstraction for it or trying to share more functionality.

lrlna (approved):

PR #454, opened 2023-02-09
feat(compiler): find queries accept &str

lrlna:
This should make it a bit easier querying for definitions. Because all find_queries are transparent, and not cacheable, we can use &str instead of String.

### DIFF
--- a/crates/apollo-compiler/src/database/hir_db.rs
+++ b/crates/apollo-compiler/src/database/hir_db.rs
@@ -78,36 +78,33 @@ pub trait HirDatabase: InputDatabase + AstDatabase {
     /// Return an fragment definition corresponding to the name and file id.
     /// Result of this query is not cached internally.
     #[salsa::transparent]
-    fn find_fragment_by_name(
-        &self,
-        file_id: FileId,
-        name: String,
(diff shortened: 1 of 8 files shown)


crates/apollo-compiler/src/database/hir_db.rs
```diff
@@ -1315,7 +1312,7 @@ fn field(
 
 fn parent_ty(db: &dyn HirDatabase, field_name: &str, parent_obj: Option<String>) -> Option<String> {
```
goto-bus-stop, line 1313: This function could take a `parent_obj: Option<&str>`

goto-bus-stop (approved):
definitely a nice simplification for typical usecases 😄
 
my only concern is that this kind of exposes an implementation detail. some `find_*` functions still require a String because they are cached, while others don't because they are not cached. and if the implementation changes in the future we may have to revert 🤔  i'm not sure how significant this is as rustc will tell you what to do, but maybe in the future it'd be worth considering having `&str` wrappers around the caching functions (even if that means an extra clone internally).

lrlna:
Going to close this as we are rewriting the way `find_` is going to be working in the new version of `apollo-compiler`, and this will no longer be relevant.

PR #455, opened 2023-02-09
feat(compiler): validate unique variables, validate variables are of input type

lrlna:
partial of #312, will validate variables match their defined type system type in a separate PR, as the work there is a bit more extensive.


crates/apollo-compiler/src/validation/variable.rs
```diff
@@ -13,81 +16,129 @@ pub fn validate_variable_definitions(
...
+            let type_def = ty.type_def(db.upcast());
+            if let Some(type_def) = type_def {
+                let ty_name = type_def.kind();
+                diagnostics.push(
+                    ApolloDiagnostic::new(db, variable.loc().into(), DiagnosticData::InputType {
+                        name: variable.name().into(),
+                        ty: ty_name,
+                    })
+                    .label(Label::new(ty.loc().unwrap(), format!("this is of `{ty_name}` type")))
+                    .help("objects, unions, and interfaces cannot be used as variables can only be of input type"),
```
goto-bus-stop, line 37: kinda nitpicky but the first time I read this as "objects, unions, and interfaces cannot be used as variables", with a different meaning of _as_ 😅 
```suggestion
                    .help("objects, unions, and interfaces cannot be used because variables can only be of input type"),
```

crates/apollo-compiler/src/validation/variable.rs
```diff
@@ -13,81 +16,129 @@ pub fn validate_variable_definitions(
...
+        .variables()
+        .iter()
+        .map(|var| ValidationSet {
+            name: var.name().into(),
+            loc: var.loc(),
+        })
+        .collect();
+    let used_vars: HashSet<ValidationSet> = op
+        .selection_set
+        .clone()
```
goto-bus-stop, line 100: this clone looks unnecessary(?)

crates/apollo-compiler/test_data/diagnostics/0054_variables_are_input_types.graphql
```diff
@@ -0,0 +1,86 @@
+query takesCat($cat: Cat) {
```
goto-bus-stop, line 1: https://github.com/apollographql/apollo-rs/pull/452 added `0054_*` test data so this PR should use 0055 and 0056

(maybe we can add a check to the test runner that each number is unique)

goto-bus-stop (commented):
looks nice :)

crates/apollo-compiler/src/validation/variable.rs
lrlna: hah fair fair

crates/apollo-compiler/test_data/diagnostics/0054_variables_are_input_types.graphql
lrlna: we probably should 🤔 becomes annoying to track manually

goto-bus-stop (approved):

PR #456, opened 2023-02-10
Check name uniqueness across different types

### LINKED ISSUE #386: validation: name uniqueness should apply across different definition types

goto-bus-stop:
Fixes #386

This replaces the multiple cases of name uniqueness validations by one single implementation, so that all type definitions (`object`/`interface`/`scalar`/`enum`/`union`/`input`) can share one scope. Fragments, directives, and operations are each in their own scope, but they can be validated by the same implementation.

image

Todo:
- [x] add a `ok` test with duplicate, but valid names in different scopes


crates/apollo-compiler/src/validation/validation_db.rs
```diff
@@ -180,9 +185,105 @@ pub fn validate(db: &dyn ValidationDatabase) -> Vec<ApolloDiagnostic> {
...
+                .children()
+                .filter_map(ast::Definition::cast)
+                // Extension names are allowed to be duplicates,
+                // and schema definitions don't have names.
+                .filter(|def| {
+                    !def.is_extension_definition()
+                        && !matches!(def, ast::Definition::SchemaDefinition(_))
+                })
+                .map(move |def| (file_id, def))
+        });
```
lrlna, line 213: should we have this as its own salsa query in hir_db?

lrlna (approved):
bless! 🙏

lrlna:
something @gocamille noticed is our current `db.fragments()` query doesn't allow for duplicate fragments to begin with. that is, we are simply silent about duplicate fragment definitions \o/. i filed #457

crates/apollo-compiler/src/validation/validation_db.rs
goto-bus-stop: in theeeory this should be the only place we use this approach, anything that does not need to know about duplicate names will use the existing queries.

crates/apollo-compiler/test_data/diagnostics/0056_variables_are_input_types.graphql
```diff
@@ -20,7 +20,7 @@ query takesCatOrDog($catOrDog: CatOrDog) {
   }
 }
 
-query takesCatOrDog($catOrDog: Dragon) {
+query takesCatOrDog2($catOrDog: Dragon) {
```
goto-bus-stop, line 23: @lrlna only noticed this because the error order changed in this PR … did this intentionally have a duplicate operation name?

lrlna: no, not intentional! human error

lrlna:
> #457 would be addressed by this PR, the HIR generally doesn't support duplicate names, so this implementation uses the AST directly instead.
> 
> In general errors _inside_ duplicate definitions are not found, but as you already get an error saying that the name is duplicate, you atleast know what to fix first.

(beyond the scope of this particular PR) i agree with the idea that the HIR should be a fault-free representation of the a given document, but i wondering if us silently ignoring some of the definitions defined in source is a bit too confusing from a user perspective.... probably a problem for future us

PR #458, opened 2023-02-14
Validate extension existence & kind

### LINKED ISSUE #408: Validation: extensions of undefined type
Two scenarios where validation should emit a diagnostic if they don’t already:

* The AST contains a type extension, but there is no type definition with that name
* There is one, but of the wrong kind (e.g. enum instead of union)

goto-bus-stop:
Extensions referencing nonexistent definitions produce a diagnostic:
image

Extensions referencing some other kind of definition produce a diagnostic:
image

This adds an `extensions()` salsa query that creates HIR nodes for _all_ declared extensions. The new validation uses that to iterate over all the extensions. The existing queries that used to collect extensions from the AST now also use this new query so we don't create the nodes in two places.

Closes #408


crates/apollo-compiler/src/validation/extension.rs
```diff
@@ -0,0 +1,75 @@
...
+pub fn validate_extensions(db: &dyn ValidationDatabase) -> Vec<ApolloDiagnostic> {
+    let mut diagnostics = Vec::new();
+
+    for extension in db.extensions().iter() {
+        let Some(definition) = db.find_type_definition_by_name(extension.name().into()) else {
+            diagnostics.push(
+                ApolloDiagnostic::new(db, extension.loc().into(), DiagnosticData::UndefinedDefinition {
+                    name: extension.name().into(),
+                })
+                .label(Label::new(extension.name_src().loc.unwrap(), format!("a type definition named `{}` does not exist", extension.name()))),
```
lrlna, line 24:
```suggestion
                .label(Label::new(extension.name_src().loc.unwrap(), format!("`{}` type definition does not exist", extension.name()))),
```

crates/apollo-compiler/test_data/diagnostics/0059_extension_orphan.graphql
```diff
@@ -0,0 +1,19 @@
+type Query {
```
lrlna, line 1: sorry, you're going to have to renumber these 😭

lrlna (approved):
yay~~~

crates/apollo-compiler/test_data/diagnostics/0059_extension_orphan.graphql
goto-bus-stop: :D

PR #463, opened 2023-02-28
Fix infinite loop for incomplete fragment spreads

allancalix:
## Summary
Incomplete fragment spreads current don't bump the token stream and cause an infinite loop. This change updates the logic to bump the fragment spread operator on an invalid next token.


goto-bus-stop (approved):
Thanks, looks good!

goto-bus-stop:
Not sure what's up with CI 😅

PR #464, opened 2023-02-28
validate leaf selection of scalar & enum

yanns:
Partial implementation of https://github.com/apollographql/apollo-rs/issues/307

From https://spec.graphql.org/October2021/#sec-Leaf-Field-Selections
```
If selectionType is a scalar or enum:

    The subselection set of that selection must be empty
```


yanns:
Closed in favor of https://github.com/apollographql/apollo-rs/pull/465

PR #465, opened 2023-03-01
Leaf selection

yanns:
implementation of https://github.com/apollographql/apollo-rs/issues/307

Contains https://github.com/apollographql/apollo-rs/pull/464

https://spec.graphql.org/October2021/#sec-Leaf-Field-Selections


crates/apollo-compiler/src/lib.rs
```diff
@@ -574,7 +574,7 @@ scalar URL @specifiedBy(url: "https://tools.ietf.org/html/rfc3986")
 "#;
 
         let product_query = r#"query getProduct { topProducts { type } }"#;
-        let customer_query = r#"{ customer }"#;
+        let customer_query = r#"{ customer { id } }"#;
```
yanns, line 577: I had to change some queries to avoid the new validations. I hope it's right.

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -29,15 +31,142 @@ pub fn validate_selection(
...
+                    ))
+                } else {
+                    diagnostic
+                };
+                diagnostics.push(diagnostic);
+            }
+        }
+    }
+
+    if leaf_validation_error.is_none() {
```
yanns, line 95: If the leaf validation fails, I did not go into the validation of the field itself as it was much noise.
For example, the `SelectionOfEnum` query is the test would have validation errors: one about the sub selection, and one about the unknown `name` field.

erikwrede: I think it makes sense to integrate this function into `validate_field`, as leaf field validation is part of field validation in the spec. Additionally we make use of the existing derived query like that. The leaf field validation could be integrated here (L78) (or merge that L20 alternatively since L78 also just requires the field to exist on the type, but with a different call):
https://github.com/apollographql/apollo-rs/blob/49043027fca3e1c16dbf3437832a629afbd2e00b/crates/apollo-compiler/src/validation/field.rs#L78-L85

This way you can avoid going into `validate_selection_set` for the selection set of an enum in case it is supposed to be a leaf field, also reducing the noise but keeping field validation in the same place, also you have all the data like `field_ty` and `field_definition` (L20) there already.

Just an idea, what's your opinion on that? 🙂

yanns: Yes, that's a good idea.
Please check https://github.com/apollographql/apollo-rs/pull/465/commits/5ee19ce54212a45668c1d8ab48cee86a76dcfce8
I could not re-use `field_definition` as I need the type definition.

yanns:
General question:
if we want the tests to be exhaustive, we should have a test for each type, and for each type test:
- with a non-null value
- with an optional value
- with a non-null list of non-null value
- with a non-null list of optional value
- with a optional list of non-null value
- with a optional list of optional value

I've tried locally and it's working.
Should we add all those test cases? On one hand, it's good to be exhaustive, on the other hand, it's adding lots of line that could be noise later.
I don't have any strong opinion here.
WDYT?

lrlna:
> General question:
if we want the tests to be exhaustive, we should have a test for each type, and for each type test:

exhaustive is good! the test_data directory is meant specifically for that; it doesn't obstruct the readability of existing files and let's us be as detailed as we want. you can likely group the examples you provided above in a single file with an operation (or several operations) that have a nullable/non-nullable/etc values.

crates/apollo-compiler/src/validation/field.rs
```diff
@@ -210,3 +214,122 @@ pub fn validate_field_definitions(
...
+        compiler.add_document(input, "schema.graphql");
+
+        let diagnostics = compiler.validate();
+        for diagnostic in &diagnostics {
+            println!("{diagnostic}");
+        }
+
+        assert_eq!(diagnostics.len(), 1)
+    }
+}
```
lrlna, line 335: these can live in `test_data` dir

crates/apollo-compiler/src/diagnostics.rs
```diff
@@ -265,6 +265,20 @@ pub enum DiagnosticData {
         /// Name of the field
         field: String,
     },
+    #[error("`{name}` field of type '{ty}' must not have a sub selection")]
+    NoSubselectionAllowed {
```
lrlna, line 269:
```suggestion
    DisallowedSubselection {
```

crates/apollo-compiler/src/diagnostics.rs
```diff
@@ -265,6 +265,20 @@ pub enum DiagnosticData {
...
+        // field name
+        name: String,
+        // field type
+        ty: String,
+    },
+    #[error("`{name}` field of type '{ty}' must have a sub selection")]
+    MandatorySubselection {
```
lrlna, line 276:
```suggestion
    MissingSubelection {
```
`rustc` diagnostics have a whole category of `Missing` errors when something that should be there isn't. @goto-bus-stop what are your thoughts?

crates/apollo-compiler/src/validation/field.rs
yanns: They already are. I put some tests here for a faster feedback during development like
```
cargo watch -x "test it_validates_enum_as_leaf_field --package apollo-compiler -- --nocapture"
```

I can remove them now.

crates/apollo-compiler/src/diagnostics.rs
yanns: https://github.com/apollographql/apollo-rs/pull/465/commits/3c531a8a5151aeb4a6d356f95ec61d0f46583c41

crates/apollo-compiler/src/diagnostics.rs
yanns: https://github.com/apollographql/apollo-rs/pull/465/commits/3c531a8a5151aeb4a6d356f95ec61d0f46583c41

yanns:
> > General question:
> > if we want the tests to be exhaustive, we should have a test for each type, and for each type test:
> 
> exhaustive is good! the test_data directory is meant specifically for that; it doesn't obstruct the readability of existing files and let's us be as detailed as we want. you can likely group the examples you provided above in a single file with an operation (or several operations) that have a nullable/non-nullable/etc values.

Thanks for the feedback.
My try: https://github.com/apollographql/apollo-rs/pull/465/commits/5c9c8d056468f27e98988e6f7cfe29a8a52c6f56

crates/apollo-compiler/src/validation/field.rs
yanns: https://github.com/apollographql/apollo-rs/pull/465/commits/1114eca6719d9aca3e9e1f70d4e4095a4c86e5a8

lrlna:
thanks so much @yanns for putting this all together! @goto-bus-stop is wrapping up field merging, and we should be good to do a release :)

Issue #420, opened 2023-01-13
modifying documents in compiler context


lrlna:
### Context
It's becoming more important for the compiler to be able to modify and work on modified versions of its type system. While compiler's internals should remain immutable it'd be useful to come up with an API to return a modified version of a given document in either a String or perhaps a HIR format that could be accessed programmatically.

### Use cases
- creating API Schema (modifying original type system to remove field marked `@inaccessible`)
- modifying executable documents to remove certain fields
- break down executable documents into several documents

### Possible example
Accessing values of API Schema:

```rust
    let schema = Path::new("crates/apollo-compiler/examples/documents/type_system.graphql");
    let src = fs::read_to_string(schema).expect("Could not read type system file.");
    compiler.add_type_system(&src, schema);
   
    let query = Path::new("crates/apollo-compiler/examples/documents/get_dog_name.graphql");
    let src = fs::read_to_string(query).expect("Could not read executable file.");
    compiler.add_executable(&src, query);
    
    // modified version of the type system that has `@inaccessible` fields removed
    let api_type_system = compiler.db.api_type_system(); 
    // should probably have the exact same surface level api as the original type system?
    let query_obj_type = api_schema.find_object_type_by_name("Query".into()); 
```

I am not entirely sure how the internals of the compiler need to change to accommodate this just yet (aside from saying let's not mutate the state!). I think it'd be good to gather a few more examples of how we'd like this to be used before we settle on the architecture and design. If anyone has any specific examples of how they'd like to use modifying documents with the compiler, please bring them forward!

Geal:
The design of immutable collection libraries could provide some inspiration here. Instead of modifying, it's building a new view that reuses elements of the old one. As.an example, if you have a tree, you can "insert" or "delete" by building a new tree that reuses some sub trees

lrlna:
>  As.an example, if you have a tree, you can "insert" or "delete" by building a new tree that reuses some sub trees
iirc that's how rowan deals with reparsing, which i think should tie together very nicely with this API as well

lrlna:
something i also just thought about is that it's likely that examples such as APISchema creation will actually have to live outside the compiler. Which means we should be providing the semantics and architecture to deal with context modification, but also allow users to specify how they want it done.

^ perhaps was already implicit part of this issue, but better be said out loud

SimonSapin:
[...]

PR #466, opened 2023-03-01
add TryFrom impls to convert from hir to encoder types (#424)

### LINKED ISSUE #424: tryfrom for HIR types for apollo_encoder types
similar to #278, it'd be good to be able to encode HIR types to apollo-encoder types in order to modify input strings. We should possibly hold off on implementation of this until after we decide on the design for #420.

zackangelo:
Took a shot at implementing these conversion impls. The issue mentioned that the approach required is dependent on some other decisions. Let me know if this aligns with the direction you'd like to take and I'll clean up and add some tests. 

Fixes #424

### DIFF
--- a/crates/apollo-compiler/Cargo.toml
+++ b/crates/apollo-compiler/Cargo.toml
@@ -34,6 +34,10 @@ expect-test = "1.1"
 miette = "5.0"
 notify = "4.0.0"
 criterion = "0.3.0"
+pretty_assertions = "0.7.1"
+apollo-encoder = { path = "../apollo-encoder", version = "0.5.1", features = [
+  "apollo-compiler",
+] }
 
 [[bench]]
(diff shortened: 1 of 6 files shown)


lrlna:
@zackangelo thank you for attempting to tackle this!

I think we need to scope out a few more use cases and a direction for #420 before committing to implementing #424. I am not super keen to add extra API surface area before then. I'll keep this PR open for now, and give you a shout if we end up needing a TryFrom for the encoder types.

zackangelo:
Sounds great, please let me know if I can be of further help.

lrlna:
@zackangelo i thought about it some more, and regardless to our decision on #420, this would be a useful addition. especially for testing :)

lrlna (approved):

PR #470, opened 2023-03-09
Implement fields_in_set_can_merge

goto-bus-stop:
This implements the validation part of https://github.com/apollographql/apollo-rs/issues/306, namely the `FieldsInSetCanMerge` spec function.

This PR doesn't produce a merged output, though the new `SelectionSet::merge` method basically does that anyways (it just doesn't dedupe fields). If the selection set was validated, the `SelectionSet::merge` result can be used safely.

Marked as draft because:
 - [ ] Some functions are arbitrarily part of Salsa while others are not: have to align this
 - [ ] Diagnostic messages need some work (but could use @lrlna input on this already 😇 )
 - [ ] It only includes tests from the spec, but I think it doesn't test every case

Current output:

**Conflicting fields aliased to the same name**
image

**Conflicting arguments selected to the same name**
image

**Conflicting types from different fragments selected to the same name**
image
Here the input is:
```graphql
fragment conflictingDifferingResponses on Pet {
  ... on Dog {
    someValue: nickname
  }
  ... on Cat {
    someValue: meowVolume
  }
}
```


crates/apollo-compiler/src/database/hir.rs
```diff
@@ -853,6 +874,31 @@ impl Value {
     pub fn is_variable(&self) -> bool {
         matches!(self, Self::Variable(..))
     }
+
+    /// Returns `true` if `other` represents the same value as `self`. This is different from the
+    /// `Eq` implementation as it ignores location information.
+    pub fn is_same_value(&self, other: &Value) -> bool {
```
lrlna, line 880: def different than `Eq`, butttt could it be a `PartialEq`? or should we save `PartialEq` for semantic equality?

crates/apollo-compiler/src/database/hir_db.rs
```diff
@@ -147,6 +147,9 @@ pub trait HirDatabase: InputDatabase + AstDatabase {
     /// Return all operation fragment spread fields in a corresponding selection set.
     fn operation_fragment_spread_fields(&self, selection_set: SelectionSet) -> Arc<Vec<Field>>;
 
+    /// Return the fields that `selection_set` selects including visiting fragments and inline fragments.
+    fn flattened_operation_fields(&self, selection_set: SelectionSet) -> Vec<Arc<Field>>;
```
lrlna, line 151: ooooo yea this is super handy; also handy to have on the hir db as well for a few future purposes

crates/apollo-compiler/src/diagnostics.rs
```diff
@@ -265,6 +267,13 @@ pub enum DiagnosticData {
         /// Name of the field
         field: String,
     },
+    #[error("operation must not select different types as the same name `{field}`")]
```
lrlna, line 270: maybe this?

```suggestion
    #[error("operation must not select different types using the same field name `{field}`")]
```

crates/apollo-compiler/src/diagnostics.rs
```diff
@@ -265,6 +267,13 @@ pub enum DiagnosticData {
...
+    ConflictingSelection {
```
lrlna, line 271: not entirely 100% on suggestion, but maybe!

```suggestion
    ConflictingFields {
```

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -1,6 +1,273 @@
...
+            field_a.loc(),
+            format!(
+                "`{}` has type `{}` here",
+                field_a.response_name(),
+                full_type_a
+            ),
+        ))
+        .label(Label::new(
+            field_b.loc(),
+            format!("but the same name has type `{full_type_b}` here"),
```
lrlna, line 50:
```suggestion
            format!("but the same field name has type `{full_type_b}` here"),
```

crates/apollo-compiler/src/database/hir.rs
lrlna: i've forgotten about that, ty!

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -1,6 +1,273 @@
...
+                        },
+                    )
+                    .label(Label::new(
+                        field_a.loc(),
+                        format!(
+                            "field `{}` is selected from field `{}`",
+                            field_a.response_name(),
+                            field_a.name()
+                        ),
+                    ))
```
lrlna, line 252: was thinking some more about this. 

i don't think i follow the diagnostic for the first screenshot titled "Conflicting fields aliased to the same name".

the validation issue is that we have an alias that's bares the same name as another field, and the two of them return different types (and therefore cannot be merged. what if we say something along the lines of "this alias {alias_name} references field {field_name} and returns type {type_definition_name}" for the original aliased reference, and for the field name definition something like this: "{alias_name} is redefined here and returns type {type_definition_name}". The `Note` portion of the diagnostic can then describe the validation rule in a bit more detail.

(field merging is a complicated rule, we are going to have to iterate on these diagnostics a few times, i am sure)

goto-bus-stop: I read this particular rule ([`FieldsInSetCanMerge` 2.b.i.](https://spec.graphql.org/October2021/#FieldsInSetCanMerge())) differently: it specifically checks that two fields that have the same response name (i.e. alias if configured, else the field name) have the same field name. It's for cases like this:
```graphql
query GetUser {
  name: displayName
  name: legalName
}
```
Even if `displayName` and `legalName` have the same types, this is not allowed.

This rule has to exist because it *is* allowed to do
```graphql
query GetUser {
  name
  name
}
```
which just results in the `name` field being selected with response name `name`.

But if a response name `name` is selected twice *from different fields*, that's a conflict.

Differing types are checked by `SameResponseShape`, so that's a separate rule

goto-bus-stop: so we do need to improve the diagnostic message to clarify the above :D

goto-bus-stop:
I think i made it nondeterministic somehow, maybe an iteration over a hashmap somewhere…

crates/apollo-compiler/src/validation/selection.rs
lrlna: ahhhhh! okay, yea let's think of a clearer diagnostic messaging then 🤔 🤔

goto-bus-stop:
Added a recursion test from graphql-js to make sure `flatten_operation_fields` can handle self-referential fragments, but it turns out there were other methods that also have to handle that. `.variables()` is the one that turned up in tests, so I've centralised the implementation in `hir::SelectionSet::variables()`. That method recursively collects all used variables on its own. The other `.variables()` methods just call into `SelectionSet`'s implementation once.

crates/apollo-compiler/src/database/hir.rs
```diff
@@ -1134,6 +1178,48 @@ impl SelectionSet {
         })
     }
 
+    /// Get all variables used in this selection set.
+    pub fn variables(&self, db: &dyn HirDatabase) -> Vec<Variable> {
+        /// Recursively collect used variables. Accounts for self-referential fragments.
```
lrlna, line 1183: does this at all help with used variables validation? feel like it would

crates/apollo-compiler/src/database/hir.rs
```diff
@@ -1205,11 +1315,16 @@ impl Field {
...
     }
 
-    /// Get a reference to the field's name.
+    /// Get the field's name, corresponding to the definition it looks up.
     pub fn name(&self) -> &str {
         self.name.src()
     }
 
+    /// Get the field's response name.
+    pub fn response_name(&self) -> &str {
```
lrlna, line 1324: i don't follow what `response_name` means here? :(

crates/apollo-compiler/src/validation/field.rs
```diff
@@ -220,8 +220,6 @@ pub fn validate_field_definitions(
     diagnostics
 }
 
-// TODO(@lrlna): don't allow this!!
-#[allow(clippy::result_large_err)]
```
lrlna, line 224: ++

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -35,6 +313,10 @@ pub fn validate_selection_set(
 ) -> Vec<ApolloDiagnostic> {
     let mut diagnostics = Vec::new();
 
+    if let Err(diagnostic) = fields_in_set_can_merge(db, selection_set.clone()) {
```
lrlna, line 316: the nittiest nit in the world of nits! we've been doing `db.func_name`, which makes it easier to know visually that it's a db query. shall we stick to that?

```suggestion
    if let Err(diagnostic) = db.fields_in_set_can_merge(selection_set.clone()) {
```

lrlna (approved):
eeee!! this is looking good

do you want to tackle multiple diagnostics per selection set in another pr?

crates/apollo-compiler/src/validation/selection.rs
lrlna: it might? am not sure

crates/apollo-compiler/src/database/hir.rs
lrlna: ok ok that makes sense. can you add an explanation to the doc comment, in case someone else is also confused?

lrlna:
i think still readable! amazing work on this @goto-bus-stop \o/

PR #471, opened 2023-03-10
chore(parser,compiler): check that test numbers are unique

goto-bus-stop:
We use the test numbers for ordering so that the file IDs generated by the compiler are consistent between runs.

Initially did this as an xtask but I think it's better if we don't have to manually run it, so now it's part of the test suite instead. It suggests a new filename if you use a duplicate. <strike>I kept the updates to xtask dependencies in this PR.</strike> pulling it out and rebasing to reduce commit noise

We already had a duplicate number 😇 so I moved that file to the end. That's why there are many file ID changes. This PR should guarantee that we never have to do that again…


lrlna (approved):
~~~ blesssss

PR #472, opened 2023-03-10
chore: remove unnecessary into_iter() calls

goto-bus-stop:
these make clippy angery in rust 1.68


goto-bus-stop:
trivial change so doesnt need review imo.

lrlna:
`angery`,,,, :D :D :D

PR #473, opened 2023-03-10
chore(xtask): update xtask dependencies

goto-bus-stop:
Started doing this for #471 but then decided to go a different route with that PR. I already did the work so we can merge this separately 🤷🏻‍♀️


lrlna (approved):
lgtm

Issue #468, opened 2023-03-02
HIR APIs should implicitly account for type extensions


SimonSapin:
Consider these methods:

```rust
impl TypeDefinition {
    pub fn directives(&self) -> &[Directive] { .. }
    pub fn field(&self, name: &str) -> Option<&FieldDefinition> { .. }
}

impl SchemaDefinition {
    pub fn directives(&self) -> &[Directive] {
    pub fn root_operation_type_definition(&self) -> &[RootOperationTypeDefinition] { .. }
    pub fn query(&self, db: &dyn HirDatabase) -> Option<Arc<ObjectTypeDefinition>> { .. }
    pub fn mutation(&self, db: &dyn HirDatabase) -> Option<Arc<ObjectTypeDefinition>> { .. }
    pub fn subscription(&self, db: &dyn HirDatabase) -> Option<Arc<ObjectTypeDefinition>> { .. }
}

impl ObjectTypeDefinition {
    pub fn directives(&self) -> &[Directive] { .. }
    pub fn fields_definition(&self) -> &[FieldDefinition] { .. }
    pub fn field(&self, name: &str) -> Option<&FieldDefinition> { .. }
    pub fn implements_interfaces(&self) -> &[ImplementsInterface] { .. }
}

impl InterfaceTypeDefinition {
    pub fn directives(&self) -> &[Directive] { .. }
    pub fn fields_definition(&self) -> &[FieldDefinition] { .. }
    pub fn field(&self, name: &str) -> Option<&FieldDefinition> { .. }
    pub fn implements_interfaces(&self) -> &[ImplementsInterface] { .. }
}

impl ScalarTypeDefinition {
    pub fn directives(&self) -> &[Directive] { .. }
}

impl EnumTypeDefinition {
    pub fn directives(&self) -> &[Directive] { .. }
    pub fn enum_values_definition(&self) -> &[EnumValueDefinition] { .. }
}

impl UnionTypeDefinition {
    pub fn directives(&self) -> &[Directive] { .. }
    pub fn union_members(&self) -> &[UnionMember] { .. }
}

impl InputObjectTypeDefinition {
    pub fn directives(&self) -> &[Directive] { .. }
    pub fn input_fields_definition(&self) -> &[InputValueDefinition] { .. }
}
```

(The above is intended to be exhaustive.)

[...]

PR #481, opened 2023-03-10
feat(compiler): rework directive-related HIR methods

SimonSapin:
Part of https://github.com/apollographql/apollo-rs/issues/468

* On HIR types for definitions that can have associated extensions, the previous `directives` method is now called `self_directives`.
* The name `directives` is instead given to new methods that return an iterator and also includes directives on extensions of that definition.
* On HIR types that have directives, add `directive_by_name` and `directives_by_name` that return an `Option` and an iterator respectively. Where applicable, they also consider directives on extensions.

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -146,7 +146,7 @@ impl TypeDefinition {
     /// For repeatable directives, see [`directives_by_name`][Self::directives_by_name] (plural).
     ///
     /// Includes directives on either the type definition or its type extensions,
-    /// like [`all_directives`][Self::all_directives].
+    /// like [`directives`][Self::directives].
     pub fn directive_by_name(&self, name: &str) -> Option<&Directive> {
         self.directives_by_name(name).next()
     }


SimonSapin:
Draft for two open questions:

Either:

* `directives_by_name` returns an iterator (as in the draft). It’s easy enough to call `.next()` on it
* `directive_by_name` (singular) returns an `Option` (implemented as as `.filter(…).next()`). This method is already about making the common case slightly more convenient, so might as well go more in that direction. For repeatable directives callers can still write the `filter` call themselves, as before.
* Return an option, but call it just `directive`. Pluralization is the only thing that separate its name from the other method. (Depending on the choice below.)

Either:

* `all_directives` is a new method with a new name, no breaking change.
* Rename the previous method to `self_directives` and give the `directives` name to the new behavior, because we want that new behavior to be the "default". This is a breaking change but perhaps "not enough" of one: a caller that only iterated the returned slice would still compile but might not realize that the behavior has changed
* Rename to `self_directives` and add `all_directives`. There’s nothing with the old name anymore, forcing all previous callers to actively make a choice

Similar questions apply to other parts of https://github.com/apollographql/apollo-rs/issues/468. I think it’d be best to pick one pattern and apply it consistently.

SimonSapin:
To the question of "should we return an option or an iterator" we landed on "yes" :) And for giving new behavior to existing names where applicable since they’re the shorter names we would have picked if this was the first version of the library.

I’ve updated the PR, I think it’s ready now

lrlna (approved):
🚀 🚀

PR #482, opened 2023-03-14
feat(compiler): implicit extensions in HIR uniquely-named-item APIs

SimonSapin:
Part of https://github.com/apollographql/apollo-rs/issues/468, for components that have a unique name such as fields of object types.

Internally, `IndexMap` is used to map names to `Vec` indices.

Renamed:

* `ObjectTypeDefinition::fields_definition` → `self_fields`
* `ObjectTypeDefinition::implements_interfaces` → `self_implements_interfaces`
* `InterfaceTypeDefinition::fields_definition` → `self_fields`
* `InterfaceTypeDefinition::implements_interfaces` → `self_implements_interfaces`
* `InputObjectTypeDefinition::input_fields_definition` → `self_fields`
* `UnionTypeDefinition::union_members` → `self_members`
* `EnumTypeDefinition::enum_values_definition` → `self_values`

Names freed by the above, redefined with new behavior (to consider extensions) and signature:

* `ObjectTypeDefinition::implements_interfaces() -> impl Iterator`
* `InterfaceTypeDefinition::implements_interfaces() -> impl Iterator`

Behavior changed to consider extensions (no signature change):

* `TypeDefinition::field(name) -> Option`
* `ObjectTypeDefinition::field(name) -> Option`
* `InterfaceTypeDefinition::field(name) -> Option`

New APIs (that all consider extensions):

* `ObjectTypeDefinition::fields() -> impl Iterator`
* `ObjectTypeDefinition::implements_interface(name) -> bool`
* `InterfaceTypeDefinition::fields() -> impl Iterator`
* `InterfaceTypeDefinition::implements_interface(name) -> bool`
* `InputObjectTypeDefinition::self_fields() -> &[_]`
* `InputObjectTypeDefinition::fields() -> impl Iterator`
* `InputObjectTypeDefinition::field(name) -> Option`
* `UnionTypeDefinition::members() -> impl Iterator`
* `UnionTypeDefinition::has_member(name) -> bool`
* `EnumTypeDefinition::values() -> impl Iterator`
* `EnumTypeDefinition::value(name) -> Option`

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -1,6 +1,6 @@
 use std::{
     collections::{HashMap, HashSet},
-    fmt,
+    fmt, hash,
     sync::Arc,
 };
 
@@ -1223,7 +1223,7 @@ impl Field {
     /// Get field's original field definition.
     pub fn field_definition(&self, db: &dyn HirDatabase) -> Option<FieldDefinition> {
         db.find_object_type_by_name(self.parent_obj.as_ref()?.to_string())?
-            .fields_definition()
+            .self_fields()
             .iter()
             .find(|field| field.name() == self.name())
             .cloned()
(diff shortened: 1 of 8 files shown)


SimonSapin:
Draft PR for early feedback on this pattern applied to object type fields. TODO: apply the same pattern to everything in extensions with unique names:

* interface fields
* input object type fields
* object type implemented interfaces
* interface type implemented interfaces
* union members
* enum values

crates/apollo-compiler/src/database/hir.rs
```diff
@@ -1288,6 +1288,92 @@ impl Float {
...
+        extension_index: usize,
+        extension_items: &[Item],
+        name: impl Fn(&Item) -> &str,
+    ) {
+        for (i, item) in extension_items.iter().enumerate() {
+            self.indices
+                .entry(name(item).to_owned())
+                .or_insert((Some(extension_index), i));
+        }
+    }
```
lrlna, line 1337: so this is perhaps dependent on how accurate we want to be in our HIR. An extension should not exist if the original definition doesn't exist. Perhaps this method should also not add an extension if we don't have an existing, matching definition. 

it'd have to check if the definition already exists, and if the definition is of the same type as the proposed extension. Much like quite a few of our recent conversations - our HIR creation is doing a bunch of the same work as validation. This would be adding even more of that work to extensions.

lrlna (commented):
This looks good so far! I would almost suggest trying to implement maybe the object definition extension validation based on this, just to see how smoothly or how much more work validation would be with this.

crates/apollo-compiler/src/database/hir.rs
```diff
@@ -1288,6 +1288,92 @@ impl Float {
...
+impl ByNameWithExtensions {
+    pub(crate) fn new<Item>(self_items: &[Item], name: impl Fn(&Item) -> &str) -> Self {
+        let mut indices = IndexMap::new();
+        for (i, item) in self_items.iter().enumerate() {
+            indices.entry(name(item).to_owned()).or_insert((None, i));
+        }
+        ByNameWithExtensions { indices }
+    }
+
+    pub(crate) fn add_extension<Item>(
```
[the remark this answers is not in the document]
SimonSapin: Well this is not for a single item (e.g. field) but for all items of one extension. (Also this isn’t a public API)

crates/apollo-compiler/src/database/hir.rs
SimonSapin: I don’t understand what you mean. This is a method of `ByNameWithExtensions` which is used in the `ObjectTypeDefinition` struct. If a object type extension extends an object type that doesn’t exist, then its `ByNameWithExtensions` also doesn’t exist and this code isn’t called. This happens if `get_mut` here returns `None`:

https://github.com/apollographql/apollo-rs/blob/3088432300072926635df95072963816f52cdf7b/crates/apollo-compiler/src/database/hir_db.rs#L384

SimonSapin:
I think this is now complete but it felt rather prone to copy-paste errors (doc-comments especially). I’ll read through it again in a few hours after clearing my mind.

lrlna (approved):

PR #484, opened 2023-03-16
feat(compiler): implicit extensions in HIR root operation APIs

SimonSapin:
A collection of loosely related changes, part of https://github.com/apollographql/apollo-rs/issues/468.

HIR no longer creates an implicit root operation if an explicit one exists in a schema extension, even if an object type named `Query`/`Mutation`/`Subscription` exists and the base schema definition doesn’t define an explicit root operation of that type.

`SchemaDefinition::root_operation_type_definition()` was renamed to `self_root_operations`.

`SchemaDefinition::root_operations() -> impl Iterator` was added. It includes root operations from extensions.

`SchemaDefinition::query()`, `mutation()`, and `subscription()` now account for schema extensions, and return `Option<&str>` with the name instead of the definition of the corresponding object type. That definition can be found at `compiler.db.objects().get(name)`. starting % with '%' will be ignored, and an empty message aborts the commit.

As a side effect of that last change these methods can return `Some(name)` for invalid documents where that name doesn’t have a type definition, or it’s not an object type. Previously they would return `None` since there is no object type definition to return. As a result, validation stopped emitting "`query` root operation type must be defined here" diagnostics for such documents. I believe this is correct, since a "not found in this scope" or "root operation type must be of an Object Type" diagnostic was and still is emitted separately.

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -437,11 +437,13 @@ impl OperationDefinition {
 
     /// Get operation's definition object type.
     pub fn object_type(&self, db: &dyn HirDatabase) -> Option<Arc<ObjectTypeDefinition>> {
-        match self.operation_ty {
-            OperationType::Query => db.schema().query(db),
-            OperationType::Mutation => db.schema().mutation(db),
-            OperationType::Subscription => db.schema().subscription(db),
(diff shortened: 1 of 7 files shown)


lrlna (approved):
this looks good! excited to see how this helps us with extension validations

PR #487, opened 2023-03-20
fix(compiler): fix variables in directive arguments being reported as unused

goto-bus-stop:
Operations can use variables in two places: field arguments and directive arguments. Previously we only counted field arguments, so we'd raise an error on things like this:

```graphql
query ($var: Boolean) {
  field @skip(if: $var)
}
```

This PR correctly counts `$var` as used.


lrlna (approved):
good catch!!

PR #488, opened 2023-03-20
feat(compiler): validate that non-repeatable directives are only used once per location

### LINKED ISSUE #246: validation: directive location usage

goto-bus-stop:
Closes #246

### DIFF
--- a/crates/apollo-compiler/src/diagnostics.rs
+++ b/crates/apollo-compiler/src/diagnostics.rs
@@ -262,6 +262,13 @@ pub enum DiagnosticData {
         /// The source location where the directive that's being used was defined.
         directive_def: Option<DiagnosticLocation>,
     },
+    #[error("non-repeatable directive {name} can only be used once per location")]
+    UniqueDirective {
+        /// Name of the non-unique directive.
+        name: String,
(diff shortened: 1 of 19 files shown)


lrlna (approved):
✔️ ✔️ ✔️ 

could you also double check that we got similar enough tests to the ones in the [js implementation ](https://github.com/graphql/graphql-js/blob/main/src/validation/__tests__/UniqueDirectivesPerLocationRule-test.ts)?

goto-bus-stop:
first use of our new HIR methods that account for extensions 😇

PR #489, opened 2023-03-21
feat(compiler): add meta types to compiler context 

### LINKED ISSUE #439: meta fields in compiler's context

lrlna:
closes #439 

Introduces introspection types, built in scalars and built in directives in a form of a source file to the compiler. This allows us to properly validate operations with introspection fields, as well as provide more detailed diagnostics when working with other built in types. 

This PR introduces a new source type - `BuiltIn`. `SourceType` now also has appropriate `is_{ty}` methods:
- `is_built_in`
- `is_document`
- `is_query`, and
- `is_schema`.

There are also a few changes to the `HirDatabase` in type queries. Implicit and built-in types should not be returned as part of the regular queries to the compiler context, so we added an additional queries to separate between user provided types and built in types:
- `db.object_types` returns only user supplied object types
- `db.object_types_with_built_ins` returns user supplied and built in (introspection) object types
- `db.enums` returns only user supplied enum types
- `db.enums_with_built_ins` returns user supplied and built in (introspection) enum types

Finding field by name on particular type does search through implicit fields (`__type`, `__typename` and `__schema`). So `type_definition.field(&compiler.db, "__type")` will return the implicit field's field definition.


lrlna:
oh goodie, diagnostic's source cache is non deterministic!

lrlna:
Here is what diagnostics would look like for an operation that cannot have `__type` introspection, like a mutation:

image

or an object that's not defined as the query root type:
image

crates/apollo-compiler/test_data/diagnostics/0016_schema_with_built_in_scalar_definition.txt
```diff
@@ -1,28 +1,58 @@
...
+            },
+            Label {
+                location: DiagnosticLocation {
+                    file_id: FileId {
+                        id: 31,
+                    },
+                    offset: 37,
+                    length: 3,
+                },
+                text: "`Int` redefined here",
```
lrlna, line 33: this change is actually incorrect and i still need to fix. because built in definitions are now loaded via a source, we exclude any duplicate scalar definitions simply by nature of index map. with redefined built-in scalars, the diagnostic should indicate to remove the redefined scalar.

crates/apollo-compiler/src/lib.rs
```diff
@@ -103,6 +103,20 @@ impl ApolloCompiler {
...
+            let file_id = FileId::new();
+            let mut sources = self.db.source_files();
+            sources.push(file_id);
+            let implicit_tys = fs::read_to_string(path).expect("built in types could not be added");
+            self.db
+                .set_input(file_id, Source::built_in(path.to_path_buf(), implicit_tys));
+            self.db.set_source_files(sources);
+        }
+    }
+
```
lrlna, line 119: these will be added to pre-computed hir, which is what we want

crates/apollo-compiler/src/lib.rs
```diff
@@ -1141,7 +1158,19 @@ type User
...
+                "__Field",
+                "__InputValue",
+                "__EnumValue",
+                "__Directive",
+                "Mutation",
+                "Product",
+                "Query",
+                "Review",
+                "User"
+            ],
```
lrlna, line 1173: this might be good do discuss: should be explicitly exclude introspection types from `db.object_types` query? on one hand, this is useful for any sort of validation, linting, execution or future introspection work, on the other hand, this is not super useful for object type manipulation which may detract

goto-bus-stop: oh hm. my initial reaction is that it's surprising for them to be included…

I don't know if this makes sense, but my expectation would be that only explicitly requesting one of the builtins returns it:
```rust
db.object_types() // → returns Mutation, Product, Query, Review, User
db.find_object_type_by_name("__Schema") // → returns the __Schema type
```

crates/apollo-compiler/src/database/hir.rs
```diff
@@ -1974,6 +1998,25 @@ impl ObjectTypeDefinition {
         )
     }
 
+    /// Introspection fields that must not be returned as part of `fields` method.
+    pub(crate) fn implicit_fields(&self, db: &dyn HirDatabase) -> &[FieldDefinition] {
+        let is_root_query = db
+            .schema()
+            .root_operations()
+            .any(|op| op.operation_ty().is_query() && op.named_type().name() == self.name());
+        if is_root_query {
```
goto-bus-stop, line 2007: Is it possible to provide the correct list when we create the HIR node for this, so we dont have to filter it down after the fact?

lrlna: right, so i am not sure how to get this right while we are creating the hir node. running a check to see if this object type is defined as the the root `query` type requires a db query, which causes a cycle. any ideas?

crates/apollo-compiler/src/lib.rs
lrlna: the implicit fields (not types) are meant to be excluded from any of the field queries, but i am wondering if having them be returned with `object_type.field("__typename")` might indeed be more useful

lrlna: all right, removed built in types from the `object_types()` query. still not 100% certain `object_type.field()` should return implicit fields 🤔 🤔 gonna think about it some more

crates/apollo-compiler/src/database/hir.rs
goto-bus-stop: Hmm. I think what bothers me a little in this method is that we have this `Arc<ImplicitFields>` list in each node when it isn't really specific to that node, especially since it's also kinda the wrong list of fields in the case of non-root operation types.

what if we didn't store `implicit_fields` in the object types themselves, but had a `db.implicit_fields()` query that returns the appropriate implicit fields for a given object type? The implementation wouldnt need to be very different most likely

lrlna: i've had something like `db.implicit_fields()` before, iirc there were some lifetime issues with that. let me try again!

crates/apollo-compiler/src/built_in_types.graphql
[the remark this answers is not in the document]
lrlna: ta! updated

crates/apollo-compiler/src/lib.rs
```diff
@@ -103,6 +103,20 @@ impl ApolloCompiler {
...
     }
 
+    // This adds the introspection type system and any built-in graphql types.
+    fn add_implicit_types(&mut self) {
+        let path = Path::new("src/built_in_types.graphql");
+        if self.db.source_file(path.to_path_buf()).is_none() {
+            let file_id = FileId::new();
+            let mut sources = self.db.source_files();
+            sources.push(file_id);
+            let implicit_tys = fs::read_to_string(path).expect("built in types could not be added");
```
SimonSapin, line 113: I think this expect the process current directory to be the apollo-compiler source tree, which typically won’t be the case when compiler is used as library dependency. Consider using `include_str!` instead

crates/apollo-compiler/src/database/hir.rs
```diff
@@ -176,21 +176,36 @@ impl TypeDefinition {
...
-    pub fn field(&self, name: &str) -> Option<&FieldDefinition> {
+    pub fn field(&self, db: &dyn HirDatabase, name: &str) -> Option<&FieldDefinition> {
         match self {
-            Self::ObjectTypeDefinition(def) => def.field(name),
-            Self::InterfaceTypeDefinition(def) => def.field(name),
+            Self::ObjectTypeDefinition(def) => def
+                .field(name)
+                .or(def.implicit_fields(db).iter().find(|f| f.name() == name)),
+            Self::InterfaceTypeDefinition(def) => {
+                def.field(name).or(if def.implicit_field().name() == name {
```
SimonSapin, line 185: May use `Option::or_else` so the implicit field lookup only runs when needed?

crates/apollo-compiler/src/database/hir.rs
SimonSapin: Is it actually a cycle? I’d expect `db.schema()` to be able to complete without constructing HIR for object types

crates/apollo-compiler/src/database/hir.rs
```diff
@@ -2556,6 +2621,10 @@ impl UnionTypeDefinition {
             .add_extension(next_index, ext.union_members(), UnionMember::name);
         self.extensions.push(ext);
     }
+
+    pub fn implicit_field(&self) -> &FieldDefinition {
```
SimonSapin, line 2625: Having a method named `implicit_field` (singular) feels weird to me. Why is it not plural and returning a slice? Alternatively, if this is always `__typename` and nothing else, why not call it `typename_implicit_field`?

crates/apollo-compiler/src/database/sources.rs
```diff
@@ -62,7 +105,7 @@ impl Source {
     }
 }
 
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, Ord, PartialOrd)]
```
SimonSapin, line 108: Instead of ordering and sorting, what if `NEXT` started at 1 and the built-in source always had ID zero?

crates/apollo-compiler/src/database/hir.rs
lrlna: yea, it calls `db.object_types` to relate root operations to the existing object types 

https://github.com/apollographql/apollo-rs/blob/aaf8d14e58fe94ee5375391d33bc9ad6bcaa08e3/crates/apollo-compiler/src/database/hir_db.rs#L655

crates/apollo-compiler/src/database/sources.rs
lrlna: we could do that too, instead of `Ord` and `PartialOrd` we'd have a `impl From<u64> for FileId` instead to cast the always-zero ID, which seems like a smaller API change anyways?

crates/apollo-compiler/src/database/hir.rs
lrlna: > maybe i'll create a private hir type that we can filter afterwards, but that will mean that that hir type will still have an Arc<Vec<FieldDefinition>>

nope this also doesn't work hm hm hmmm

crates/apollo-compiler/src/lib.rs
lrlna: yea ok, let's have `object_type.field()` return implicit fields too

crates/apollo-compiler/src/database/sources.rs
lrlna: ah, but it doesn't matter as the Debug output which we write to the test output still spits the file paths out in an indeterminate order. i think easier to have `Ord` and `PartialOrd` then.

crates/apollo-compiler/test_data/diagnostics/0016_schema_with_built_in_scalar_definition.txt
lrlna: 🙏 fixed

crates/apollo-compiler/src/database/hir.rs
lrlna: you're right, it is weird for it to be singular. let's do `implicit_fields` and returning a slice for all of them. we can try something else in the future if this doens't work out

crates/apollo-compiler/src/database/sources.rs
```diff
@@ -88,6 +137,6 @@ impl FileId {
     /// Reset file ID back to 0, used to get consistent results in tests.
```
yanns, line 137: to 1 ?

yanns:
For info, I've started using this change, and it's working well! 💯

crates/apollo-compiler/src/database/hir.rs
```diff
@@ -2556,6 +2610,10 @@ impl UnionTypeDefinition {
             .add_extension(next_index, ext.union_members(), UnionMember::name);
         self.extensions.push(ext);
     }
+
+    pub fn implicit_fields(&self) -> &[FieldDefinition] {
```
lrlna, line 2614:
```suggestion
    pub(crate) fn implicit_fields(&self) -> &[FieldDefinition] {
```

PR #491, opened 2023-03-23
chore(compiler): use salsa invoke for ease of navigating

lrlna:
Using `salsa::invoke` to indicate where the function comes from in our `hir_db`. Makes it _soooo_ much easier to click through to the function definition and just to read where the function is.


goto-bus-stop (approved):
guuuut 🥳 

maybe we should also do this for the ast and input databases? While the implementations are in the same module, something like `#[salsa::invoke(source_code)]` does still work to help jump-to-implementation find the function

lrlna:
that's a great idea, everything gets a `salsa::invoke`!

PR #492, opened 2023-03-24
Add a test for operation field type resolution v.s. type extensions

SimonSapin:
This type resolution in https://github.com/apollographql/router/issues/2710 was the motivation to start work on
https://github.com/apollographql/apollo-rs/issues/468


lrlna (approved):
tysm!

Issue #248, opened 2022-06-09
validation: unused & undefined directive definition


lrlna:
when a directive is not used or not defined, the compiler should create a diagnostic. this can probably be done alongside #246 
- [x] undefined directive
- [ ] unused directive
- [x] directive cannot reference itself directly or indirectly

goto-bus-stop:
Some indirect self-referential cases:
```graphql
extend scalar String @deprecatedType(reason: "use OurCustomString instead")
directive @deprecatedType(reason: String!) on OBJECT | INTERFACE | ENUM | SCALAR | UNION

directive @loopA(arg: Boolean @loopB) on ARGUMENT_DEFINITION
directive @loopB(arg: Boolean @loopC) on ARGUMENT_DEFINITION
directive @loopC(arg: Boolean @loopA) on ARGUMENT_DEFINITION

input InputObject {
  name: String @wrong
}
directive @wrong(input: InputObject) on FIELD
```

lrlna:
unused directives are not part of the spec, so this is done!

PR #494, opened 2023-03-27
feat(compiler): validate indirectly self-referential directives

goto-bus-stop:
Ref #248 

When defining a directive, that directive may not also be used on any type or directive references by the new directive. We already checked simple things like:
```graphql
directive @selfReference(arg: Int @selfReference) on ARGUMENT_DEFINITION
```

This PR recurses into types and directives referenced by a directive definition.

Examples of new error cases:

**Indirect recursion through directives** (reports one error per directive definition)
```graphql
directive @loopA(arg: Boolean @loopB) on ARGUMENT_DEFINITION
directive @loopB(arg: Boolean @loopC) on ARGUMENT_DEFINITION
directive @loopC(arg: Boolean @loopA) on ARGUMENT_DEFINITION
```
image

**Indirect recursion through a type extension**
```graphql
extend scalar String @deprecatedType(reason: "use OurCustomString instead")
directive @deprecatedType(reason: String!) on OBJECT | INTERFACE | ENUM | SCALAR | UNION
```
image

**Recursion through a field in an input object**
```graphql
enum Enum { NON_RECURSIVE }
extend enum Enum { RECURSIVE @wrong }
input Name { value: String!, type: Enum }
input InputObject { name: Name }
directive @wrong(input: InputObject) on INPUT_FIELD_DEFINITION
````
image
(in the future we could trace it through all the types that make up the recursion & point to each step)

### CODE REVIEW

File: crates/apollo-compiler/src/validation/directive.rs [hunk 1]
```diff
@@ -7,6 +7,130 @@ use crate::{
...
+        self.0.pop();
+    }
+}
+
+/// This struct just groups functions that are used to find self-referential directives.
+/// The way to use it is to call `FindRecursiveDirective::check`.
+struct FindRecursiveDirective<'a> {
+    db: &'a dyn ValidationDatabase,
+}
+impl FindRecursiveDirective<'_> {
```
lrlna, line 38: wondering if we can use this sort of thing for fragments forming cycles (self referential fragments, if you will) 🤔

lrlna:
ah, you too are having issues with circle ci! i had to log out of circle, log back in, and then push a commit again. than the config.yaml "magically" appeared

Issue #496, opened 2023-03-27
Validation: used variable is reported as unused


yanns:
## Description

A variable that is used is being reported as unused.

### Steps to reproduce

- schema:
```
type Product {
  attributes(includeNames: [String!]): [String!]
}

type Query {
  topProducts(first: Int = 5): [Product]
}
```

- query:
```
query Test($attributeName: String!) {
  topProducts { 
    attributes(includeNames: [$attributeName])
  } 
}
```

- variables:
```
{ "attributeName": "color" }
```

### Expected result

The variable `attributeName` is used, and the query should be reported as valid.

### Actual result

Validation error:
```
Error: unused variable: `attributeName`
   ╭─[query.graphql:1:12]
   │
 1 │ query Test($attributeName: String!) {
   │            ───────────┬───────────  
   │                       ╰───────────── this variable is never used
───╯
```

goto-bus-stop:
I see, variables used inside lists or input objects are not yet accounted for. Thanks for the report!

PR #497, opened 2023-03-27
fix(compiler): fix used variables validation for lists and objects

yanns:
Closes https://github.com/apollographql/apollo-rs/issues/496

Opening this PR to get an early feedback if the general approach is ok.


goto-bus-stop:
Thanks for getting started on this 😁 I think we'll need to put a `.variables()` method on the `Value` struct itself. Variables could be used inside lists, but also inside nested lists, and inside input objects. So it needs to recurse to handle deeply nested situations like
```graphql
query ($v: Int) {
  field(opts: {
    prop: [[2], [$v], [4]],
  })
}
```

goto-bus-stop:
I think the vecs are fine—typically there won't be that many variables, and it's not super obvious to me if allocating a boxed iterator for every intermediate step is gonna make much of a difference compared to allocating temporary vecs for only the intermediate steps that actually contain variables. Plus we are already doing things like this in many places 🤷🏻‍♀️

PR #501, opened 2023-03-28
Include built-in enum types in db.type_system()

SimonSapin:
Like built-in object types are included. This is consistent with the implicit input for built-in types being skipped when using set_type_system_hir()


lrlna (approved):
✨

PR #502, opened 2023-03-28
fix(compiler): field_definition works for interface types

### LINKED ISSUE #498: HIR field_definition method returns None for all interface Fields
## Description

`field_definition` on HIR `Field` type does not work on interface fields. This seems to be because the method assumes the field is on an object (it uses `find_object_type_by_name`): 

https://github.com/apollographql/apollo-rs/blob/main/crates/apollo-compiler/src/database/hir.rs#L1382-L1388

Is this correct behavior? Should I be using another method to get the field definition for fields on interface types?

zackangelo:
Modify `field_definition` to also work for interface types. 

Verified that test fails with original code: 

```
failures:

---- database::hir::tests::field_definition stdout ----
thread 'database::hir::tests::field_definition' panicked at 'field definition not found on creature.name selection', crates/apollo-compiler/src/database/hir.rs:4000:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    database::hir::tests::field_definition
```

Fixes #498

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -1380,11 +1380,14 @@ impl Field {
 
     /// Get field's original field definition.
     pub fn field_definition(&self, db: &dyn HirDatabase) -> Option<FieldDefinition> {
-        db.find_object_type_by_name(self.parent_obj.as_ref()?.to_string())?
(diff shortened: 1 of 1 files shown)


lrlna (approved):
tysm!

PR #505, opened 2023-03-29
feat(compiler): validate circular input objects

lrlna:
A circular input object is defined as one that references itself in a non-nullable fashion, all other instances of self-referencing are permitted. 

This is permitted,
```graphql
input Example {
  self: Example
  value: String
}
```
where as this is not,
```graphql
input Example {
  self: Example!
  value: String
}
```

image

### DIFF
--- a/crates/apollo-compiler/src/validation/input_object.rs
+++ b/crates/apollo-compiler/src/validation/input_object.rs
@@ -2,8 +2,71 @@ use std::{collections::HashMap, sync::Arc};
 
 use crate::{
     diagnostics::{ApolloDiagnostic, DiagnosticData, Label},
-    hir, ValidationDatabase,
+    hir,
+    validation::RecursionStack,
+    ValidationDatabase,
 };
+
+// Implements [Circular References](https://spec.graphql.org/October2021/#sec-Input-Objects.Circular-References)
+// part of the input object validation spec.
+struct FindRecursiveInputValue<'a> {
+    db: &'a dyn ValidationDatabase,
+}
+
+impl FindRecursiveInputValue<'_> {
+    fn input_value_definition(
+        &self,
+        seen: &mut RecursionStack<'_>,
+        def: &hir::InputValueDefinition,
+    ) -> Result<(), hir::InputValueDefinition> {
+        let ty = def.ty();
+        return match ty {
+            hir::Type::NonNull { ty, loc: _ } => match ty.as_ref() {
+                // NonNull type followed by Named type is the one that's not allowed
+                // to be cyclical, so this is only case we care about.
+                //
+                // Everything else may be a cyclical input value.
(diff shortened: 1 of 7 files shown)


crates/apollo-compiler/src/validation/input_object.rs
```diff
@@ -2,9 +2,72 @@ use std::{collections::HashMap, sync::Arc};
...
+    ) -> Result<(), hir::InputValueDefinition> {
+        let ty = def.ty();
+        return match ty {
+            hir::Type::NonNull { ty, loc: _ } => match ty.as_ref() {
+                // NonNull type followed by Named type is the one that's not allowed
+                // to be cyclical, so this is only case we care about.
+                //
+                // Everything else may be a cyclical input value.
+                hir::Type::Named { name, loc: _ } => {
+                    if seen.contains(name) {
```
goto-bus-stop, line 30: In the recursive directive definition code I also checked `seen.first()` here because otherwise, an input type `B` referencing an input type `A` where `A` is cyclical would report an error on the definition of `B`. For example:

```graphql
input A {
  cycle: A!
}
input B {
  field: A!
}
```
image

Here the error you get on `B` is quite confusing, and unnecessary as you already have the error on `A`, and that one shows what you need to solve.

So I think we should do the same here (just return `Ok(())` if there is a cycle in a different input type than we started with)

goto-bus-stop (changes requested):

crates/apollo-compiler/src/validation/input_object.rs
```diff
@@ -2,9 +2,72 @@ use std::{collections::HashMap, sync::Arc};
...
+                        return Err(def.clone());
+                    } else {
+                        if let Some(def) = self.db.find_input_object_by_name(name.into()) {
+                            self.input_object_definition(seen.push(name.into()), def.as_ref())?
+                        }
+                        Ok(())
+                    }
+                }
+                hir::Type::NonNull { ty: _, loc: _ } | hir::Type::List { ty: _, loc: _ } => Ok(()),
```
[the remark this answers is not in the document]
lrlna: surprised clippy didn't suggest this tbh is clippy slacking off?

crates/apollo-compiler/src/validation/input_object.rs
lrlna: but what happens if you have two inputs that reference each other? if you only check the first item in the stack, you end up with something like this being valid:

```graphql
input A {
  cycle: B!
}

input B {
  cycle: A!
}
```

lrlna: does that ever happen with directives? ^

lrlna:
@goto-bus-stop pls re-review ^__^

goto-bus-stop (approved):
🎉

PR #483, opened 2023-03-14
feat(compiler): Add fragment definition validation

gocamille:
### Description
This PR adds support for validating fragment spread type existence, validating fragments on composite types, and validating that a fragment is used in `ApolloCompiler`.

#### Methods added
- Added a new method `validate_fragment_type_condition` which validates that a type condition is valid: it exisits in the schema, and is on a composite type
- Added a new method `validate_fragment_used` which validates that a fragment is used in a query.

### To Do
- [ ] Move the inline tests to the `test_data` directory

### Related Issue
closes #309

Issue #310, opened 2022-09-21
validation: fragment spread


lrlna:
- [ ] fragment spread target must be defined, [spec](https://spec.graphql.org/October2021/#sec-Fragment-spread-target-defined)
- [ ] fragment spreads must not form cycles, [spec](https://spec.graphql.org/October2021/#sec-Fragment-spreads-must-not-form-cycles)
- [ ] fragment spread is possible, [spec](https://spec.graphql.org/October2021/#sec-Fragment-spread-is-possible) (this one is very gnarly)
    - [ ] object spreads in object scope, [spec](https://spec.graphql.org/October2021/#sec-Object-Spreads-In-Object-Scope) (equivalently gnarly) 
    - [ ] abstract spreads in object scope, [spec](https://spec.graphql.org/October2021/#sec-Abstract-Spreads-in-Object-Scope) (you guessed it, gnarly)
    - [ ] object spreads in abstract scope, [spec](https://spec.graphql.org/October2021/#sec-Abstract-Spreads-in-Object-Scope) (......)
    - [ ] abstract spreads in abstract scope, [spec](https://spec.graphql.org/October2021/#sec-Abstract-Spreads-in-Abstract-Scope) (....)

PR #511, opened 2023-04-04
feat(compiler): validate fragment spread is possible

goto-bus-stop:
Part of #310

Hopefully it doesn't collide too much with #483, lets merge that PR before this one if possible.

This PR checks that fragment spreads `... frag` and inline spreads `... { }` can be applied to the selection set they are used in.

I added a `FragmentSelection` HIR node which gets created on the fly and represents either type of fragment (named or inline). This way the validation is only implemented once.

This is ready for review but there are a few more things to check before merging:
- [x] Recursion in the type system should not cause `get_possible_types` to loop
- [x] Need some tests for inline spreads without a type condition, like when it's only used to apply a directive `... @skip(if: $bool) {}`


lrlna:
@goto-bus-stop i merged #483 just now ^__^ I really like this `FragmentSelection` thing you have going here, would be good to get that with some of that with the type condition validation.

apologies for all the test file conflicts 🙈 we were _really_ not use semantically correct graphql with _any_ of our fragments

goto-bus-stop:
I think this PR also reports errors when you use a fragment on a non-composite type, need to address that. It's not meant to happen for the validation in this PR and already done by the validations in Camille's PR.

goto-bus-stop:
OK I'm gonna keep the "double" errors for now. It's not that trivial to avoid. Ideally we would also silence errors if you have something like this:

```graphql
fragment wrong on Int {
  fieldA
  fieldB
}
```

Currently we'll report that you can't put a fragment on an Int, and then also report that the fields `fieldA` and `fieldB` do not exist, but this is noise for the end user. So it would be better to not emit those diagnostics if something else is already wrong. But that's a bit complicated to attack I think and it's mostly about UI which we can tweak and improve after finishing everything else.

crates/apollo-compiler/src/validation/fragment.rs
```diff
@@ -1,8 +1,175 @@
...
+
+    diagnostics.extend(db.validate_directives(
+        inline.directives().to_vec(),
+        hir::DirectiveLocation::InlineFragment,
+    ));
+
+    diagnostics.extend(db.validate_fragment_type_condition(
+        inline.type_condition().map(|t| t.to_string()),
+        inline.loc(),
+    ));
```
lrlna, line 166: i wonder if perhaps some of the "double" reporting can be avoided if `validate_fragment_type_condition` returns a result. we can then only proceed to `validate_fragment_selection` if the previous validation is `Ok`?

lrlna (approved):
>Currently we'll report that you can't put a fragment on an Int, and then also report that the fields fieldA and fieldB do not exist, but this is noise for the end user. So it would be better to not emit those diagnostics if something else is already wrong. But that's a bit complicated to attack I think and it's mostly about UI which we can tweak and improve after finishing everything else.

yeaaaa i hear you. i think there are a couple of places in field validation where we have validation rules return a result, and only proceed after a successful validation. we could use the same strategy in a few other places, but i think that's a bit of a larger task we can do in one go after all the validation rules are complete

crates/apollo-compiler/src/validation/fragment.rs
goto-bus-stop: hmmm thats worth a shot 🤔

goto-bus-stop: did this by checking `validate_fragment_type_condition().is_empty()

PR #512, opened 2023-04-04
breaking(compiler): add parser's token limiting to compiler's API

### LINKED ISSUE #509: include parser's token limiting in compiler
Token limits are currently not propagated to compiler's API, and it be useful to trigger the same limiting from the compiler.

lrlna:
Adding the token API that already works nicely in the parser to the compiler. This *breaks* the current recursion_limit API we have in the compiler, but hopefully makes it a lot more ergonomic in terms of usage.

Here is how you'd set both limits:
```rust
let schema = r#"
    query {
        a {
            a {
                a {
                    a
                }
            }
        }
    }
"#;
let mut compiler = ApolloCompiler::new().token_limit(22).recursion_limit(10);
compiler.add_document(schema, "schema.graphql");
```

closes #509

### DIFF
--- a/crates/apollo-compiler/src/database/ast.rs
+++ b/crates/apollo-compiler/src/database/ast.rs
@@ -30,12 +30,14 @@ fn ast(db: &dyn AstDatabase, file_id: FileId) -> SyntaxTree {
     // which we don’t want to re-parse.
     let input = db.input(file_id).text();
 
-    let parser = ApolloParser::new(&input);
-    let parser = if let Some(limit) = db.recursion_limit() {
-        parser.recursion_limit(limit)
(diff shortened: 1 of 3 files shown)


goto-bus-stop (approved):
I think the fact that we have panic conditions in at least 3 of our compiler methods suggests we should think about a builder pattern or some other type-level enforcement of those constraints 🤔 
Maybe good to think about in the near ish future. LGTM in the mean time!

lrlna:
> we should think about a builder pattern or some other type-level enforcement of those constraints 🤔

agreed!

lrlna:
opened #513

PR #515, opened 2023-04-04
chore(encoder): add missing clone impl to encoder types

zackangelo:
Noticed that `Clone` was derived on some but not all encoder types when I tried to `.clone()` an `encoder::ObjectDefinition` instance. 

Not sure if there's a good reason, but ahead and opened this PR if it just happened to be an omission.


goto-bus-stop (approved):
Gonna say it was just an omission. I think we'll want to make apollo-encoder more capable in the future (also modifying nodes in more ways than we support now) and it'd be important for everythhing to be cloneable then. Thanks for the contribution!

PR #518, opened 2023-04-06
feat(compiler): validate fragment definitions do not contain cycles

goto-bus-stop:
Previously i added a test in the `ok` folder that asserted the opposite of this PR, because I was looking at a different part of the spec, where fragments are applied. When fragment spreads are used you just need to account for recursion so you don't overflow the stack, there's no check for recursive spreads there. That check is instead done when *defining* a fragment, and that's what this PR adds 😄 

It works the same as other recursion validations, with a `RecursionStack` and only reporting a validation error if the first item matches (so you only get relevant errors).


crates/apollo-compiler/src/validation/fragment.rs
```diff
@@ -48,7 +50,72 @@ pub fn validate_fragment_definitions(
...
+                    detect_fragment_cycles(db, inline.selection_set(), visited)?;
+                }
+                _ => {}
+            }
+        }
+
+        Ok(())
+    }
+
+    // Split RecursionStack initialisation for lifetime reasons
```
lrlna, line 102: :''''''')

lrlna (approved):

PR #519, opened 2023-04-06
chore(parser): remove recursion in field parsing, fixes #514

### LINKED ISSUE #514: `recursion_limit` intended use is unclear
## Description
This issue may not actually be a bug but i've found the `recursion_limit` option on the parser to not work the way I expect. My expectation is that this feature is guarding against stack overflows and should correlate to the number of stack frames deep the parser is while parsing but this doesn't appear to be the case.

For example, the following query fails when the recursion limit is set to 5:
```graphql
    {
        field
        field2
        field3
        field4
        field5
        field6
    }
```

```rust
    let parser = Parser::new(QUERY)
        .recursion_limit(5)
        .parse()
```

When digging into the code, it appears that sibling fields [happen sequentially](https://github.com/apollographql/apollo-rs/blob/e55a76538ae37b929f0a6c49544ba6d33c67c103/crates/apollo-parser/src/parser/grammar/selection.rs#L74) (allocating a `field(p)` stack, returning, and then allocating another one for the next field) meaning that I would expect the maximum recursion depth to be ~2 for this query but instead I get `Iter([ERROR@75:75 "parser limit(5) reached" ])`.

I admit this could be my misunderstanding this feature but I'd appreciate some clarity on how this feature should be used.

### Steps to reproduce
You can see the full example [here.](https://github.com/apollographql/apollo-rs/compare/main...allancalix:apollo-rs:recursion-limit)

### Expected result
Sibling fields should reset the recursion depth after returning for a single selection set.

### Actual result
Sibling fields do not reset the recursion depth and selection sets with many fields (possibly an arbitrarily large number given the ability to use aliases) hit the recursion limit.

### Environment

- Operating system and version: macos
- Shell (bash/zsh/powershell): fish
- `apollo-rs` crate: parser
- Crate version: `HEAD` as of 2022/04/04

goto-bus-stop:
The `selection::selection_set` parser already supports parsing multiple fields, so it looks like this recursion is unnecessary. Tests pass right now but I'll also run the fuzzer for a bit to make sure.


lrlna (approved):
lgtm

lrlna:
how is the fuzzer? is it happy?

goto-bus-stop:
Heh I couldn't get to the bottom of the issue where we continued parsing after a recursion limit was reached, but I think I just put the check in the wrong place yesterday. Got it right with fresh eyes this morning 😇

lrlna:
excellllllent ^_^

PR #520, opened 2023-04-12
fix(compiler): add limit exceeded diagnostic

lrlna:
Noticed the diagnostics for exceeded limits were noting them as "syntax errors", when they are in fact "limit exceeded" errors.


goto-bus-stop (approved):
👍🏻 lgtm

PR #524, opened 2023-04-18
feat(compiler): validate object type extensions

goto-bus-stop:
Closes https://github.com/apollographql/apollo-rs/issues/476

Progress:
* [x] The named type must already be defined
* [x] the already defined type must be an Object type.
* [x] The fields of an Object type extension must have unique names; no two fields may share the same name.
* [x] Any fields of an Object type extension must not be already defined on the original Object type.
* [x] Any non-repeatable directives provided must not already apply to the original Object type.
* [x] Any interfaces provided must not be already implemented by the original Object type.
* [x] The resulting extended object type must be a super-set of all interfaces it implements.

[spec](https://spec.graphql.org/October2021/#sec-Object-Extensions.Type-Validation)

### DIFF
--- a/crates/apollo-compiler/src/diagnostics.rs
+++ b/crates/apollo-compiler/src/diagnostics.rs
@@ -221,6 +221,8 @@ pub enum DiagnosticData {
     },
     #[error("the required argument `{name}` is not provided")]
     RequiredArgument { name: String },
+    #[error("`implements` declarations must be unique")]
+    DuplicateImplementsInterface { name: String },
     #[error(
         "Transitively implemented interfaces must also be defined on an implementing interface or object"
     )]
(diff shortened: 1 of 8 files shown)


goto-bus-stop:
> Any interfaces provided must not be already implemented by the original Object type.

graphql-js doesnt seem to validate this 🤔 and at first glance it doesn't seem like a great rule to me. but maybe i misunderstand it. I'll check their testsuite if they have any tests for this.

I assume it's meant to report on this:
```graphql
type Object implements Intf {
  a: Int
}
extend type Object implements Intf {
}
```

lrlna (approved):
lesssgo

goto-bus-stop:
OK, to follow up on my comment above, that is indeed meant to be illegal, and graphql-js does validate it, I was just using the wrong method 😇 . Now this does too 👍🏻

PR #525, opened 2023-04-18
feat(parser): add SyntaxTree::token_limit(), similar to recursion_limit()

SimonSapin:
This enables finding out how many tokens were present in a succesful parse, which can be useful to choose where to set the limit.

Currently if no limit is set the counter is disabled entirely and this method will return `None`. So this use case requires setting the limit to `usize::MAX` or some other large number (which is different from not setting a limit). Perhaps this should be changed to always have a counter to let the new method not return an `Option`?


lrlna (approved):
merci bien!

lrlna:
> Perhaps this should be changed to always have a counter to let the new method not return an `Option`?

I don't see a problem with it not being an `Option`. But @goto-bus-stop wrote this feature, and I am not sure if they had special considerations here. Will let them have a say here!

goto-bus-stop:
I think using a default limit of `usize::MAX` is good (or any sufficiently high limit, but MAX definitely fits that bill). Internally it doesnt matter if it's an option or not, but if it's public it should not be wrapped I think. This is a reasonable use case regardless of whether a limit is actually configured.

SimonSapin:
Oh I didn’t realized you’d already replied. The latest push makes the limit non-optional with a default of `usize::MAX`.

PR #535, opened 2023-04-25
feat(compiler): validate executable documents do not contain type definitions, fixes #409

### LINKED ISSUE #409: Validation: forbidden kind of definition for restricted inputs
When using `create_schema` or `update_schema`, the input string is expected not to contain any executable definition. Similarly with `create_executable` or `update_executable` with executable definitions.

If such misplaced definitions exist, the parser will happily create AST node for them, and HIR construction will happily ignore these nodes. Validation needs to actively looks for relevant AST nodes (if it doesn’t already).

goto-bus-stop:
This doesn't check the opposite, so you can put executable definitions in a schema document ... the spec doesn't appear to prohibit that as far as I can tell.

we have to check the `source_type` is `SourceType::Executable` before running this check, as `SourceType::Document` sources are also run through this `validate_executable` function.

### DIFF
--- a/crates/apollo-compiler/src/database/sources.rs
+++ b/crates/apollo-compiler/src/database/sources.rs
@@ -6,7 +6,7 @@ use std::{
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum SourceType {
     Schema,
-    Query,
+    Executable,
     Document,
     BuiltIn,
 }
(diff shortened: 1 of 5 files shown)


crates/apollo-compiler/src/database/sources.rs
```diff
@@ -28,12 +28,12 @@ impl SourceType {
...
 
-    /// Returns `true` if the source type is [`Query`].
+    /// Returns `true` if the source type is [`Executable`].
     ///
-    /// [`Query`]: SourceType::Query
+    /// [`Executable`]: SourceType::Executable
     #[must_use]
-    pub fn is_query(&self) -> bool {
-        matches!(self, Self::Query)
+    pub fn is_executable(&self) -> bool {
```
goto-bus-stop, line 35: renaming for consistency

lrlna: 🙏

lrlna (approved):
need to make sure we remember that this requires a release of `apollo-parser` too

goto-bus-stop:
ya good shout added to the commit title 👍🏻

Issue #319, opened 2022-09-30
create multiple errors when lexing a string value


lrlna:
When lexing a String Value we can have three different error types:
- unterminated string
- unexpected escape character
- unexpected line terminator

If there is a unexpected escape character or unexpected line terminator *and also* the string value is unterminated, we only get the unterminated string error message, as the errors get overridden. 

For example:

```
let gql = "\"\n hello"
let lexer = Lexer::new(gql);
```

produces only `ERROR@0:7 "unterminated string value" "hello` error, where it would be really helpful for users to also have the unexpected line terminator error as well.

goto-bus-stop:
This is also relevant for escape sequences:

```graphql
"\u123 \u0008 \uZZZ"
```

Now you get an error for the whole string, it would be better if we could point to only the incorrect parts.

PR #541, opened 2023-04-28
fix(encoder,parser): escape and unescape strings 

goto-bus-stop:
This PR improves the string value printing in apollo-encoder. It tries to sensibly decide between a multi-line block string, a single-line block string, and a non-block string depending on the contents it needs to print. It escapes special characters when writing non-block strings. When writing block strings, only `"""` is escaped as `\"""` so the block isn't ended prematurely.

The parser's `String::from(StringValue)` implementation also un-escapes characters.


goto-bus-stop:
I'm not really happy with this but I think I should revisit later rather than spend more time on it now. I'd like to more or less do the string parsing from scratch and also address #319 at that time. Atleast this should improve the string printing in apollo-encoder so it doesn't produce invalid output anymore.

crates/apollo-encoder/src/field.rs
```diff
@@ -332,9 +332,7 @@ mod tests {
             r#"  spaceCat(
     "The type of treats given in space"
     treat: Treat,
-    """
-    Optimal age of a "space" cat
-    """
+    """Optimal age of a "space" cat"""
```
lrlna, line 335: so i think we should always format block string values with leading and trailing new line. There is a bit on that in [spec here](https://spec.graphql.org/October2021/#sel-FAFdPTDPAADVCA5ke)

lrlna (changes requested):
a comment on formatting block string values. otherwise looks good!

yea, string parsing needs a brand new coat of paint. will probably be easier to solve multi-error lexing issue too.

crates/apollo-encoder/src/field.rs
goto-bus-stop: i took this idea from graphql-js, i kinda like it for short single-line strings. but i also like having consistent output style in the encoder 🤔

lrlna: iiiiiinteresting they write _one_ thing in the spec, but do something _else_

lrlna: but sure, sounds good, let's go with what graphql-js does then

lrlna (approved):

crates/apollo-encoder/src/field.rs
goto-bus-stop: i read the spec example as only a style recommendation for multi-line strings. for single-line strings you could do something else in theory 😄 

i removed it now (it's also simpler to not do it). There is one reason to do it i think: we decide whether to print a string as a block string or a non-block string based on its contents, and for single-line strings that include " but no other escapable characters, i think it does look nicer to use a single-line block string. but we can add that back if we add configurable readability things in the future.

goto-bus-stop:
the parser unescaping implementation that i have here does the wrong thing for block strings btw. As block strings don't actually support escapes except for `\""""`. I'd improve that in the future string improvements PR.

Issue #523, opened 2023-04-18
Consider removing `impl Default for ApolloCompiler`?


SimonSapin:
The type does have a parameterless `new()` constructor that matches the shape of `Default::default()`, but the returned compiler isn’t useful or "really initialized" until we add some inputs. Compared to an explicit `new()` call, `Default` enables generic code (unaware of the specificities of the type) to implicitly initialize a value.

The first version of https://github.com/apollographql/router/pull/2883 accidentally made deserialization produce an empty compiler by removing `OnceCell` in a `#[serde(skip)]` struct field. Would it be preferable to not implement `Default`? That `derive(Deserialize)` impl would have not compiled, requiring instead explicit initialization code (where it’s not as easy to forget to add inputs)

lrlna:
we can remove the `Default` impl

PR #542, opened 2023-04-28
breaking(compiler): remove impl Default for ApolloCompiler

dariuszkuc:
Remove unnecessary `impl Default for ApolloCompiler`.

Resolves: #523


lrlna (approved):
thank you!

Issue #543, opened 2023-05-03
Stack overflow on `validate_executable` for fragment cycles


gmac:
## Description

The following fails with a `fatal runtime error: stack overflow`...

```rust
let source = r#"
   { ...fragA }
   fragment fragA on Query { ...fragA }
"#;
let mut compiler = apollo_compiler::ApolloCompiler::new();
let file_id = compiler.add_executable(source, "");
compiler.db.validate_executable(file_id);
```

Interestingly, this only happens when the fragment spread is the initial selection in both scopes. Either of these permutations where the spread is preceded by another selection avoids the stack overflow:

```graphql
{ sfoo ...fragA }
fragment fragA on Query { ...fragA }
```

Or...

```graphql
{ ...fragA }
fragment fragA on Query { sfoo ...fragA }
```

### Steps to reproduce

See snippet above.

### Expected result

Should get an error rather than a stack overflow.

### Actual result

Stack overflow.

### Environment

- Operating system and version: macos, 13.3.1
- Shell (bash/zsh/powershell): zsh
- `apollo-rs` crate: `apollo-compiler`
- Crate version: `0.8.0`

PR #544, opened 2023-05-04
Prevents recursion for operations referencing fragments with cycles

allancalix:
Addresses #543

The core problem is that the `is_introspection` method is not context-aware and if it's in the context of a fragment spread for itself will continue to recurse until the stack overflows. Fundamentally fixing this issue with `is_introspection` would require a some added context like exists with the recursion limit counter in the parser or changing the function signature to control whether or not to recurse into a fragment spread.

I tried to avoid reimplementing fragment definition detection by reusing what already exists and should mean the result is memoized when `validate_fragment_definition` is called during `validate_executable`.

I considered performing fragment definition validation first and skipping operation validation if there are cycles present but there are some valuable validations that do not depend on traversing the operation.

### DIFF
--- a/crates/apollo-compiler/src/database/document.rs
+++ b/crates/apollo-compiler/src/database/document.rs
@@ -199,6 +199,21 @@ pub(crate) fn operation_inline_fragment_fields(
     Arc::new(fields)
 }
 
+pub(crate) fn operation_fragment_references(
+    db: &dyn HirDatabase,
+    selection_set: SelectionSet,
+) -> Arc<Vec<Arc<FragmentDefinition>>> {
+    let fields = selection_set
+        .selection()
+        .iter()
+        .filter_map(|sel| match sel {
+            Selection::FragmentSpread(fragment_spread) => Some(fragment_spread.fragment(db)?),
+            _ => None,
+        })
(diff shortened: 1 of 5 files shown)


goto-bus-stop:
thanks for the PR! with this implementation, am I understanding it right that `is_introspection()` can still recurse? I think it would be most robust if we can bail out of recursive loops inside `is_introspection()`. We have a few places where we do that, for example in `variables()`: https://github.com/apollographql/apollo-rs/blob/770c26c67412f567396a35cd38d0fb2c33578508/crates/apollo-compiler/src/database/hir.rs#L1237-L1280

allancalix:
It can definitely still recurse and if the caller can guarantee there are no fragment cycles (like happens here) it should be fine to use. I agree that ultimately `is_introspection` needs to be adjusted to either not recurse at all or keep additional context for detecting cycles.

My goal with this PR was to prevent stack overflows because from what I can tell adjust `is_introspection` would be a bit of a lift.

goto-bus-stop (approved):
OK, happy to take this first, and I can look at adjusting `is_introspection` soon. thanks! 🥳

PR #547, opened 2023-05-10
apollo-parser@0.5.2

goto-bus-stop:
thought we'd be going to 0.6.0 but looking at the changes i don't think that's necessary 🤔


lrlna (approved):

lrlna:
ahhh it's nice when we don't do breaking changes hehe

PR #550, opened 2023-05-10
Validate coercible input

lrlna:
This PR validates if `Value` is of correct type. I am trying to avoid doing actual coercions here, and instead verifying if types in the type system match the value's types. We can do the actual coercion in execution.

To do this correctly recursively (for object and list values) and to be able to show diagnostics on specific values, this PR adds location information to `Value` type in the hir. I think I should implement a partial eq specifically so we don't compare on loc info. Thoughts?

The diff is huge, because I brought over all the tests from graphql-js. Glad I did, because there were a ton of nitty gritty things!

Some examples of diagnostics:

**Values of incorrect type**
image

image

image

**Missing required args in a object value**
image

**Value inside an object value does not exist on the specified input object**
(the display on this is weird, because I (for the moment) have not figured out how to do this outside of the value_of_correct_type function, and I don't have the arg location there anymore, so it points to the value
image

**Enum variant that's being selected doesn't extist**
image


crates/apollo-compiler/src/validation/value.rs
```diff
@@ -0,0 +1,263 @@
...
+                _ => diagnostics.push(unsupported_type!(db, val, ty)),
+            },
+            // When expected as an input type, only valid Unicode string input
+            // values are accepted. All other input values must raise a request
+            // error indicating an incorrect type.
+            // When expected as an input type, any string (such as "4") or
+            // integer (such as 4 or -4) input value should be coerced to ID
+            Value::String { .. } => match &type_def {
+                TypeDefinition::ScalarTypeDefinition(scalar) => {
+                    if scalar.is_int() || scalar.is_float() || scalar.is_boolean() {
```
goto-bus-stop, line 108: the other branches all check if the assignment is _valid_ and otherwise report an error, but this checks if the assignment is _invalid_. it's a bit surprising and i think its more obvious to check all the ways that strings *can* be used instead of the ways they cannot

crates/apollo-compiler/src/validation/value.rs
```diff
@@ -0,0 +1,263 @@
...
+            },
+            Value::Object { value: ref obj, .. } => match &type_def {
+                TypeDefinition::InputObjectTypeDefinition(input_obj) => {
+                    let undefined_field = obj.iter().find_map(|(name, value)| {
+                        let is_undefined = !input_obj.fields().any(|f| f.name() == name.src());
+                        if is_undefined {
+                            return Some((name, value));
+                        }
+                        None
+                    });
```
goto-bus-stop, line 202:
```suggestion
                    let undefined_field = obj.iter().find(|(name, value)| {
                        !input_obj.fields().any(|f| f.name() == name.src())
                    });
```
I think this should work

crates/apollo-compiler/src/validation/value.rs
```diff
@@ -0,0 +1,263 @@
...
+                        }
+
+                        obj.iter().for_each(|(name, v)| {
+                            let type_def = ty.type_def(db.upcast());
+                            if name.src() == f.name() {
+                                if let Some(_def) = type_def {
+                                    value_of_correct_type(db, ty, v, var_defs.clone(), diagnostics)
+                                }
+                            }
+                        })
```
goto-bus-stop, line 256: it looks like this `for_each` is only meant to actually do the work for 1 of the fields? maybe we can use `.find()` to get at the thing we need instead?

goto-bus-stop:
this makes sense to me! the nesting makes it look quite complicated but you can just read it top to bottom and understand it so that's great. i just have a few code nitpicks 😇

crates/apollo-compiler/src/validation/value.rs
lrlna: oh yea, definitely, let me change this

lrlna: ohhhh i remember why - this is like this because of custom scalars. custom scalars can be coerced into a string, unless otherwise specified. 

so when you have a directive like this, it passes this validation.
```graphql
directive @join__field(graph: join__Graph, requires: join__FieldSet, provides: join__FieldSet) on FIELD_DEFINITION

scalar join__FieldSet

enum join__Graph {
  REVIEWS
}

type Query {
  retailPrice: String @join__field(graph: REVIEWS, requires: "price")
}
```

lrlna: specifically looking at `requires: "price"`

crates/apollo-compiler/src/validation/value.rs
goto-bus-stop: we're already inside a loop, so `f` is a single field in the input object type here. there should be only one match unless the value has duplicate keys, which is already caught by other validations

PR #555, opened 2023-05-12
apollo-compiler@0.9.0

lrlna:
for when we land #550 & checking for pre-validated hir input

### DIFF
--- a/crates/apollo-compiler/Cargo.toml
+++ b/crates/apollo-compiler/Cargo.toml
@@ -17,7 +17,7 @@ categories = [
 edition = "2021"
 
 [dependencies]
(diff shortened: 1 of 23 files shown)


crates/apollo-compiler/Cargo.toml
```diff
@@ -17,7 +17,7 @@ categories = [
 edition = "2021"
 
 [dependencies]
-apollo-parser = { path = "../apollo-parser", version = "0.5.0" }
+apollo-parser = { path = "../apollo-parser", version = "0.5.3" }
```
lrlna, line 20: should i bump this to `0.5.3`? or just leave it at `0.5.0`; the latest is currently within the semver range, so probably `0.5.3` is not necessary eh?

PR #563, opened 2023-05-22
fix(compiler): add a specific UndefinedVariable diagnostic

goto-bus-stop:
The diagnostic message for undefined variables read like this:

> `varName` type not found in this document

which is a bit wrong. This adds a specific message for variables so it
instead reports:

> variable `varName` is not defined


lrlna (approved):
lgtm!

PR #564, opened 2023-05-22
feat(compiler): add as_$type() methods to hir::Value

goto-bus-stop:
We usually expected people to just `match` on the HIR values, but that's
a little verbose if you just need one type. For example the router uses
this custom function:

```rust
fn as_string(value: &hir::Value) -> Option<&String> {
    if let hir::Value::String(string) = value {
        Some(string)
    } else {
        None
    }
}
```

It actually does help with readability and that's even more true now
that you have to take into account the location information on hir values.

This PR adds a couple of methods like this to the `hir::Value` type.
It's quite similar to `serde_json::Value`.

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -1154,6 +1154,45 @@ impl Value {
     pub fn is_null(&self) -> bool {
         matches!(self, Self::Null { .. })
     }
+
+    /// Returns an `i32` if the value is a number and can be represented as an i32.
+    #[must_use]
+    pub fn as_i32(&self) -> Option<i32> {
+        i32::try_from(self).ok()
+    }
+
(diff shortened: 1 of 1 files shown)


crates/apollo-compiler/src/database/hir.rs
```diff
@@ -1154,6 +1154,65 @@ impl Value {
     pub fn is_null(&self) -> bool {
         matches!(self, Self::Null { .. })
     }
+
+    /// Returns an `i32` if the value is a number and can be represented as an i32.
+    #[must_use]
+    pub fn as_i32(&self) -> Option<i32> {
```
lrlna, line 1160: `as_` methods are usually from borrowed to borrowed values, but we are probably ok here right? There is even a precedent in [`Duration::as_secs`](https://doc.rust-lang.org/core/time/struct.Duration.html#method.as_secs) in std 🤔

PR #565, opened 2023-05-23
fix(compiler): non-nullable variables should be accepted for nullable args

### LINKED ISSUE #558: Mandatory variable cannot be used for optional argument since apollo-compiler 0.9.0
## Description

Since the update to apollo-compiler 0.9.0, we have new validation errors that do not seem valid.

### Steps to reproduce

Considering the schema like:
```
type Query {
  category(
    "Queries with specified ID"
    id: String,

    "Queries with specified key"
    key: String): Category
}
```
The following query:
```
          query CategoryByKey($key: String!){
            category(key: $key) {
              id, key, name(locale:"en")
            }
          }
```
with the key as variable is not valid anymore.
It fails with `variable `key` cannot be used for argument `key` as their types mismatch`

Changing the query to:
```
          query CategoryByKey($key: String){
            category(key: $key) {
              id, key, name(locale:"en")
            }
          }
```

works.

### Expected result

Not 100% sure, but a mandatory variable can be used for an optional argument.

lrlna:
fixes #558 

also noticed a glitch for `null` being provided for a non-nullable input values - those should also error.

### DIFF
--- a/crates/apollo-compiler/src/validation/value.rs
+++ b/crates/apollo-compiler/src/validation/value.rs
@@ -142,13 +142,17 @@ pub fn value_of_correct_type(
                         if var_def.ty().name() != type_def.name() {
                             diagnostics.push(unsupported_type!(db, val.clone(), ty));
                         } else if let Some(default_value) = var_def.default_value() {
-                            value_of_correct_type(
-                                db,
-                                var_def.ty(),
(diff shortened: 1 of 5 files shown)


goto-bus-stop (approved):
yay, lgtm!!

PR #566, opened 2023-05-23
apollo-compiler@0.9.2

goto-bus-stop:
[chglg](https://github.com/apollographql/apollo-rs/blob/apollo-compiler%400.9.2/crates/apollo-compiler/CHANGELOG.md#092---2023-05-23)


lrlna (approved):
chglg lg thx ^_^

PR #567, opened 2023-05-23
fix(compiler): nonnullable list variables in nullable args

lrlna:
similar to #565


goto-bus-stop (approved):
hopefully we caught everything this time 😇

PR #570, opened 2023-05-26
breaking(compiler): rename `SelectionSet::merge` to `concat`

goto-bus-stop:
As brought up by Sylvain. The comment already talked about concatenation instead of merging. The `merge` name suggests it's doing field merging which is not happening. We may add an actual field merging method in the future.

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -1454,7 +1454,7 @@ impl SelectionSet {
     ///
     /// This does not deduplicate fields: if the two selection sets both select a field `a`, the
     /// merged set will select field `a` twice.
-    pub fn merge(&self, other: &SelectionSet) -> SelectionSet {
+    pub fn concat(&self, other: &SelectionSet) -> SelectionSet {
         let mut merged: Vec<Selection> =
             Vec::with_capacity(self.selection.len() + other.selection.len());
         merged.append(&mut self.selection.as_ref().clone());
--- a/crates/apollo-compiler/src/validation/selection.rs
+++ b/crates/apollo-compiler/src/validation/selection.rs
@@ -104,7 +104,7 @@ pub(crate) fn same_response_shape(
         }
         // 6. Assert: typeA and typeB are both composite types.
         (def_a, def_b) if def_a.is_composite_definition() && def_b.is_composite_definition() => {
-            let merged_set = field_a.selection_set().merge(field_b.selection_set());
+            let merged_set = field_a.selection_set().concat(field_b.selection_set());
             let fields = db.flattened_operation_fields(merged_set);
             let grouped_by_name = group_fields_by_name(fields);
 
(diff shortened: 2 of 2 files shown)


lrlna:
i feel like every instance of this in the spec uses the term `merge`, what’s our reasoning for deviating from that terminology?

pcmanus:
For context, the current `SelectionSet::merge` does not "coalesce" fields of the same name/response name or recurse in any way. And so it behaves this way (in pseudo-code):
```
SelectionSet::merge({ a b } , { a }) == { a b a }
SelectionSet::merge({ t { x } } , { t { y } }) == { t { x } t { y } }
```

This is certainly something that can have it's uses, but note that even for simple examples, it easily produce somewhat odd-looking/inefficient results. It follows that it is imo often not really what you want when manipulating selection sets programmatically.

I also don't think that this si what the spec really refers to when it mentions merging. More precisely, merging is used 2 contexts:
1. for validation, in [Field Selection Merging](https://spec.graphql.org/draft/#sec-Field-Selection-Merging), but `FieldsInSetCanMerge` has both the notion of "coalescing" fields with same response names (step 1 of that method) and do so recursively.
2. for execution, in [merging selection sets](https://spec.graphql.org/draft/#sec-Value-Completion.Merging-Selection-Sets). But what this section illustrates is that:
    ```graphql
    {
      me {
        firstName
      }
      me {
        lastName
      }
    }
    ```
    is executed as:
    ```graphql
    {
      me {
        firstName
        lastName
      }
    }
    ```
    That is, this is explicitely about specifying that for execution, things executes by "coalescing" fields with the same name and doing the same recursively to their sub-selections (note that the [`MergeSelectionSets`](https://spec.graphql.org/draft/#MergeSelectionSets()) method of this section does not applies to selection sets, it applies to a set of `fields` _with the same name_ and while it does not recurse by itself (only appending the sub-selections of those same-named fields), this is a bit deceiving because in practice, the result of `MergeSelectionSets` is fed to `ExecutedSelectionSet` ([point 5.c here](https://spec.graphql.org/draft/#sec-Value-Completion)) which passes it in turn to `CollectFields` ([point 1 here](https://spec.graphql.org/draft/#ExecuteSelectionSet())) and that will group the fields with the same names and eventually call again `MergeSelectionSets` on them; so there is recursivity). To state the obvious, calling the `me` resolver twice in the example above would be incorrect execution and this is what merging is here to specify.

Long story short, when the spec talk of merging, I'm fairly confident it's really refering to the notion of merging fields with the same name (at the same level) and doing so recursively in the sub-selections. 

But spec aside, I also think this behavior (of actually merging same named field together and recursively) is more generally useful and what I would expect from a merge function, because as you manipulate selection sets, it maintains them in a more readable and easy to work with state.

And to clarify, I mean that I'd expect a `merge` to be a different function than the current one, one that does:
```
SelectionSet::merge2({ a b } , { a }) == { a b }
SelectionSet::merge2({ t { x } } , { t { y } }) == { t { x y } }
```

Anyway, long story short, the idea that led to this ticket is that:
1. I suggest that eventually a method with the behaviour of `SelectionSet::merge2` above should be added (not the goal of this ticket obviously but ...).
2. that when both the existing "merge" and such new "merge2" methods exists, then it would be better/more intuitive if the latter "merge2" is the one actually called `merge`. But that's only possible if the current `SelectionSet::merge` method is renamed differently, hence this ticket.

And fwiw, `concat` is only a suggestion for renaming. I feel that name convey well what the method does, it concats the 2 sub-select and no more, but something like `merge_without_coalescing` or the like could be another option.

And of course, not renaming (and using another name for a future `merge2`) is also an option...

goto-bus-stop:
I agreed with the `concat` naming as in the doc comment for the `merge` method, I already basically wrote "this does not merge, but concatenate" 😄 

the thing that this method actually does is not described in the spec.

lrlna (approved):
all right, feel to merge @goto-bus-stop!

Issue #562, opened 2023-05-19
"Long cannot be represented by a Int value" since apollo-compiler 0.9.0


yanns:
## Description

Since the update to apollo-compiler 0.9.0, we have new validation errors that do not seem valid.

### Steps to reproduce

We have a schema with variables represented as `Long!`
Some queries using an `Int!` value for it.

### Expected result

Before the update to apollo-compiler 0.9.0, the values are accepted.
It's possible to send `1` for a `Long!` variable.

### Actual result

The queries are rejected with the following error:
```
Long cannot be represented by a Int value
```

Those values should be valid.

lrlna:
hey @yanns! we didn't have value validation in `0.8.0` which is why validation was passing. `Long` isn't a GraphQL type that's in the spec, so we don't have a special validation for it, but I see how you'd want to coerce an `Int` to `Long`... How do you handle this scalar in your schema? Do you define it separately? Do you do something specific with it on the execution side of things? And lastly, before you handle your schema in Rust, how did you go about validating an operation with a `Long` in the other language(s)?

yanns:
Thanks for the feedback!
I was not aware that `Long` isn't a GraphQL standard type.
It's a GraphQL type defined in [sangria](https://github.com/apollographql/apollo-rs/issues/562#issuecomment-1556655624)

For the moment, we don't even expose it in the schema. It "just" works.
But based on your information, a validator sticking to the specs should not accept it.

I could add a `scalar Long` in the schema SDL.
But I don't know how we could define the coercion from an `Int` from the schema.

I could also add this scalar manually when instantiating the `ApolloCompiler` in rust. Do you know if I could also define the coercion with this?

goto-bus-stop:
maybe we should just accept any int/float/string/bool value as input for a custom scalar in apollo-compiler?

yanns:
> maybe we should just accept any int/float/string/bool value as input for a custom scalar in apollo-compiler? you would be allowed to pass `true` as a `Long`, but it seems better than not being able to use it at all.
> 
> there could be apollo-compiler APIs to set up which spec type a custom scalar is related to to tighten up the validation, but there will be many cases where you don't actually have that information available while using the compiler.

Yes that's a good point.
The next step (resolver) would then fail if needed. And the resolver should have all information to know it if can accept / parse some inputs.

lrlna:
> maybe we should just accept any int/float/string/bool value as input for a custom scalar in apollo-compiler
[...]

PR #575, opened 2023-06-01
feat(compiler): accept any scalar type for custom value validation

lrlna:
A possible solution for #562. Would be really curious to hear what other people think!

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -2632,6 +2632,11 @@ impl ScalarTypeDefinition {
         self.extensions.push(ext);
     }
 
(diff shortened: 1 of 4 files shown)


goto-bus-stop:
would having an `.is_custom()` method on scalars make sense? so we can check `.is_int() || .is_custom()` instead of listing all the other possibilities

lrlna:
@goto-bus-stop we can do that too. it'd look like this - https://github.com/apollographql/apollo-rs/pull/575/commits/ecf8b077c0fecc11dbe713cd402d879759e723d0

goto-bus-stop (approved):
thanks, I think I prefer this ^^

PR #577, opened 2023-06-05
feat(compiler): add type name to `UndefinedField` error message

goto-bus-stop:
This adds a bit more context to the "main" error message for
UndefinedField diagnostics. It was already there in the pretty-printed
version but for use cases like an HTTP backend, it's nice to have as
much context as possible in the main message, so it can just dump that
as a JSON string.


lrlna (approved):
lgtm

PR #579, opened 2023-06-09
Add `schema.root_operation_name(OperationType)` HIR helper method

SimonSapin:
I’ve recently needed this in a couple different places of Apollo Router. (For now I’ve used the explicit `match` instead.)


lrlna (approved):
yepp totally makes sense

PR #582, opened 2023-06-15
fix: schema extension without root fields should skip empty {} block

dariuszkuc:
Per [GraphQL spec](https://spec.graphql.org/draft/#sec-Schema-Extension)

>Schema extensions without additional operation type definitions must not be followed by a { (such as a query shorthand) to avoid parsing ambiguity.

Currently generated definition cannot be parsed by `apollo-compiler`.

```
extend schema @foo {
}
```


goto-bus-stop (approved):
super, thank you!

PR #585, opened 2023-06-20
fix(compiler): accept any input object / list AST as custom scalar value

goto-bus-stop:
per https://github.com/graphql/graphql-spec/issues/688#issuecomment-584010837, custom scalars can take any value, including input objects.


goto-bus-stop:
The tests on fed-next pass with this so will merge to unblock.

PR #586, opened 2023-06-20
breaking(compiler): improve type condition handling on inline fragments

### LINKED ISSUE #584: Unwanted inline fragment type condition inference
## Description

`hir::InlineFragment::type_condition` may return a type even if no type condition syntax was used.

### Steps to reproduce

```rust
    #[test]
    fn type_condition_is_not_inferred() {
        let mut compiler = ApolloCompiler::new();
        let input = r#"
            type Query {
                a: Int
            }
            { ... { a } }
        "#;
        let id = compiler.add_document(input, "");
        let op = compiler.db.find_operation(id, None).unwrap();
        let hir::Selection::InlineFragment(inline_fragment) = 
            &op.selection_set().selection()[0]
            else { panic!("expected inline fragment") };
        // the `on <name>` syntax was not used:
        assert_eq!(inline_fragment.type_condition(), None);
    }
}
```

### Expected result

`type_condition()` should return `None`, based on the absence of type condition syntax in the string input.

### Actual result

The HIR contains an inferred synthetic type condition based on type definitions in the input:

```rust
thread 'database::hir::tests::type_condition_is_not_inferred' panicked at 'assertion failed: `(left == right)`
  left: `Some("Query")`,
 right: `None`', crates/apollo-compiler/src/database/hir.rs:4422:9
```

The code responsible for this is:

https://github.com/apollographql/apollo-rs/blob/6a0b13f191cc77f0cfc76913e86b0a223f427d32/crates/apollo-compiler/src/database/hir_db.rs#L1576-L1579

### Context

In my opinion this behavior is undesirable. The information of what the original input looked like is lost, whereas the type inference information is already available through `InlineFragment::parent_type`.

Adding an (accurate) type condition that wasn’t there before is semantically correct, but my current use case involves serializing parts of a HIR back to GraphQL syntax with precise control of the generated strings. These strings are use as HashMap keys (to identify `@defer` sub-selections) and need to match exactly strings generated separately by TypeScript code inside the Router’s query planner.

goto-bus-stop:
Fixes #584 and some things @SimonSapin brought up in Slack 

- `InlineFragment::type_condition()` now returns `None` if a type condition was not explicitly written.
- `InlineFragment::parent_type()` returns the type definition of the selection set that an inline fragment is used in
- `InlineFragment::type_def()` returns the type definition for the fragment itself: the type condition if given, or else the selection set that it's used in
- `validate_fragment_type_condition` does not take an `Option` anymore. This used to only happen for inline fragments without a type condition, in theory, but that would never actually happen because we inferred the type condition if it wasn't set. So we can get rid of that case entirely and skip the validation for inline fragments without type conditions.

### DIFF
--- a/crates/apollo-compiler/src/database/hir.rs
+++ b/crates/apollo-compiler/src/database/hir.rs
@@ -1746,44 +1746,95 @@ impl InlineFragment {
         self.type_condition.as_ref().map(|t| t.src())
     }
 
-    /// Get the type this fragment is spread onto.
+    /// Get the type this inline fragment is spread onto.
+    ///
+    /// This method does not look up type conditions, so in the example below `b.parent_type()`
+    /// returns the type for "Intf", not "X". To look up the type that an inline fragment *applies*
+    /// to, use [`InlineFragment::type_def`][InlineFragment::type_def].
     ///
     /// ## Examples
     /// ```rust
     /// # fn main() { example().expect("unexpected None") }
     /// # fn example() -> Option<()> { use apollo_compiler::*;
+    /// # use apollo_compiler::validation::ValidationDatabase;
     /// let mut compiler = ApolloCompiler::new();
     /// compiler.add_type_system(r#"
-    ///     type X {
+    ///     interface Intf {
     ///         subField: Int!
     ///     }
+    ///     type X implements Intf {
+    ///         subField: Int!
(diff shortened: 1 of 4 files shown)


SimonSapin (approved):
Looks great, thank you!

PR #591, opened 2023-07-10
Create EOF token with empty data

allancalix:
tl;dr Make consuming the token stream's data produce an identical string to the original input of the lexer.

I was working on a streaming processor using the streaming lexer and noticed that the `Token::Eof` has `"EOF"` as a string hard-coded as it's data. This doesn't appear to be used anywhere (though is maybe a breaking change?) and the tests still pass without this.

The formatter for `Debug` prefixes the location with `EOF` so the debugging output is not effected. The practical effect of this is that __consuming a token stream's data does not produce the same output as the original input.__

### DIFF
--- a/crates/apollo-parser/test_data/parser/err/0041_operation_definition_with_missing_selection_set.txt
+++ b/crates/apollo-parser/test_data/parser/err/0041_operation_definition_with_missing_selection_set.txt
@@ -1,10 +1,10 @@
-- DOCUMENT@0..19
-    - OPERATION_DEFINITION@0..19
+- DOCUMENT@0..16
+    - OPERATION_DEFINITION@0..16
         - OPERATION_TYPE@0..5
             - query_KW@0..5 "query"
         - WHITESPACE@5..6 " "
         - NAME@6..16
             - IDENT@6..16 "__typename"
-        - ERROR@16..19 "EOF"
+        - ERROR@16..16 ""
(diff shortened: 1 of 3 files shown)


crates/apollo-parser/test_data/parser/err/0041_operation_definition_with_missing_selection_set.txt
```diff
@@ -1,10 +1,10 @@
...
-    - OPERATION_DEFINITION@0..19
+- DOCUMENT@0..16
+    - OPERATION_DEFINITION@0..16
         - OPERATION_TYPE@0..5
             - query_KW@0..5 "query"
         - WHITESPACE@5..6 " "
         - NAME@6..16
             - IDENT@6..16 "__typename"
-        - ERROR@16..19 "EOF"
+        - ERROR@16..16 ""
```
allancalix, line 8: Not sure what to make of this here, the `OPERATION_DEFINITION` is now showing the correct character span (the source is 16 characters long) but I don't know why this error message is empty.

goto-bus-stop: It's an error token, not an error message, so IMO it's correct like this

goto-bus-stop (approved):
I think this was used in the past but now isn't, so its a good bit of cleanup. thanks!

PR #627, opened 2023-08-22
experiment: apollo-wasm crate with TypeScript bindings

EverlastingBugstopper:
# `apollo-wasm`

This crate (in `./crates/apollo-wasm`) provides a WASM build for `apollo-compiler` and generates TypeScript/JavaScript bindings for it.

## Prerequisites

1. Install [`rustup`](https://rustup.rs)
1. Install [`wasm-pack`](https://rustwasm.github.io/wasm-pack/installer/)
1. Run `rustup target add wasm32-unknown-unknown`

## Build

`wasm-pack build` will build the WASM package and the TS/JS bindings and output them to `./pkg`.

## Usage

Currently the API exposes a single `validate_document` function that takes in a document and the path to a document. The path can be the empty string if it is not a document from the file system or if it does not have a name.


lrlna (approved):
Averrryyyy! Thank you so much 🙏

PR #628, opened 2023-08-23
feat: support NO_COLOR when emitting diagnostics

### LINKED ISSUE #499: reporting diagnostic with no colors
`apollo-compiler` is using [`ariadne`](https://github.com/apollographql/apollo-rs/blob/main/crates/apollo-compiler/Cargo.toml#L21) to report diagnostics.

Those reports are beautiful 🤩 , and colorful.
If someone wants to use those reports inside a log entry in json format, it'd be great to remove those color information:

Screenshot 2023-03-28 at 10 44 44

It seems that enabling a crate feature might sufficient to support [`NO_COLOR`](https://no-color.org/):
- https://github.com/zesterer/ariadne#cargo-features
- https://crates.io/crates/concolor

WDYT of that?

EverlastingBugstopper:
fixes #499 by enabling the `auto-color` feature of the `ariadne` dependency which supports `NO_COLOR` and detects if a TTY is attached before coloring the diagnostic output.

I'm happy to put this behind an optional feature in `apollo-rs` as well if that is desired, though I think enabling it by default is useful enough to leave it around - interested in thoughts.

### DIFF
--- a/crates/apollo-compiler/Cargo.toml
+++ b/crates/apollo-compiler/Cargo.toml
@@ -18,7 +18,7 @@ edition = "2021"
 
 [dependencies]
 apollo-parser = { path = "../apollo-parser", version = "0.6.0" }
-ariadne = "0.3.0"
+ariadne = { version = "0.3.0", features = ["auto-color"] }
 indexmap = "2.0.0"
 rowan = "0.15.5"
 salsa = "0.16.1"


goto-bus-stop (approved):

goto-bus-stop:
Yes default is perfect I think. thanks!

lrlna (approved):
I think this should work for the current use case in Rover @EverlastingBugstopper. I know @goto-bus-stop has had some ideas of doing come `cfg` styling specific to diagnostics to ensure what we output is still legible, but imo that could totally be a follow up PR in the near future!

PR #631, opened 2023-08-25
Support validating standalone operations without a known schema

### LINKED ISSUE #629: feat: run a version of validation in `apollo-compiler` that does not validate against a parsed schema
feature request: in `apollo-compiler`, validate a list of operations without having access to the schema and emit error diagnostics for malformed operations along with the HIR of successfully parsed operations

goto-bus-stop:
Fixes #629 

This adds a `compiler.db.validate_standalone_executable(file_id)` function that only runs those validations that do not require a schema. Including variable usage, name uniqueness, fragment usage, things like that. I think "standalone" is a decent name but open to bikeshedding.

Also includes a fix for name collisions in executable documents--these actually did not raise an error before.

There are many changes in the test data because this PR sorts the diagnostics by location. There are no changes in the actual diagnostics reported (as test data all uses the validation function that checks against schema.

- [x] Forgot about some nested selection set validations, eg spreading a named fragment that's not defined

### DIFF
--- a/crates/apollo-compiler/src/validation/validation_db.rs
+++ b/crates/apollo-compiler/src/validation/validation_db.rs
@@ -445,19 +445,19 @@ fn validate_executable_names(
         .syntax()
         .children()
         .filter_map(ast::Definition::cast)
-        // Extension names are allowed to be duplicates,
-        // and schema definitions don't have names.
         .filter(|def| def.is_executable_definition());
 
     for ast_def in executable_definitions {
         let ty_ = match ast_def {
             ast::Definition::OperationDefinition(_) => "operation",
             ast::Definition::FragmentDefinition(_) => "fragment",
+            // Type system definitions are not checked here.
             _ => unreachable!(),
         };
         let scope = match ast_def {
             ast::Definition::OperationDefinition(_) => &mut operation_scope,
             ast::Definition::FragmentDefinition(_) => &mut fragment_scope,
+            // Type system definitions are not checked here.
             _ => unreachable!(),
         };
 
--- a/crates/apollo-compiler/src/validation/validation_db.rs
+++ b/crates/apollo-compiler/src/validation/validation_db.rs
@@ -30,7 +30,15 @@ pub trait ValidationDatabase:
     fn validate_type_system(&self) -> Vec<ApolloDiagnostic>;
 
     /// Validate an executable document.
+    #[salsa::invoke(validate_executable)]
     fn validate_executable(&self, file_id: FileId) -> Vec<ApolloDiagnostic>;
+
+    /// Validate a standalone executable document, without knowledge of the type system it executes
+    /// against.
+    ///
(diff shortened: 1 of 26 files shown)


goto-bus-stop:
Oh, this doesn't validate if fragments exist, eg if you write `{...doesNotExist}`, it doesn't report an error. So that also needs to run in the untyped version of the validations.

SimonSapin (approved):
Looks good!

crates/apollo-compiler/src/validation/validation_db.rs
```diff
@@ -434,10 +524,14 @@ pub fn validate_type_system(db: &dyn ValidationDatabase) -> Vec<ApolloDiagnostic
 
     diagnostics.extend(db.validate_extensions());
 
+    diagnostics.sort_by_key(location_sort_key);
     diagnostics
 }
 
-pub fn validate_executable(db: &dyn ValidationDatabase, file_id: FileId) -> Vec<ApolloDiagnostic> {
+pub fn validate_standalone_executable(
```
lrlna, line 531: I'd also have a wrapping query that iterates over all the executable files and calls this function. That way all executable files can be validated in one go and the user doesn't have to keep all file ids around. I imagine @EverlastingBugstopper  got a couple dozen of files they are trying to validate.

EverlastingBugstopper: ^ Yeah this is a nice to have - was able to implement what I need w/o this but a nicer fn signature wouldn't require passing a file ID.

goto-bus-stop:
Got a bit distracted from this due to the string issues 😇

I'll hopefully complete the other validations that can be done without a type system tomorrow.

goto-bus-stop:
Had to refactor the approach a bit to support validating nested selection sets, to support warning on undefined fragments like `{ a { a { a { a { ...NonExistentFragment } } } }` or undefined variables inside directive applications. Instead of passing down a `Arc<Vec<VariableDefinition>>` i pass a `OperationValidationConfig {}` which includes the available variables and a flag `has_schema` that says whether it should check against the schema.

crates/apollo-compiler/src/validation/validation_db.rs
```diff
@@ -225,19 +238,22 @@ pub trait ValidationDatabase:
...
+        context: OperationValidationConfig,
     ) -> Vec<ApolloDiagnostic>;
 
     #[salsa::invoke(variable::validate_variable_definitions)]
-    fn validate_variable_definitions(&self, defs: Vec<VariableDefinition>)
-        -> Vec<ApolloDiagnostic>;
+    fn validate_variable_definitions(
+        &self,
+        defs: Arc<Vec<VariableDefinition>>,
+        has_schema: bool,
```
SimonSapin, line 255: Should this be `OperationValidationConfig` like in other queries?

SimonSapin (approved):

Issue #608, opened 2023-08-03
Panic with surrogate code points in EscapedUnicode


SimonSapin:
## Description

apollo-parser 0.5.3 provides `impl From<&'_ ast::StringValue> for String` to extract the value represented by a string literal by resolving escape sequences. This part of the spec:

https://spec.graphql.org/October2021/#sec-String-Value.Semantics
> *StringCharacter :: `\u` EscapedUnicode*
> 1. Let *value* be the 16-bit hexadecimal value represented by the sequence of hexadecimal digits within *EscapedUnicode*.
> 2. Return the code point *value*.

Is implemented here:

https://github.com/apollographql/apollo-rs/blob/126816fc6717bf5df429dac48cd71995e5538a7d/crates/apollo-parser/src/ast/node_ext.rs#L169-L176

The spec is written assuming JavaScript-like strings made of 16-bit code units, but Rust’s `char` represents a Unicode scalar value which excludes the range of surrogate code units, U+D800 to U+DFFF. (Surrogates are reserved to leave space for UTF-16 to encode code points beyond U+FFFF as a pair of leading of leading and trailing surrogates.) As a result, this `.unwrap()` call can panic.

The draft spec fixes this (and adds a new bit of syntax):

https://spec.graphql.org/draft/#sec-String-Value

> two forms of Unicode escape sequences: one with a fixed-width of 4 hexadecimal digits (for example \u000A) and one with a variable-width most useful for representing a [supplementary character](https://www.unicode.org/glossary#supplementary_character) such as an Emoji (for example \u{1F4A9}).
> […]
> For legacy reasons, a [supplementary character](https://www.unicode.org/glossary#supplementary_character) may be escaped by two fixed-width unicode escape sequences forming a [surrogate pair](https://www.unicode.org/glossary#surrogate_pair). For example the input "\uD83D\uDCA9" is a valid [StringValue](https://spec.graphql.org/draft/#StringValue) which represents the same [Unicode text](https://spec.graphql.org/draft/#unicode-text) as "\u{1F4A9}". While this legacy form is allowed, it should be avoided as a variable-width unicode escape sequence is a clearer way to encode such code points.
> […]
[...]

PR #633, opened 2023-08-25
fix(parser): fix some escape sequences in strings and indentation in block strings

### LINKED ISSUE #609: Block strings are parsed like non-block strings
## Description

Both in the lexer and in the `impl From<&'_ ast::StringValue> for String` conversion, block strings and "normal" string literals are treated the same except for the number of double-quote delimiters. Block strings should accept different escape sequences (only `\"""`), remove leading/trailing empty lines, and remove common indent.

https://spec.graphql.org/October2021/#sec-String-Value.Semantics

### Steps to reproduce

`crates/apollo-parser/src/ast/node_ext.rs`
```rust
#[cfg(test)]
mod tests {
    use crate::ast;
    use crate::Parser;

    #[test]
    fn test_block_string_1() {
        let input = r#"
            query ($arg: Int!) {
                field(
                    block_string: """
                          other\n\"""

                        \"text\" 
                    """
                )
            }
        "#;
        let values = extract_argument_values(input);
        let ast::Value::StringValue(block_string) = &values[0] else { panic!() };

        // Removed: common indent of non-empty lines, leading/trailing empty lines,
        // but not trailing space of a non-empty line
        // Unescaped: `\"""` only
        assert_eq!(
            String::try_from(block_string).as_deref(),
            Ok(r#"  other\n"""

\"text\" "#)
        );
    }

    #[test]
    fn test_block_string_2() {
        let input = r#"
            query ($arg: Int!) {
                field(
                    block_string: """\""""""
                )
            }
        "#;
        let values = extract_argument_values(input);
        let ast::Value::StringValue(block_string) = &values[0] else { panic!() };
        assert_eq!(String::try_from(block_string).as_deref(), Ok(r#"""""#));
    }

    fn extract_argument_values(input: &str) -> Vec<ast::Value> {
        let ast = Parser::new(input).parse();
        assert_eq!(ast.errors().as_slice(), []);
        let ast::Definition::OperationDefinition(operation) = ast
            .document()
            .definitions()
            .next()
            .unwrap()
        else { panic!("expected an operation") };
        let ast::Selection::Field(field) = operation
            .selection_set()
            .unwrap()
            .selections()
            .next()
            .unwrap()
        else { panic!("expected a field") };
        field
            .arguments()
            .unwrap()
            .arguments()
            .map(|arg| arg.value().unwrap())
            .collect()
    }
}
```

### Expected result

Tests pass

### Actual result

```
---- ast::node_ext::tests::test_block_string_2 stdout ----
thread 'ast::node_ext::tests::test_block_string_2' panicked at 'assertion failed: `(left == right)`
  left: `[ERROR@142:142 "expected R_PAREN, got EOF" EOF, ERROR@142:142 "expected R_CURLY, got EOF" EOF]`,
 right: `[]`', crates/apollo-parser/src/ast/node_ext.rs:333:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- ast::node_ext::tests::test_block_string_1 stdout ----
thread 'ast::node_ext::tests::test_block_string_1' panicked at 'assertion failed: `(left == right)`
  left: `Ok("\n                          other\n\"\"\"\n\n                        \"text\" \n                    ")`,
 right: `Ok("  other\\n\"\"\"\n\n\\\"text\\\" ")`', crates/apollo-parser/src/ast/node_ext.rs:309:9
```

### Environment

- Crate version: current main, 126816fc6717bf5df429dac48cd71995e5538a7d

### LINKED ISSUE #611: Incorrect parsing of string with trailing \"
## Description

The conversion of `StringValue` to `String` is incorrect when the value ends with an escaped double-quote `\"`. There is likely a similar bug with block strings that end with `\"""` or starts with `"` but it is masked by https://github.com/apollographql/apollo-rs/issues/609. The offending code removes any number of `"` characters here:

https://github.com/apollographql/apollo-rs/blob/126816fc6717bf5df429dac48cd71995e5538a7d/crates/apollo-parser/src/ast/node_ext.rs#L201

### Steps to reproduce

`crates/apollo-parser/src/ast/node_ext.rs`
```rust
#[cfg(test)]
mod tests {
    use crate::ast;
    use crate::Parser;

    #[test]
    fn test_tailing_escaped_quote() {
        let input = r#"
            query ($arg: Int!) {
                field(
                    string: "a\""
                )
            }
        "#;
        let values = extract_argument_values(input);
        let ast::Value::StringValue(block_string) = &values[0] else { panic!() };
        assert_eq!(String::try_from(block_string).as_deref(), Ok(r#"a""#));
    }

    fn extract_argument_values(input: &str) -> Vec<ast::Value> {
        let ast = Parser::new(input).parse();
        assert_eq!(ast.errors().as_slice(), []);
        let ast::Definition::OperationDefinition(operation) = ast
            .document()
            .definitions()
            .next()
            .unwrap()
        else { panic!("expected an operation") };
        let ast::Selection::Field(field) = operation
            .selection_set()
            .unwrap()
            .selections()
            .next()
            .unwrap()
        else { panic!("expected a field") };
        field
            .arguments()
            .unwrap()
            .arguments()
            .map(|arg| arg.value().unwrap())
            .collect()
    }
}
```

### Expected result

Test passes

### Actual result

```rust
thread 'ast::node_ext::tests::test_tailing_escaped_quote' panicked at 'assertion failed: `(left == right)`
  left: `Ok("a\\")`,
 right: `Ok("a\"")`', crates/apollo-parser/src/ast/node_ext.rs:301:9
```

### Environment

- Crate version: either 0.5.3 or `main` 126816fc6717bf5df429dac48cd71995e5538a7d

goto-bus-stop:
Follow-up to #638, fixing conversion of StringValue nodes to Rust strings.

Fixes #609
Fixes #611

 This handles
- [x] escape sequences in single-quoted strings and
- [x] indentation in block strings #609
- [x] wrong behaviours when a string ends with an escaped `"` #611 

This does not handle these unicode escape issues:
- https://github.com/apollographql/apollo-rs/issues/608
- https://github.com/apollographql/apollo-rs/issues/640

### DIFF
--- a/crates/apollo-parser/src/ast/node_ext.rs
+++ b/crates/apollo-parser/src/ast/node_ext.rs
@@ -438,11 +438,40 @@ fn text_of_first_token(node: &SyntaxNode) -> TokenText {
 }
 
 #[cfg(test)]
+mod string_tests {
+    use super::unescape_string;
+
+    #[test]
+    fn it_parses_strings() {
+        assert_eq!(unescape_string(r"simple"), "simple");
+        assert_eq!(unescape_string(r" white space "), " white space ");
+    }
+
+    #[test]
+    fn it_unescapes_strings() {
+        assert_eq!(unescape_string(r#"quote \""#), "quote \"");
+        assert_eq!(
+            unescape_string(r"escaped \n\r\b\t\f"),
+            "escaped \n\r\u{0008}\t\u{000c}"
+        );
+        assert_eq!(unescape_string(r"slashes \\ \/"), r"slashes \ /");
+        assert_eq!(
+            unescape_string("unescaped unicode outside BMP \u{1f600}"),
+            "unescaped unicode outside BMP \u{1f600}"
+        );
+        assert_eq!(
+            unescape_string(r"unicode \u1234\u5678\u90AB\uCDEF"),
+            "unicode \u{1234}\u{5678}\u{90AB}\u{CDEF}"
+        );
+    }
+}
+
+#[cfg(test)]
 mod block_string_tests {
     use super::{split_lines, unescape_block_string};
 
     #[test]
-    fn test_split_graphql_lines() {
+    fn it_splits_lines_by_graphql_newline_definition() {
         let plain_newlines: Vec<_> = split_lines(
             r#"source text
     with some
--- a/crates/apollo-parser/src/ast/node_ext.rs
+++ b/crates/apollo-parser/src/ast/node_ext.rs
@@ -281,6 +281,8 @@ fn unescape_block_string(raw_value: &str) -> String {
         line.chars().take_while(|&c| is_whitespace(c)).count()
     }
 
+    // 1. Let lines be the result of splitting rawValue by LineTerminator.
+    // 2. Let commonIndent be null.
     // 3. For each line in lines:
     let common_indent = split_lines(raw_value)
         // 3.a. If line is the first item in lines, continue to the next line.
@@ -312,6 +314,8 @@ fn unescape_block_string(raw_value: &str) -> String {
         // 5.a. Remove the first item from lines.
         .skip_while(|line| is_whitespace_line(line));
 
+    // (Step 6 is done at the end so we don't need an intermediate allocation.)
+
     // 7. Let formatted be the empty character sequence.
     let mut formatted = String::with_capacity(raw_value.len());
 
@@ -331,12 +335,13 @@ fn unescape_block_string(raw_value: &str) -> String {
         // 8.b.ii. Append formatted with line.
         replace_into(line, ESCAPED_TRIPLE_QUOTE, TRIPLE_QUOTE, &mut formatted);
 
+        // Track the last non-whitespace line for implementing step 6 in the spec.
         if !is_whitespace_line(line) {
             final_char_index = formatted.len();
         }
     }
 
-    // Remove WhiteSpace-only lines from the end.
+    // 6. Implemented differently: remove WhiteSpace-only lines from the end.
     formatted.truncate(final_char_index);
 
     // 9. Return formatted.
--- a/crates/apollo-parser/src/ast/node_ext.rs
+++ b/crates/apollo-parser/src/ast/node_ext.rs
@@ -491,13 +491,10 @@ mod block_string_tests {
     #[test]
     fn it_does_not_unescape_block_strings() {
         assert_eq!(
-            unescape_block_string(r#"escaped \n\r\b\t\f"#),
-            r#"escaped \n\r\b\t\f"#
+            unescape_block_string(r"escaped \n\r\b\t\f"),
+            r"escaped \n\r\b\t\f"
         );
-        assert_eq!(
-            unescape_block_string(r#"slashes \\ \/"#),
-            r#"slashes \\ \/"#
-        );
+        assert_eq!(unescape_block_string(r"slashes \\ \/"), r"slashes \\ \/");
         assert_eq!(
             unescape_block_string("unescaped unicode outside BMP \u{1f600}"),
             "unescaped unicode outside BMP \u{1f600}"
@@ -513,30 +510,30 @@ mod block_string_tests {
 
         assert_eq!(
             unescape_block_string(
-                r#"
+                r"
             This is
             indented
             quite a lot
-    "#
+    "
             ),
-            r#"This is
+            r"This is
 indented
-quite a lot"#
+quite a lot"
         );
 
         assert_eq!(
             unescape_block_string(
-                r#"
+                r"
 
         spans
           multiple
             lines
 
-    "#
+    "
             ),
-            r#"spans
+            r"spans
   multiple
-    lines"#
+    lines"
         );
     }
 }
--- a/crates/apollo-parser/src/ast/node_ext.rs
+++ b/crates/apollo-parser/src/ast/node_ext.rs
@@ -456,8 +456,8 @@ mod string_tests {
         );
         assert_eq!(unescape_string(r"slashes \\ \/"), r"slashes \ /");
         assert_eq!(
-            unescape_string("unescaped unicode outside BMP \u{1f600}"),
-            "unescaped unicode outside BMP \u{1f600}"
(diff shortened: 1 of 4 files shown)


crates/apollo-parser/src/ast/node_ext.rs
```diff
@@ -193,12 +196,164 @@ fn unescape_string(input: &str) -> String {
...
+        let Some(index) = memchr::memchr2(b'\r', b'\n', self.input.as_bytes()) else {
+            self.finished = true;
+            return Some(self.input);
+        };
+        let line = &self.input[..index];
+        let rest = match self.input.get(index..=index + 1) {
+            Some("\r\n") => &self.input[index + 2..],
+            _ => &self.input[index + 1..],
+        };
+        self.input = rest;
```
SimonSapin, line 240: The above is fine. It can be marginally nicer with `split_at` and `strip_prefix`:
```suggestion
        let (line, rest) = self.input.split_at(index);
        // `rest` starts with \r or \n
        self.input = rest.strip_prefix("\r\n").unwrap_or_else(|| &rest[1..])
```

crates/apollo-parser/src/ast/node_ext.rs
```diff
@@ -193,12 +196,164 @@ fn unescape_string(input: &str) -> String {
...
+        line.chars().take_while(|&c| is_whitespace(c)).count()
+    }
+
+    // 3. For each line in lines:
+    let common_indent = split_lines(raw_value)
+        // 3.a. If line is the first item in lines, continue to the next line.
+        .skip(1)
+        .filter_map(|line| {
+            // 3.b. Let length be the number of characters in line.
+            let length = line.len();
```
SimonSapin, line 290: Please add a comment to explain that while `str.len()` counts UTF-8 bytes and `count_indent` counts `char`s, comparing them is still ok because the WhiteSpace characters we’re looking for are in the ASCII range and so occupy one UTF-8 byte each.

crates/apollo-parser/src/ast/node_ext.rs
```diff
@@ -193,12 +196,164 @@ fn unescape_string(input: &str) -> String {
...
+                // 4.a.ii. Remove commonIndent characters from the beginning of line.
+                &line[common_indent.min(line.len())..]
+            }
+        })
+        // 5. While the first item line in lines contains only WhiteSpace:
+        // 5.a. Remove the first item from lines.
+        .skip_while(|line| is_whitespace_line(line));
+
+    // 7. Let formatted be the empty character sequence.
+    let mut formatted = String::with_capacity(raw_value.len());
```
SimonSapin, line 316: Please add a comment to say step 6 is handled below

crates/apollo-parser/src/ast/node_ext.rs
```diff
@@ -281,3 +436,133 @@ fn text_of_first_token(node: &SyntaxNode) -> TokenText {
...
+        assert_eq!(unescape_string(r#"quote \""#), "quote \"");
+        assert_eq!(
+            unescape_string(r"escaped \n\r\b\t\f"),
+            "escaped \n\r\u{0008}\t\u{000c}"
+        );
+        assert_eq!(unescape_string(r"slashes \\ \/"), r"slashes \ /");
+        assert_eq!(
+            unescape_string("unescaped unicode outside BMP \u{1f600}"),
+            "unescaped unicode outside BMP \u{1f600}"
+        );
```
SimonSapin, line 461: This test case is misleading. It looks like it’s testing for https://github.com/apollographql/apollo-rs/issues/640 but that’s not implemented yet. The test passes because the Rust input string is not a raw string so `\u{1f600}` is unescaped by the Rust parser and `unescape_string` doesn’t end up unescaping anything in this case.

crates/apollo-parser/src/ast/node_ext.rs
```diff
@@ -281,3 +436,133 @@ fn text_of_first_token(node: &SyntaxNode) -> TokenText {
...
+    fn it_does_not_unescape_block_strings() {
+        assert_eq!(
+            unescape_block_string(r"escaped \n\r\b\t\f"),
+            r"escaped \n\r\b\t\f"
+        );
+        assert_eq!(unescape_block_string(r"slashes \\ \/"), r"slashes \\ \/");
+        assert_eq!(
+            unescape_block_string("unescaped unicode outside BMP \u{1f600}"),
+            "unescaped unicode outside BMP \u{1f600}"
+        );
```
SimonSapin, line 530: Same as previous comment

SimonSapin (approved):
Looks great!

crates/apollo-parser/src/parser/grammar/value.rs
```diff
@@ -186,6 +186,48 @@ enum Test @dir__one(string: "string value", int_value: -10, float_value: -1.123e
         }
     }
 
+    #[test]
```
lrlna, line 189: Should we add these to the test_data dir? or are these better addressed here?

lrlna (approved):
\o/

crates/apollo-parser/src/ast/node_ext.rs
goto-bus-stop: Right, these are meant to test literal non-escaped unicode characters in the string, I'll add a comment to clarify.

crates/apollo-parser/src/parser/grammar/value.rs
goto-bus-stop: `test_data` only checks the AST/CST resulting from parsing, it doesn't check the conversion from CST StringValue to String.

crates/apollo-parser/src/ast/node_ext.rs
SimonSapin: Perhaps use non-escaped rust syntax? Like 그라프큐엘 in the expected result of another test

crates/apollo-parser/src/ast/node_ext.rs
goto-bus-stop: my personal feeling is that `split_at` looks nicer but the manual `match` is more obviously correct despite being a bit noisy... though as i look back and forth i don't feel strongly anymore.

PR #634, opened 2023-08-25
breaking(all): rename `ast` to `cst`

SimonSapin:
The Rowan-based typed syntax tree emitted by the parser used to be called Abstract Syntax Tree (AST) but is in fact not very abstract: it preserves text input losslessly, and all tree leaves are string-based tokens. This renames it to Concrete Syntax Tree (CST) and renames various APIs accordingly.

This leaves the AST name available for a new tree representation (coming soon).


SimonSapin:
On second thought it will make more sense to have this rename in the same PR that adds the new AST, because then apollo-compiler may not need to expose the CST at all.

PR #635, opened 2023-08-25
chore(compiler): add directive extraction example

### LINKED ISSUE #636: added a silly directive extraction example
## Description

Added a silly example showing how one can use the `apollo_compiler` crate to extract directives used by a given query. I had to do something similar for my day job and figured I'd share the love :). Opened https://github.com/apollographql/apollo-rs/pull/635 for this

maxnachlinger:
Added a silly example showing how one can use the apollo_compiler crate to extract directives used by a given query. I had to do something similar for my day job and figured I'd share the love :)  Fixes #636

### DIFF
--- a/crates/apollo-compiler/examples/extract_directives_used_by_query.rs
+++ b/crates/apollo-compiler/examples/extract_directives_used_by_query.rs
@@ -1,3 +1,4 @@
+//! This example collects all directives declared on the fields that are queried by an operation.
 use apollo_compiler::{
     hir::{Directive, Field},
     ApolloCompiler, FileId, HirDatabase,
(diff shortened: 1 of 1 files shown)


maxnachlinger:
Fixes #636

crates/apollo-compiler/examples/extract_directives_used_by_query.rs
```diff
@@ -0,0 +1,176 @@
...
+        "#;
+        let query_id = compiler.add_executable(&query_src, "not-used-here.graphql");
+
+        let directives = get_directives_used_in_query(&compiler, &query_id);
+
+        assert_eq!(directives.len(), 4);
+    }
+
+    #[test]
+    fn it_only_directives_used_by_query() {
```
maxnachlinger, line 143:
```suggestion
    fn it_only_returns_directives_used_by_query() {
```

crates/apollo-compiler/examples/extract_directives_used_by_query.rs
```diff
@@ -0,0 +1,176 @@
...
+              }
+            }
+          }
+        }
+        "#;
+    let query_id = compiler.add_executable(query_src, "not-used-here.graphql");
+
+    let directives = get_directives_used_in_query(&compiler, &query_id);
+
+    assert_eq!(directives.len(), 4);
```
maxnachlinger, line 67: There's room for improvement here of course, `directives` has duplicates present, I couldn't find a simple way to get a hash for uniqueness, the directive location bits differ amongst otherwise duplicate directives. Solving this seems a bit out of scope for a simple example though :)

crates/apollo-compiler/examples/extract_directives_used_by_query.rs
```diff
@@ -0,0 +1,101 @@
...
+            .fold(vec![], |mut acc, operation_definition| {
+                acc.extend(operation_definition.selection_set().fields().to_vec());
+                acc
+            });
+
+    let mut directives = vec![];
+
+    // depth first search for nested fields with directives
+    while !stack.is_empty() {
+        if let Some(field) = stack.pop() {
```
goto-bus-stop, line 22: Here `stack.is_empty()` and `stack.pop()` are basically doing the same check, so it can be simplified to
```rust
while let Some(field) = stack.pop() {
```

I'll apply that change before merging!

goto-bus-stop (approved):
Great thanks! This is a pretty common use case so really nice to have an example for it.

Issue #632, opened 2023-08-25
parser: 0.6.0 fails to lex block string comment


lrlna:
The following lexed properly in `apollo-parser@0.5.3`:

```
"""
'/\W/'
"""
type Query {
  a: String
}
```

But now, on `apollo-parser@0.6.0` produces an error:

```
errors: ERROR@173:187 "unexpected escaped character" """
'/\W/'
"""
```
related apollographql/router#3680

SimonSapin:
I think this is another case https://github.com/apollographql/apollo-rs/issues/609

PR #638, opened 2023-08-28
fix(parser): fix lexing escape sequences in block strings

goto-bus-stop:
This is the most basic fix for #632, but there are other parse bugs left that may be we should address with a bunch more tests. This fixes a regression in apollo-parser 0.6.0.

Block strings are not meant to support escape sequences in general. But the lexer handled strings and block strings the same way, so an invalid escape sequence in a block string 
would cause a lex error. This rips out all the escape sequence handling from block strings except for `\"""`.

I'm running it through the corpus of Apollo schemas right now, about 75% done without problems, so it looks good.

### DIFF
--- a/crates/apollo-parser/src/lexer/mod.rs
+++ b/crates/apollo-parser/src/lexer/mod.rs
@@ -40,7 +40,6 @@ enum State {
     StringLiteralEscapedUnicode(usize),
     StringLiteral,
     StringLiteralStart,
-    BlockStringLiteralEscapedUnicode(usize),
     BlockStringLiteral,
     BlockStringLiteralBackslash,
     StringLiteralBackslash,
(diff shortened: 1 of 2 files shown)


crates/apollo-parser/src/lexer/mod.rs
```diff
@@ -379,19 +351,17 @@ impl<'a> Cursor<'a> {
...
                 State::BlockStringLiteralBackslash => match c {
                     '"' => {
-                        while self.eatc('"') {}
+                        // If this is \""", we need to eat 3 in total, and then continue parsing.
+                        // The lexer does not un-escape escape sequences so it's OK
+                        // if we take this path for \"", even if that is technically not an escape
+                        // sequence.
+                        if self.eatc('"') {
+                            self.eatc('"');
+                        }
```
goto-bus-stop, line 360: There's a test below to make sure `"""\""""""` is correctly lexed as a block string containing only `\"""`.

SimonSapin (approved):

PR #642, opened 2023-09-08
fix(compiler): validate input value types

goto-bus-stop:
This is a pretty big thing that we missed until now, so I added it to the current validation implementation. Obviously will
also address this in the validation on top of the new AST. Found by @o0Ignition0o \ 😄 /

Input values (in `input object` declarations and in argument declarations) did not have full validation. Only their names were checked, but if you declared a `arg: NonExistentType`, it would happily truck on.

This PR checks that the declared type exists, and is an input type. If you declare an input value or argument with an output type, you get an error.

As a small cleanliness refactor, this moves the object field output type validation into `validate_field_definition`, as it doesn't need context about other fields in a type definition.

The input type validations for variables and arguments/input objects now use different diagnostics so the messages make more sense.

### DIFF
--- a/crates/apollo-compiler/README.md
+++ b/crates/apollo-compiler/README.md
@@ -173,6 +173,11 @@ fn main() -> Result<()> {
       weight: Int
     }
 
+    enum join__Graph {
+      INVENTORY,
+      PRODUCTS,
+    }
+    scalar join__FieldSet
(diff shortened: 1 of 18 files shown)


lrlna (approved):
wow such a good catch!! thanks @goto-bus-stop @o0Ignition0o

Issue #641, opened 2023-09-01
apollo-compiler@0.12.0 checklist


lrlna:
- [ ] validation on top of current data structure
- [ ] set_type_system_hir
- [ ] remove `hir` module
- [ ] adding to the test suite with old `hir` to test the new `hir`
- [ ] port the router to the new APIs
- [ ] run on the schema corpus
- [ ] docs for the new architecture
- [ ] docs for new API usage (rust inline docs + readme)
- [ ] more examples:
  - [ ] operation manipulations
  - [ ] synthetic/logic based schema creation (only merge specific fields in an object type)
  - [ ] create a schema and use that as set_type_system_hir to start a compiler

SimonSapin:
> set_type_system_hir

Similarly, users may want to validate a programatically-created `ExecutableDocument` without serializing and re-parsing it. So we’d need a naming scheme for all four combinations of adding input to a compiler, on two axes:

* A string to be parsed v.s. a high-level data structure
* Schema v.s. executable

(Plus one for a mixed schema-and-executable string to be parsed, though there’s no data structure equivalent)

PR #645, opened 2023-09-18
WIP: New crate implementing execution and schema introspection

SimonSapin:
**Note:** this depends on ongoing work for https://github.com/apollographql/apollo-rs/issues/641, so for now this PR targets a temporary branch in order to only show relevant diffs.

---

This is an initial draft implementation of (trying to be) spec-compliant:

* [Execution](https://spec.graphql.org/October2021/#sec-Execution):
  - A trait abstraction for objects with asynchronously resolvable fields
  - Input coercion / type checking of variable values and argument values
  - Field collection (handling `@skip`, `@include`, and type conditions)
  - Handling of field error (collecting the relevant path, propagating `null`), raised by a resolver or otherwise
  - Result coercion / type checking of response data
  - `serde`-serializable data structures of a GraphQL response, request errors, field errors, and field error path
  - JSON data for variable values and response data is represented with `serde_json_bytes::Value` (which uses a wrapper of reference-counted [`Bytes`](https://crates.io/crates/bytes) for strings)

* [Schema introspection](https://spec.graphql.org/October2021/#sec-Schema-Introspection):
  - Removing and collecting schema introspection selections (`__schema` and `__type`) from an executable document, so the rest can be executed separately (whether with the above, or not like in Apollo Router)
  - Execution of those collected selections, returning a response that can be merged with the response for the rest of the request

See `crates/apollo-introspection/tests/doc_example.rs` for an API usage example.

Testing is minimal. Experience integrating with Apollo Router would be good to have before merging this PR. Execution is included in support of introspection and not exposed in the public API. If/when another use case comes we can expose it, with some more thought on the API design (especially for the resolver trait).

### DIFF
--- a/crates/apollo-introspection/tests/doc_example.rs
+++ b/crates/apollo-introspection/tests/doc_example.rs
@@ -1,0 +1,35 @@
+use apollo_compiler::ExecutableDocument;
+use apollo_compiler::Schema;
+use apollo_introspection::get_operation;
+use apollo_introspection::JsonMap;
+use apollo_introspection::RequestErrorResponse;
+use apollo_introspection::Response;
+use apollo_introspection::SchemaIntrospectionQuery;
+use apollo_introspection::VariableValues;
+
+/// `schema` and `document` are presumed valid
+pub fn execute_request(
+    schema: &Schema,
+    mut document: ExecutableDocument,
+    operation_name: Option<&str>,
+    variable_values: &JsonMap,
+) -> Result<Response, RequestErrorResponse> {
+    let introspection = SchemaIntrospectionQuery::split_from(&mut document, operation_name)?;
+    let operation = get_operation(&document, operation_name)?
+        .definition()
+        .clone();
+    let coerced_variable_values = VariableValues::coerce(schema, &operation, variable_values)?;
+    let response =
+        execute_non_introspection(schema, &document, operation_name, &coerced_variable_values)?;
+    let intropsection_response = introspection.execute_sync(schema, &coerced_variable_values)?;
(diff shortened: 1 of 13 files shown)


crates/apollo-introspection/tests/doc_example.rs
```diff
@@ -0,0 +1,35 @@
...
+) -> Result<Response, RequestErrorResponse> {
+    let introspection = SchemaIntrospectionQuery::split_from(&mut document, operation_name)?;
+    let operation = get_operation(&document, operation_name)?
+        .definition()
+        .clone();
+    let coerced_variable_values = VariableValues::coerce(schema, &operation, variable_values)?;
+    let response =
+        execute_non_introspection(schema, &document, operation_name, &coerced_variable_values)?;
+    let intropsection_response = introspection.execute_sync(schema, &coerced_variable_values)?;
+    Ok(response.merge(intropsection_response))
```
yanns, line 25:
```suggestion
    let introspection_response = introspection.execute_sync(schema, &coerced_variable_values)?;
    Ok(response.merge(introspection_response))
```

SimonSapin: Maybe some day we’ll name something interoperability section, but not today.

Good catch, thanks!

SimonSapin:
> Input coercion / type checking of variable values and argument values

Apollo Router has roughly the same functionality in a function called `validate_input_value`. As part of introspection work, let’s have apollo-rs expose input coercion in a way that can be used by the Router instead of `validate_input_value`.

SimonSapin:
Moved to https://github.com/apollographql/apollo-rs/pull/758

PR #646, opened 2023-09-19
Fix integer lexing

SimonSapin:
* A plus sign is an error
* A minus sign followed by EOF without any digit is an error

Found by fuzzing! (New fuzz target upcoming in another PR)


goto-bus-stop (approved):
great!

PR #648, opened 2023-09-20
chore(fuzz): differential fuzzing for validation

lrlna:
This works of a branch in `routet-bridge` that calls js validation. This should be a more accurate representation of what happens in the router, making this more of Gala vs. Golden Delicious apple comparison.


fuzz/fuzz_targets/validation.rs
```diff
@@ -0,0 +1,69 @@
...
+    compiler.add_document(&ts, "ts.graphql");
+    let rust_diagnostics = compiler.validate();
+
+    debug!("======= OPERATION =======");
+    debug!("{}", op);
+    debug!("========================");
+    debug!("{}", ts);
+    debug!("========================");
+    let js_diagnostics =
+        router_bridge::validate::validate(&ts, &op).expect("js validation should succeed");
```
goto-bus-stop, line 30:
```suggestion
        router_bridge::validate::validate(&ts, &op).expect("bridge should return validation result");
```
would emphasise here that it's not that JS validation did not pass, but that something else went wrong with Deno or somewhere, not the actual GraphQL validation

fuzz/fuzz_targets/validation.rs
```diff
@@ -0,0 +1,69 @@
...
+    let (op, ts) = match generate_valid_operation(data) {
+        Ok(d) => (d.0, d.1),
+        Err(_err) => {
+            return;
+        }
+    };
+
+    let mut compiler = ApolloCompiler::new();
+    compiler.add_executable(&op, "op.graphql");
+    compiler.add_document(&ts, "ts.graphql");
```
goto-bus-stop, line 21:
```suggestion
    compiler.add_type_system(&ts, "ts.graphql");
```
just to match actual usage, the behaviour of `document` is a bit different

Cargo.toml
```diff
@@ -8,3 +8,11 @@ members = [
     "crates/apollo-smith",
     "fuzz",
 ]
+
+default-members = [
```
lrlna, line 12: temporarily removing fuzz from workspace members for CI. not a good idea to merge it like this though.

lrlna: (router-bridge is extremely fussy and is refusing to build in CI due to too many failures)

lrlna:
Closing this. Moved any deno-related validation to [its own repo](https://github.com/apollographql/apollo-rs-fuzz) so we don't have to build it as part of apollo-rs.

PR #650, opened 2023-09-22
Introduce ast::IntValue and FloatValue containing a String

SimonSapin:
Rather converting to i32 or f64 during parsing, we preserve the parsed token as-is. Conversion is left for input value coercion, when the expected type is known (`Float`, `Int`, `ID`, custom scalar, …)


goto-bus-stop (approved):
I think this is a great solution 😸

PR #652, opened 2023-09-24
Fix various lexer bug

SimonSapin:
As fuzzing eventually showed, https://github.com/apollographql/apollo-rs/pull/646 turned out to be only a couple cases of a deeper problem. When reaching EOF, the state-machine-based lexer would handle a few cases specifically and have a catch-all pattern to happily return whatever partial token was being built in other cases.

The commits of this PR: 

* Change the snapshot test harness to put errors where they are emitted instead of at the end
* Refactor the lexer to make flow control a bit more visible, especially in the EOF case, and replace the offending catch-all pattern with explicit ones.
* Add snapshot test cases that show incorrect lexing
* Fix various bugs and update test expectations


crates/apollo-parser/src/lexer/mod.rs
```diff
@@ -528,40 +578,58 @@ impl<'a> Cursor<'a> {
...
-
-        // Bidi markers
-        | '\u{200E}' // LEFT-TO-RIGHT MARK
-        | '\u{200F}' // RIGHT-TO-LEFT MARK
-
-        // Dedicated whitespace characters from Unicode
-        | '\u{2028}' // LINE SEPARATOR
-        | '\u{2029}' // PARAGRAPH SEPARATOR
+        // https://spec.graphql.org/October2021/#UnicodeBOM
+        | '\u{FEFF}' // Unicode BOM (Byte Order Mark)
```
goto-bus-stop, line 619: yess

PR #655, opened 2023-09-25
Remove old HIR

SimonSapin:
* Port tests and examples to use the new HIR (`Schema` and `ExecutableDocument`) instead of the CST or the old HIR
* Remove the old HIR
* Remove access to the CST. The new AST should cover those use cases. apollo-parser is now a private dependency of apollo-compiler, so a breaking API change in the former is not necessarily a breaking change for the latter anymore.


crates/apollo-compiler/src/database/hir_db.rs
lrlna, line 1: How are we keeping track that all of the `hir_db` (and `hir` getters for that matter) methods remain in the rewrite?

crates/apollo-compiler/src/lib.rs
```diff
@@ -187,7 +168,7 @@ impl ApolloCompiler {
     ///
     /// Returns a `FileId` that you can use to update the source text of this document.
     pub fn add_type_system(&mut self, input: &str, path: impl AsRef<Path>) -> FileId {
-        if self.db.type_system_hir_input().is_some() || self.db.schema_input().is_some() {
+        if self.db.schema_input().is_some() {
```
lrlna, line 171: to double check my understanding, we are currently calling:
- `schema` - a hir-based (and programmatic also?) type system representation
- `type_system` - a string-based type system representation

crates/apollo-compiler/src/validation/scalar.rs
```diff
@@ -27,7 +27,7 @@ pub fn validate_scalar_definition(
     if !scalar_def.is_built_in() {
         // Custom scalars must provide a scalar specification URL via the
         // @specifiedBy directive
-        let has_specified_by = scalar_def.directives.get("specifiedBy").is_some();
+        let has_specified_by = scalar_def.directives.has("specifiedBy");
```
lrlna, line 30: small but mighty api change >.<

crates/apollo-compiler/tests/misc.rs
```diff
@@ -0,0 +1,1037 @@
...
+  price
+  dimensions
+  size
+  weight
+}
+      "#;
+
+    let mut compiler = ApolloCompiler::new();
+    compiler.add_document(schema, "schema.graphql");
+    compiler.add_executable(query, "query.graphql");
```
lrlna, line 30: oh we never actually asserted anything in this test? oops

crates/apollo-encoder/src/from_hir.rs
lrlna, line 1: gotta keep track the encoder is going to need a major bump with this release too

lrlna (approved):
I don't have as much context as @goto-bus-stop, but nothing immediately jumps out at me. I am liking how we are simplifying the overall API though!

crates/apollo-compiler/src/database/hir_db.rs
SimonSapin: Hmm there isn’t a formal way to do this. We can look at `cargo doc` outputs side by side, or find out when porting the Router that some helper function is missing and add it back as needed.

crates/apollo-compiler/src/lib.rs
SimonSapin: It’s not 100% complete with this PR, but the plan is to have v1.0 no longer use `type_system` terminology at all.

`Schema` and `ExecutableDocument` are the two main types. Together we can informally call them "HIR" but that’s also not in public APIs.

crates/apollo-encoder/src/from_hir.rs
lrlna: We are then going to need to have a path forward for encoding json to sdl, as [rover currently uses the encoder](https://github.com/apollographql/rover/blob/0d242064c09e78f6ac340fd1762bc377d4230a2a/crates/rover-client/src/operations/graph/introspect/schema.rs) to create sdl for introspection.

goto-bus-stop: Rover would use the new high level types to create documents out of introspection results I imagine

SimonSapin: Right, either that or AST

This rover code is already doing a conversion from serde-deserializable types that match the structure of JSON introspection response to apollo-encoder types. This would change the target type of that conversion.

Or even, this functionality could be part of apollo-compiler or apollo-introspection, wherever we end up exposing https://github.com/apollographql/apollo-rs/pull/645

lrlna: ok! i made an issue to update rover with whichever path we chose here.

PR #660, opened 2023-09-26
chore(compiler): use implementers_map in GetPossibleTypes

goto-bus-stop:
Replaces the linear scan of the entire schema. This code runs for every fragment spread so it can be medium-hot on pathological operations.

GetPossibleTypes only operates on concrete object types. So for interface implementers, other interfaces do not count, only object types. This means that we do have to remove interfaces from the result of `implementers_map().get()`. In theory it still results in the same amount of allocations and far fewer hashing operations than before.

While the GraphQL spec thinks of implementers as only object types, other use cases may also want to check interfaces. Eg. authorization in the router uses the implementers map to check whether any implementation of an interface requires authorization, and you can in fact apply auth directives to a child interface. This PR uses a `struct Implementers {objects, interfaces}` so you can access objects and child interfaces separately or at once depending on your use case.


goto-bus-stop:
Changed it to option 3, with an `Implementers` type that has separate `objects` and `interfaces` sets and is also iterable itself if you don't care.

crates/apollo-compiler/src/schema/mod.rs
```diff
@@ -831,6 +871,36 @@ impl From<Node<InputObjectType>> for ExtendedType {
...
 
+impl Implementers {
+    pub fn iter(&self) -> impl Iterator<Item = &'_ Name> {
+        self.objects.iter().chain(&self.interfaces)
+    }
+}
+
+impl IntoIterator for Implementers {
+    type Item = Name;
+    type IntoIter = std::iter::Chain<
```
goto-bus-stop, line 882: I think I have to write out the whole type here, right? There's no such thing as `type IntoIter = impl Iterator`

SimonSapin: Right, or at least there isn’t yet.
Maybe the `iter` method is enough?

SimonSapin:
I didn’t mean to close this, sorry! I forgot to change this PR’s target branch from `mut` before I renamed `mut` to `compiler-v1.0`. Should be fixed now. (well, except for conflicts)

goto-bus-stop:
Heh, i probably should've merged this as soon as i pushed the iter() change!

goto-bus-stop:
To rebase Monday

goto-bus-stop:
ah great! 😁

PR #661, opened 2023-09-26
chore(compiler): rewrite RecursionStack to avoid copies to String

goto-bus-stop:
Now that this is only used with the new `NodeStr` type, it's wasteful and inconvenient to have it use String instances.

Also split it into `RecursionStack`/`RecursionGuard` types so you can't write something that has lifetime errors as easily, and so you don't have to create your own vec for storage.

### DIFF
--- a/crates/apollo-compiler/src/validation/mod.rs
+++ b/crates/apollo-compiler/src/validation/mod.rs
@@ -17,27 +17,56 @@ mod union_;
 mod value;
 mod variable;
 
+use crate::NodeStr;
 pub use validation_db::{ValidationDatabase, ValidationStorage};
 
 /// Track used names in a recursive function.
-///
-/// Pass the result of `stack.push(name)` to recursive calls. Use `stack.contains(name)` to check
-/// if the name was used somewhere up the call stack.
-struct RecursionStack<'a>(&'a mut Vec<String>);
-impl RecursionStack<'_> {
-    fn push(&mut self, name: String) -> RecursionStack<'_> {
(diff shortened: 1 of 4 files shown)


crates/apollo-compiler/src/validation/mod.rs
```diff
@@ -17,27 +17,56 @@ mod union_;
...
+struct RecursionGuard<'a>(&'a mut Vec<NodeStr>);
+impl RecursionGuard<'_> {
+    /// Mark that we saw a name.
+    fn push(&mut self, name: &NodeStr) -> RecursionGuard<'_> {
+        self.0.push(name.clone());
+        RecursionGuard(self.0)
     }
+    /// Check if we saw a name somewhere up the call stack.
     fn contains(&self, name: &str) -> bool {
         self.0.iter().any(|seen| seen == name)
```
SimonSapin, line 59: Could the `Vec` be replaced with `IndexSet` to make this O(1)?

SimonSapin (approved):

crates/apollo-compiler/src/validation/mod.rs
goto-bus-stop: Huh, when I try that the structure seems to not work anymore, which is a bit suspicious. Looking into why that might be...

goto-bus-stop: It was suspicious because the directives recursion implementation was registering directive definitions twice!

SimonSapin (approved):

PR #662, opened 2023-09-26
fix(parser): apply recursion limit everywhere, reduce default to 500

SimonSapin:
Sorry for the noisy diff!

This adds a new test in `crates/apollo-parser/src/parser/mod.rs` that would cause stack overflows without the new uses of limits. See also comments in that same file about the change of default limit.


goto-bus-stop (approved):
sweet!

PR #663, opened 2023-09-26
Introduce a pair of `Directives` struct

SimonSapin:
One for the `ast` and `executable` modules, one for the `schema` module. (Containing `Node` v.s. `Component`)

* `x.directive_by_name(n)` becomes `x.directives.get(n)`
* `x.directives_by_name(n)` becomes `x.directives.get_all(n)`
* `x.directives` dereferences to `Vec`

Code that operates on directives of various things can now accept `&Directives` and still use those `get` or `get_all` methods


goto-bus-stop (approved):
I like it :smile:

PR #665, opened 2023-09-27
Ignore component origin in == and hash, turn some maps into sets

SimonSapin:
`Component<T>` and `ComponentStr` now compare equal based on their content, without considering anymore whether they come from a definition or an extension.

Because `ExtensionId` is an arbitrary integer, parsing the same SDL twice yields different IDs. This would cause the two `Schema` structs not to compare equal, which made it impossible to use the invariant "serializing and reparsing gives the same tree" for fuzzing.

Additionally, instances of `IndexMap<NodeStr, ComponentOrigin>` have been simplified to the now-equivalent `IndexSet<ComponentStr>`.

### DIFF
--- a/crates/apollo-compiler/src/schema/component.rs
+++ b/crates/apollo-compiler/src/schema/component.rs
@@ -1,6 +1,7 @@
 use crate::Node;
 use crate::NodeLocation;
 use crate::NodeStr;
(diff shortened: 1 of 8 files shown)


goto-bus-stop (approved):
Makes sense to me

PR #668, opened 2023-09-28
Add JSON line/column information to diagnostics

goto-bus-stop:
This does a few things...
- Work around ariadne using character indices. We build a map of byte indices to character indices to use for this. This takes up to 4x the memory of input source text. Storage could be reduced by only storing for example every 16th character and scan through `char_indices()` from there whenever we look up an index. that would still be easier to do than to patch ariadne.
- Add a `get_line_column()` to diagnostics that uses ariadne to look up the line/column for a byte offset.
- Add a `to_json()` and JSON error structures that match the spec. I'm not sure this belongs in apollo-rs, though JSON is a common output format. This would be used in the router to serialize validation errors.

Targeting main as this is intended for use very soon, likely before the router is ported to the new apollo-rs AST.

### DIFF
--- a/crates/apollo-compiler/src/database/inputs.rs
+++ b/crates/apollo-compiler/src/database/inputs.rs
@@ -9,19 +9,66 @@ use std::sync::Arc;
 #[error("Unknown file ID")]
 struct UnknownFileError;
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct ByteCharIndex {
+    map: Vec<u32>,
+}
+
+impl ByteCharIndex {
+    fn new(input: &str) -> Self {
+        let mut map = vec![0; input.len() + 1];
+        let mut char_index = 0;
+        for (byte_index, _) in input.char_indices() {
+            map[byte_index] = char_index;
+            char_index += 1;
+        }
+
+        // Support 1 past the end of the string, for use in exclusive ranges.
+        map[input.len()] = char_index;
+
+        Self { map }
+    }
+}
+
+/// Source text structure with a precomputed index from byte offsets to character offsets and to
+/// line/column numbers.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MappedSource {
+    ariadne: AriadneSource,
+    index: ByteCharIndex,
+}
+
+impl MappedSource {
+    fn new(input: &str) -> Self {
+        let ariadne = AriadneSource::from(input);
+        let index = ByteCharIndex::new(input);
+
+        Self { ariadne, index }
+    }
+
+    pub(crate) fn map_index(&self, byte_index: usize) -> usize {
+        self.index.map[byte_index] as usize
+    }
+}
+
 /// A Cache implementation for `ariadne` diagnostics.
 ///
 /// Use [`InputDatabase::source_cache`] to construct one.
 #[derive(Clone, PartialEq, Eq)]
 pub struct SourceCache {
-    sources: HashMap<FileId, Arc<AriadneSource>>,
+    sources: HashMap<FileId, Arc<MappedSource>>,
     paths: HashMap<FileId, PathBuf>,
 }
 
 impl SourceCache {
+    pub(crate) fn get_source(&self, id: FileId) -> Option<&'_ MappedSource> {
+        self.sources.get(&id).map(|arc| arc.as_ref())
+    }
+
     pub fn get_line_column(&self, id: FileId, index: usize) -> Option<(usize, usize)> {
         let source = self.sources.get(&id)?;
-        let (_, line, column) = source.get_offset_line(index)?;
+        let char_index = source.map_index(index);
+        let (_, line, column) = source.ariadne.get_offset_line(char_index)?;
         Some((line, column))
     }
 }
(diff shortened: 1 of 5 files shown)


crates/apollo-compiler/src/database/inputs.rs
```diff
@@ -9,19 +9,63 @@ use std::sync::Arc;
...
+    sources: HashMap<FileId, Arc<MappedSource>>,
     paths: HashMap<FileId, PathBuf>,
 }
+
+impl SourceCache {
+    pub(crate) fn get_source(&self, id: FileId) -> Option<&'_ MappedSource> {
+        self.sources.get(&id).map(|arc| arc.as_ref())
+    }
+
+    pub fn get_line_column(&self, id: FileId, index: usize) -> Option<(usize, usize)> {
```
lrlna, line 56: should line/column be its own struct or be part of another struct? feels a bit strange api-wise with a tuple. i imagine there is going to be loads of use cases for this functionality

crates/apollo-compiler/src/diagnostics.rs
```diff
@@ -400,17 +421,35 @@ impl ApolloDiagnostic {
...
+
+    pub fn to_json(&self) -> GraphQLError {
+        let mut locations = vec![];
+
+        if let Some((line, column)) = self.get_line_column() {
+            locations.push(GraphQLLocation { line, column });
+        }
+
+        GraphQLError {
+            message: self.data.to_string(),
```
lrlna, line 449: ok awesome! this should totally get us far in validation swap. i was thinking some time in the future when we are less swamped, we can tweak the messages for serialisable errors even, so the stuff that was written out before with the aid of source code would be more readable on its own.

lrlna (approved):
> This takes up to 4x the memory of input source text

Do you mean this current PR now takes 4x the memory when looking up in source text? I guess we gotta do what we gotta do 😅 

Otherwise looks good! Thanks for getting this over the line!

SimonSapin:
Exposing types that implement `serde::Serialize` makes a lot of sense, users can pick whatever `Serializer` they need.

The choice of `serde_json::Value` is a bit more opinionated: the Router uses `serde_json_bytes::Value` instead. BTW this is only loosely related to JSON syntax, both of these types likely work with other serializers. Rather, this is a choice of memory representation for dynamically-typed maps and arrays.

I see two possible ways forward:

* Just make a choice in apollo-compiler and stick with it
* Try to keep apollo-compiler agnostic regarding this choice. [Execution and introspection](https://github.com/apollographql/apollo-rs/pull/645) will need a representation for variable values, but they could be kept in a separate apollo-rs crate. (`apollo-execution` maybe?). Meanwhile in this PR `Error::extensions` could be removed (it’s always empty anyway in errors created by apollo-compiler, right?) and `Error::path` changed to a vec of `PathElement` below. Or even replace the whole `Error` struct with separate `message` and `path` methods on `ApolloDiagnostic`.

```rust
#[derive(Debug, Clone, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
#[serde(untagged)] // https://serde.rs/enum-representations.html#untagged
pub enum PathElement {
    Field(Name),
    ListItem { index: usize },
}
```

crates/apollo-compiler/src/database/inputs.rs
SimonSapin: +1, `GraphQLLocation` is being added anyway so using it here instead of a tuple makes sens.

SimonSapin:
> Do you mean this current PR now takes 4x the memory when looking up in source text?

If any diagnostic refers to a 1 KB source file, there’s a new `Vec<u32>` (shared with `Arc`) that allocates ~4 KB to map each possible byte offset to a `char` count. It’s kind of a brute-force solution but it can be improved later.

crates/apollo-compiler/src/database/inputs.rs
goto-bus-stop: Haha good thought, I added graphqllocation after adding this method, totally agree.

goto-bus-stop:
I think I want to use `extensions` in the error message improvements, so we can output an error code based on the `DiagnosticData` enum variant. But we can see if that is worthwhile once that actually happens (if we assume `GraphQLError` is only meant to be created by us, then we could simply use a serializable struct with the extension properties we use).

`PathElement` is definitely the right call for paths

crates/apollo-compiler/src/diagnostics.rs
goto-bus-stop: Yeah I have a separate branch for that, mostly adding the names of things to the messages themselves. I think it's actually pretty crucial for landing validation because it's really confusing without.

goto-bus-stop:
Also note to self: ariadne returns zero-indexed line/column, check if that aligns with the spec and document

SimonSapin:
https://spec.graphql.org/October2021/#sec-Errors.Error-result-format
> each location is a map with the keys `line` and `column`, both positive numbers starting from `1` which describe the beginning of an associated syntax element.

It doesn’t define how columns are counted though

goto-bus-stop:
`path` is meant to be a path into the response data which is not relevant yet for us, it doesn't point into the operation. I'll just omit both `path` and `extensions` for now instead of making a decision on format that we may revisit later.

SimonSapin (approved):

PR #669, opened 2023-09-29
Fix fuzzing invariant `Schema::parse(s.serialize()) == s`

SimonSapin:


lrlna (approved):

SimonSapin:
An upcoming PR makes most of this obsolete (implicit v.s. explicit `schema` definition will be represented by a Rust `Option`) and includes the rest.

PR #670, opened 2023-10-04
Add validate methods that don’t require users to create a compiler

SimonSapin:
A compiler is still used internally, until all validation code is ported to the mechanism


goto-bus-stop:
I'll look more closely but my first impression is great, I wanted to eventually have a `&mut errors` argument for validation so we could also implement diagnostic limits for example.

lrlna (approved):
Thank you so much for all your work on this! Much like Renée said - nothing immediately jumps out - and the API is going to be wonderful to use. I think we should move quickly merging this in and wrapping up a beta release. I imagine anything that comes up afterwards we can always fix as we'd still be in beta.

Issue #442, opened 2023-01-29
compiler: Document side effects when adding documents vs adding executables


erikwrede:
When using the compiler with a single GQL-Schema added via `add_type_system` and multiple GQL-Operations that are added after an initial validation of the schema, there can be significant performance differences in the validation of the operations  depending on the procedure. 

According to the documentation, `add_document` should behave similarly to `add_executable`, but with the added ability to accept SDL input. However, it seems to trigger re-validation of the initially validated schema, even without any new types added, which can be costly for large schemas.

For my testing & benchmarking, I used the [GitLab schema](https://gitlab.com/api/graphql) with 60000 lines of SDL. Validation using `apollo-compiler` takes `~ 40 ms` on my machine. Validating a 250 line nested query with inputs and variables against this schema takes `~1ms`. If I use `add_document` for a query against this schema, the operation validation time increases to `~40ms`. I have attached a simple example below. 

The issue seems to be the definition of  `type_definition_files` in the `InputDatabase`.  Since at this stage, all of the documents should be unparsed sources, there is no way to exclude them from the type system in case the root document node doesn't contain any types, am I reading that correctly?

If this is the intended behavior, it would be very helpful to clarify the possible side effects in the documentation or adjust the nomenclature. 
Intuitively, every executable is a document node with definition nodes consisting of only operation and fragment nodes, so there should be no issues with `add_document` in this case.

---
<details>
  <summary>Repro</summary>

```rust
fn bench_simple_query_compiler_with_validation() {
    let query = include_str!("testdata/simple_query.graphql");
    let schema = include_str!("testdata/simple_schema.graphql");

    let mut compiler = ApolloCompiler::new();
    compiler.add_type_system(schema, "schema.graphql");

    compiler.db.object_types();
    compiler.validate();

    let start = Instant::now();

    let file_id = compiler.add_document(query, ""); // FIXME use add_executable instead

    let elapsed_parsing = start.elapsed();
    let mut diagnostics = compiler.db.ast(file_id)
        .errors()
        .into_iter()
        .map(|err| {
            ApolloDiagnostic::SyntaxError(SyntaxError {
                src: compiler.db.source_code(file_id),
                span: (err.index(), err.data().len()).into(),
                message: err.message().into(),
            })
        }).collect::<Vec<ApolloDiagnostic>>();

    let elapsed_parser_syntax_check = start.elapsed();
    // We only want to validate the new operation against the schema here, everything else is already validated
[...]

PR #676, opened 2023-10-06
Add an example measuring parsing time and validation time

SimonSapin:
This is a rough port to apollo-compiler 1.0 of the reproduction code provided in https://github.com/apollographql/apollo-rs/issues/442

This is **not** a proper benchmark and I expect we still have low-hanging fruit in terms of improving performance, so don’t look at the numbers too closely.

----

Example run with GitLab’s 34k lines schema:

https://www.apollographql.com/docs/rover/commands/graphs/#graph-introspect
```
rover graph introspect https://gitlab.com/api/graphql > /tmp/gitlab_schema.graphql
```

… and a trivial query:

```
cargo run --release --example timed \
    /tmp/gitlab_schema.graphql \
    <(echo '{ __typename }')
```
```
Schema parse (1227833 bytes): 63.997 ms
Schema validation: 10.353 ms
Executable document parse (15 bytes): 0.018 ms
Executable document validation: 1.852 ms
```

### DIFF
--- a/crates/apollo-compiler/examples/timed.rs
+++ b/crates/apollo-compiler/examples/timed.rs
@@ -1,0 +1,52 @@
+//! Parse and validate a schema and executable document provided as files.
+//! Print the time taken by each step.
+
+use apollo_compiler::ExecutableDocument;
+use apollo_compiler::Schema;
+use std::process::ExitCode;
+use std::time::Instant;
+
+fn main() -> ExitCode {
+    let mut args = std::env::args_os();
+    let _arg_0 = args.next(); // filename of this program
+    let arg_1 = args.next();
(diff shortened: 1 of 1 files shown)


SimonSapin:
That validation time for a trivial query seems surprisingly high, I wouldn’t expect schema size to have an impact. Worth looking into.

goto-bus-stop:
Looking at a flame graph, it builds a high-level Schema again from the input ASTs when validating an executable document.

SimonSapin:
Good catch! Fixed in https://github.com/apollographql/apollo-rs/pull/726

lrlna (approved):
should definitely also have something like this as a benchmark. but this is already super handy!

PR #697, opened 2023-10-13
Validation: never panic or omit a diagnostic on missing location

SimonSapin:
The "main" location of the diagnostic defaults to offset zero in a pseudo-file displayed as "(no source file)".

Each label of a diagnostic is omitted when its location is None. For this reason, essential information should be in the main message instead of only in labels.

### DIFF
--- a/crates/apollo-compiler/CHANGELOG.md
+++ b/crates/apollo-compiler/CHANGELOG.md
@@ -47,8 +47,20 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   }
   ```
 
+## Fixes
+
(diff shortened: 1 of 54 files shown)


crates/apollo-compiler/CHANGELOG.md
```diff
@@ -17,6 +17,22 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## Maintenance
 ## Documentation-->
 
+# [x.x.x] (unreleased) - 2023-mm-dd
+
+## Fixes
+
+- **Don’t panic in validation or omit diagnostics when a source location is missing - [SimonSapin], [pull/697]:**
```
SimonSapin, line 24: > Don’t panic

and bring a towel

lrlna (approved):

PR #698, opened 2023-10-13
feat(compiler): add get_line_column for Diagnostic

### LINKED ISSUE #692: Port PR #690 to 1.0: JSON line/column info for diagnostics
PR https://github.com/apollographql/apollo-rs/pull/668 was reverted in https://github.com/apollographql/apollo-rs/pull/690 (but kept in the [`compiler-v0.11`](https://github.com/apollographql/apollo-rs/commits/compiler-v0.11) branch) in order to help compiler v1.0 to be merged into main.

Merging this feature into 1.0 with git has proven difficult, so we’ll need to port it manually.

lrlna:
This brings back line/column numbers and serialisability to diagnostics. It essentially fits in the work @goto-bus-stop already did into the new version of the compiler.

The PR includes:
- [x] (re-) introduction of `MappedSource` where we store bite indices for source information
- [x] line/column numbers for each `SourceFile`
- [x] (re-) introduction of `get_line_column` to public API, accessible via Diagnostic<'a>
- [x] (re-) introduction of `to_json` public api for Diagnostic<'a>
- [x] (re-) introduction of `GraphQLError` and `GraphQLLocation`
- [x] creation of `Display` for `DiagnosticData` in order to gracefully display seriasable diagnostic message in `to_json` functionality

closes #692

### DIFF
--- a/crates/apollo-compiler/src/parser.rs
+++ b/crates/apollo-compiler/src/parser.rs
@@ -23,17 +23,6 @@ pub struct Parser {
     tokens_reached: usize,
 }
 
-/// Records for validation information about a file that was parsed
-#[derive(Clone)]
-pub struct SourceFile {
-    pub(crate) path: PathBuf,
-    pub(crate) source_text: String,
-    pub(crate) parse_errors: Vec<apollo_parser::Error>,
-    pub(crate) ariadne: OnceLock<ariadne::Source>,
-}
-
-pub type SourceMap = Arc<IndexMap<FileId, Arc<SourceFile>>>;
-
 /// Parse a schema and executable document from the given source text
 /// containing a mixture of type system definitions and executable definitions.
 /// This is mostly useful for unit tests.
@@ -115,7 +104,7 @@ impl Parser {
             path,
             source_text,
             parse_errors: tree.errors().cloned().collect(),
-            ariadne: OnceLock::new(),
+            source: OnceLock::new(),
         });
         (tree, source_file)
     }
@@ -245,6 +234,44 @@ impl Parser {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MappedSource {
+    ariadne: ariadne::Source,
+    map: Vec<u32>,
+}
+
+impl MappedSource {
+    fn new(input: &str) -> Self {
+        let ariadne = ariadne::Source::from(input);
+
+        let mut map = vec![0; input.len() + 1];
+        let mut char_index = 0;
+        for (byte_index, _) in input.char_indices() {
+            map[byte_index] = char_index;
+            char_index += 1;
+        }
+
+        // Support 1 past the end of the string, for use in exclusive ranges.
+        map[input.len()] = char_index;
+
+        Self { ariadne, map }
+    }
+
+    pub(crate) fn map_index(&self, byte_index: usize) -> usize {
+        self.map[byte_index] as usize
+    }
+}
+/// Records for validation information about a file that was parsed
+#[derive(Clone)]
+pub struct SourceFile {
+    pub(crate) path: PathBuf,
+    pub(crate) source_text: String,
+    pub(crate) parse_errors: Vec<apollo_parser::Error>,
+    pub(crate) source: OnceLock<MappedSource>,
+}
+
+pub type SourceMap = Arc<IndexMap<FileId, Arc<SourceFile>>>;
+
 impl SourceFile {
     /// The filesystem path (or arbitrary string) used in diagnostics
     /// to identify this source file to users.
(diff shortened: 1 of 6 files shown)


crates/apollo-compiler/src/parser.rs
```diff
@@ -23,17 +23,6 @@ pub struct Parser {
     tokens_reached: usize,
 }
 
-/// Records for validation information about a file that was parsed
-#[derive(Clone)]
-pub struct SourceFile {
```
SimonSapin, line 28: (It’s only a mild preference, but I think grouping type definitions at the start of a module helps looking at them together instead of hunting for them in the middle of `impl` blocks that can be much longer. Data structures are essential.)

crates/apollo-compiler/src/parser.rs
```diff
@@ -245,6 +234,44 @@ impl Parser {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MappedSource {
```
SimonSapin, line 238: Can this be `pub(crate)`? I’d like to minimize the public API, so after we make 1.0 "final" we’ll have fewer constraints to evolve things without yet making a 2.0 with breaking changes.

crates/apollo-compiler/src/parser.rs
```diff
@@ -257,8 +284,16 @@ impl SourceFile {
     }
 
     pub fn ariadne(&self) -> &ariadne::Source {
```
SimonSapin, line 286: * Same as above re `pub(crate)`
* We may want this to return `&MappedSource` (and be named `mapped_source`) instead, see below.

crates/apollo-compiler/src/parser.rs
```diff
@@ -257,8 +284,16 @@ impl SourceFile {
...
-        self.ariadne
-            .get_or_init(|| ariadne::Source::from(&self.source_text))
+        &self
+            .source
+            .get_or_init(|| MappedSource::new(&self.source_text))
+            .ariadne
+    }
+
+    pub fn get_line_column(&self, index: usize) -> Option<(usize, usize)> {
+        let char_index = self.source.get()?.map_index(index);
```
SimonSapin, line 294: `OnceLock::get` returns `None` if the value doesn’t happen to already be initialized. We want to call `get_or_init` instead, or something that does (see method above)

crates/apollo-compiler/src/validation/mod.rs
```diff
@@ -266,6 +287,40 @@ impl DiagnosticData {
...
+            })
+    }
+
+    /// Get serde_json serialisable version of the current diagnostic.
+    pub fn to_json(&self) -> GraphQLError {
+        let mut locations = vec![];
+
+        if let Some(location) = self.get_line_column() {
+            locations.push(location);
+        }
```
SimonSapin, line 313: nit:
```suggestion
        let locations = self.get_line_column().into_iter().collect();
```

crates/apollo-compiler/src/validation/mod.rs
```diff
@@ -266,6 +287,40 @@ impl DiagnosticData {
...
+
+        GraphQLError {
+            // TODO(@lrlna): this needs DiagnosticData to have a Display impl
+            // using only diagnostic messages.
+            message: self.to_string(),
```
SimonSapin:
```suggestion
            message: self.data.details.to_string(),
```

This relies on the `Display` impl generated by `derive(thiserror::Error)`  based on attributes like `#[error("type extension for undefined type `{name}`")]`.

It the same that we pass to `ariadne::ReportBuilder::with_message`. Later we can match on each `Detail` variant and add more details that Ariadne reports have in labels/help/note, but this is a good starting point.

crates/apollo-compiler/src/parser.rs
lrlna: omg i am the complete opposite haha!! i feel like i always need my structs with my impls so i can look up to see what i am working with 😅 😅 

i'll move this one back up since it seems like everything else uses your system

crates/apollo-compiler/src/validation/mod.rs
SimonSapin: The Router also ended up needing this:

```rust
    pub fn message(&self) -> &impl fmt::Display {
        &self.data.details
    }
```

crates/apollo-compiler/src/validation/mod.rs
```diff
@@ -266,6 +287,38 @@ impl DiagnosticData {
     }
 }
 
+impl<'a> Diagnostic<'a> {
+    /// Get the line and column number where this diagnostic was raised.
+    pub fn get_line_column(&self) -> Option<GraphQLLocation> {
+        // let source = self.sources.get(&id)?;
+        // let char_index = source.map_index(index);
+        // let (_, line, column) = source.ariadne.get_offset_line(char_index)?;
+        // Some((line, column))
```
SimonSapin, line 296: nit: remove this?

SimonSapin (approved):

Issue #710, opened 2023-10-19
Make ast::Name a type wrapper that checks validity


SimonSapin:
This is breaking change that we should either make before 1.0 leaves beta, or decide not to make.

---

In 1.0.0-beta.4 `apollo_compiler::ast::Name` is a type alias for `apollo_compiler::NodeStr`, a smart string type with infallible conversion from arbitrary `&str`. As a result it is possible to generate invalid syntax:

```rust
let mut doc = apollo_compiler::ast::Document::parse("{ field }", "input.graphql");
if let apollo_compiler::ast::Definition::OperationDefinition(op) = &mut doc.definitions[0] {
    op.make_mut().name = Some("++".into())
}
assert_eq!(
    doc.serialize().no_indent().to_string(),
    "query ++ { field }"
)
```

This simple example uses AST, but even full validation in `Schema` or `ExecutableDocument` would not catch this.

We could change validation to find every `Name` and check its validity (must be made of one `NameStart` character followed by zero or more `NameContinue` characters) but it would be easy for validation code to miss some, and delay finding a problem until the user actually calls validation.

## Proposal

Instead, `ast::Name` could be changed to a wrapper type of `NodeStr` that checks validity. Conversion from `&str` would be fallible. To avoid run-time `unwrap` when converting from a literal string, there could be a `name!("example")` macro that checks validity at compile-time.

PR #713, opened 2023-10-20
Check validity of `ast::Name` (and assorted breaking changes) 

SimonSapin:
Fixes https://github.com/apollographql/apollo-rs/issues/710

Featuring unsafe tagged pointers and Miri on CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,6 +57,13 @@ workflows:
             parameters:
               platform: [linux, macos, windows]
               rust_channel: [stable, nightly]
+  miri:
+    jobs:
+      - miri:
+          name: Miri (rust nightly on linux, subset of tests)
+          matrix:
+            parameters:
+              platform: [linux]
   security-scans:
     jobs:
       - secops/gitleaks:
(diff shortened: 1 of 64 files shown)


goto-bus-stop:
I like it a lot overall and I'm not opposed to a bit of easily auditable unsafe. How much use would we expect the no-alloc static representation to have? Well-known names like "Query" or directives implemented by a server would use it but I'm wondering if that's enough allocations to be worth the complexity 😅  There is an alternative option to validate the `name!()` input at compile time but still copy it into an allocation at runtime right? That might be a footgun if you use it in a hot loop I guess.

SimonSapin:
I’d expect almost all static `NodeStr`s to come from the `name!` macro. Yes, having `name!` still check validity at compile-time but allocate on the heap is also a possibility. And yes, only a small minority of all `Name`s involved e.g. when handling one request in the Router come from `name!`.

Even so, my opinion is that the complexity here is far from excessive: it’s entirely encapsulated in a ~300 lines modules (~100 of which are safe boilerplate impls). And verification by Miri gives some confidence in correctness. We routinely rely on Rowan and other libraries that also use unsafe internally (in specific modules, encapsulated with a safe abstraction.)

goto-bus-stop (approved):

PR #718, opened 2023-11-02
feat(parser, compiler): add parsing a type reference from a string

### LINKED ISSUE #715: add parsing support for `Type`
Similar to #681, `@field` directive can be provided with `Type` information in a string form, which will then need to be parsed before being used in the schema. It seems like it does not need to be validated specifically (but we should do it anyways)

To make this change, we need to:

1. add `parse_type` method to `apollo-parser`
2. add a `cst::Type` specific method for `SyntaxTree` implementation to return the CST node, similar to doc.document(), and field_set.field_set()
3. create a `parse_type` method in `apollo-compiler`
4. add `Type` struct to `apollo-compiler`'s schema that would also be able to include `BuildErrors`
5. add a `parse` and `validate` methods to the `Type` struct impl

lrlna:
To parse standalone source text like `[Type]!`, for example for the federation directive [`@fields(type:)`](https://specs.apollo.dev/join/v0.3/#@field). 

closes #715

### DIFF
--- a/crates/apollo-compiler/src/parser.rs
+++ b/crates/apollo-compiler/src/parser.rs
@@ -3,7 +3,6 @@ use crate::ast::from_cst::Convert;
 use crate::ast::Document;
 use crate::executable;
 use crate::schema;
-use crate::schema::FieldType;
 use crate::schema::SchemaBuilder;
 use crate::validation::Details;
 use crate::validation::Diagnostics;
@@ -248,19 +247,18 @@ impl Parser {
         &mut self,
         source_text: impl Into<String>,
         path: impl AsRef<Path>,
-    ) -> schema::FieldType {
+    ) -> Option<schema::FieldType> {
         let (tree, source_file) =
             self.parse_common(source_text.into(), path.as_ref().to_owned(), |parser| {
                 parser.parse_type()
             });
         let file_id = FileId::new();
-        match tree.ty().convert(file_id) {
-            Some(ty) => return Ok(FieldType{
-                sources: Arc::new([(file_id, source_file)]),
+        tree.ty().convert(file_id).map(|ty| {
+            return schema::FieldType {
+                sources: Arc::new([(file_id, source_file)].into()),
                 ty,
-            }),
-            None => ,
-        }
+            };
+        })
     }
 
     /// What level of recursion was reached during the last call to a `parse_*` method.
--- a/crates/apollo-compiler/src/parser.rs
+++ b/crates/apollo-compiler/src/parser.rs
@@ -196,7 +196,7 @@ impl Parser {
         self.parse_ast(source_text, path).to_mixed()
     }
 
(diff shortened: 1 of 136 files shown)


crates/apollo-compiler/src/parser.rs
```diff
@@ -234,6 +236,35 @@ impl Parser {
...
 
+    /// Parse the given source of a field type.
+    ///
+    /// `path` is the filesystem path (or arbitrary string) used in diagnostics
+    /// to identify this source file to users.
+    pub fn parse_field_type(
+        &mut self,
+        source_text: impl Into<String>,
+        path: impl AsRef<Path>,
+    ) -> Result<(schema::FieldType, Diagnostics), Diagnostics> {
```
lrlna, line 247: Returning a `Result` here for field types which are not syntactically valid (and therefore will not have a FieldType), but will have syntax diagnostics, e.g. `[]` as a field type.

goto-bus-stop: `Diagnostics` can only have syntax errors or limit errors here, so perhaps we can just have `Result<FieldType, Diagnostics>` and always return `Err(errors)` if there are any?

goto-bus-stop:
Also Also Also, should this just be parsing into a `Type` directly instead of into a `FieldType` wrapper? A type reference doesn't *have* to be in a field, field/directive argument types are the same.

SimonSapin:
Agree on both points: this can use `ast::Type` without introducing a `FieldType` struct, and the `Ok` result doesn’t need to contain diagnostics (or conversely, return `Err` if any diagnostic)

goto-bus-stop:
@SimonSapin Ready for review again I think!

crates/apollo-parser/src/parser/mod.rs
```diff
@@ -146,12 +146,15 @@ impl<'a> Parser<'a> {
...
-            syntax_tree::SyntaxTreeWrapper::FieldSet(_) => {
+            syntax_tree::SyntaxTreeWrapper::Type(_)
+            | syntax_tree::SyntaxTreeWrapper::FieldSet(_) => {
                 unreachable!("parse constructor can only construct a document")
             }
         }
     }
 
+    /// Parse selection set tokens. Useful for specifically parsing selections
+    /// sets which are part of specific directives, like `@requires`.
```
SimonSapin, line 157:
```suggestion
    /// Parse a selection set with optional outer braces.
    /// This is the expected format of the string value of the `fields` argument of some directives
    /// like [`@requires`](https://www.apollographql.com/docs/federation/federated-types/federated-directives/#requires).
```

SimonSapin: Braces? Brackets?

goto-bus-stop: IMO braces means `{}`, and brackets can mean that or one of the other pairs

Issue #666, opened 2023-09-27
Parse repeated syntax with loops instead of recursion


SimonSapin:
PR https://github.com/apollographql/apollo-rs/pull/662 made the parser recursion limit apply to every case recursion was used (hopefully i didn’t miss any). In same cases (nested selection set, nested values, arguably nested list types) recursion does make parsing significantly easier. But as of this writing recursion is also used in many case to parse repetition, where it could be easily replaced with a loop.

This issue is fixed if the test added by https://github.com/apollographql/apollo-rs/pull/662 still passes after uncommenting this assertion:

https://github.com/apollographql/apollo-rs/blob/865be968ccac214c0192359ccd9461f363a88385/crates/apollo-parser/src/parser/mod.rs#L636

Issue #659, opened 2023-09-26
parser::grammar::object::implements_interfaces is too permissive


SimonSapin:
## Description

This parser function accepts any combination of name tokens and `&` tokens, whereas the grammar requires them to alternate:

https://spec.graphql.org/October2021/#ImplementsInterfaces
> *ImplementsInterfaces*:
> * *ImplementsInterfaces* `&` *NamedType*
> * `implements` `&`[opt] *NamedType*

### Steps to reproduce

Add three `.graphql` files under `crates/apollo-parser/test_data/parser/err/`:

```graphql
type Obj implements A & { field: Int }
```
```graphql
type Obj implements A B { field: Int }
```
```graphql
type Obj implements A && B { field: Int }
```

Additionally, apply this diff to an existing test case:

```diff
--- a/crates/apollo-parser/test_data/parser/ok/0018_object_type_definition.graphql
+++ b/crates/apollo-parser/test_data/parser/ok/0018_object_type_definition.graphql
@@ -1,5 +1,8 @@
 "description of type"
-type Person implements Human {
+type Person implements
+    & Human
+    & Entity
+{
     """
     description of field
     """
```

### Expected result

All of the new `err` files contain errors.

### Actual result

None of them do.

PR #721, opened 2023-11-03
chore(parser): remove recursion when parsing non-nested lists of nodes

goto-bus-stop:
GraphQL has many list-like things that have to be non-empty. In those cases I try to explicitly parse one node and error if not found, then parse with a `while let` if the next token indicates another of the same node is coming. I wrapped it up in a helper function/macro first but it's frankly just as much code at each usage site so I didn't find it worth it for this very small pattern.

Fixes https://github.com/apollographql/apollo-rs/issues/666
Also fixes https://github.com/apollographql/apollo-rs/issues/659, and the same for unions (`union X = A || B |` is now illegal)

### DIFF
--- a/crates/apollo-parser/CHANGELOG.md
+++ b/crates/apollo-parser/CHANGELOG.md
@@ -16,6 +16,26 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## Maintenance
 
 ## Documentation -->
+# [x.x.x] (unreleased) - 2023-mm-dd
+
+## Fixes
+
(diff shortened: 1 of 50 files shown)


lrlna (approved):
This looks good! The only thing I wanted to say is that last time we "fixed" recursion limits, we broke the router because the recursion high number technically changes. So I think this should be a "breaking"

goto-bus-stop:
I think we shouldn't consider the exact number a breaking change as it is tied to the internal details of how we parse stuff. Last time the correction was *increasing* the number which would trigger the limit on inputs that previously didn't, which made it a breaking change. This time any input that worked in the past still works, just more inputs are accepted than before. But it's worth calling it out better in the changelog

Issue #664, opened 2023-09-27
schema printer does not support multiline comments


dariuszkuc:
## Description

GraphQL supports both single line (wrapped in `"`) and multiline comments (wrapped in `"""`). `apollo-rs` currently does not support multiline comments and coalesce it to a single line, i.e.

```
"""
First line comment
Second line comment
"""
```

becomes

```
"First line comment\nSecond line comment"
```

### Steps to reproduce

```rust
#[test]
    fn comments() {
        let raw_comment = r#""""
First Line
Second Line
""""#;
        let multiline_comment = "First Line
Second Line";

        let foo = ObjectType {
            name: Name::new("foo"),
            description: Some(NodeStr::new(multiline_comment)),
            directives: vec![],
            fields: IndexMap::new(),
            implements_interfaces: IndexMap::new(),
        };

        let mut schema = Schema::new();
        schema
            .types
            .insert(Name::new("foo"), ExtendedType::Object(Node::new(foo)));
        println!("{}", schema);
    }
```

### Expected result

Printed descriptions should preserve multiline comments.

### Actual result

When using `raw_string` (from above) prints following schema

```graphql
"\"\"\"\nFirst Line\nSecond Line\n\"\"\""
type foo
```

When using `multiline_comment` (from above) prints following schema

```graphql
"First Line\nSecond Line"
type foo
```

### Environment

- Operating system and version: MacOS Ventura
- Shell (bash/zsh/powershell): `zsh`
- `apollo-rs` crate: `mut` branch
- Crate version: `1.72.0`

SimonSapin:
Implementation PR: https://github.com/apollographql/apollo-rs/pull/724

See also stylistic question: https://github.com/apollographql/apollo-rs/pull/724#issuecomment-1794633948

PR #724, opened 2023-11-06
Serialize multi-line strings as block strings 

SimonSapin:
Fixes https://github.com/apollographql/apollo-rs/issues/664

### DIFF
--- a/crates/apollo-compiler/src/ast/serialize.rs
+++ b/crates/apollo-compiler/src/ast/serialize.rs
@@ -12,7 +12,7 @@ pub struct Serialize<'a, T> {
 
 #[derive(Debug, Clone)]
 pub(crate) struct Config<'a> {
-    ident_prefix: Option<&'a str>,
+    indent_prefix: Option<&'a str>,
 }
 
 pub(crate) struct State<'config, 'fmt, 'fmt2> {
@@ -29,13 +29,13 @@ impl<'a, T> Serialize<'a, T> {
     /// `prefix` is repeated at the start of each line by the number of indentation levels.
     /// The default is `"  "`, two spaces.
     pub fn indent_prefix(mut self, prefix: &'a str) -> Self {
-        self.config.ident_prefix = Some(prefix);
+        self.config.indent_prefix = Some(prefix);
         self
     }
 
     /// Disable indentation and line breaks
     pub fn no_indent(mut self) -> Self {
-        self.config.ident_prefix = None;
+        self.config.indent_prefix = None;
         self
     }
 }
@@ -43,7 +43,7 @@ impl<'a, T> Serialize<'a, T> {
 impl Default for Config<'_> {
     fn default() -> Self {
         Self {
-            ident_prefix: Some("  "),
+            indent_prefix: Some("  "),
         }
     }
 }
@@ -89,7 +89,7 @@ impl<'config, 'fmt, 'fmt2> State<'config, 'fmt, 'fmt2> {
     }
 
     fn new_line_common(&mut self, space: bool) -> fmt::Result {
-        if let Some(prefix) = self.config.ident_prefix {
+        if let Some(prefix) = self.config.indent_prefix {
             self.write("\n")?;
             for _ in 0..self.indent_level {
                 self.write(prefix)?;
@@ -104,7 +104,7 @@ impl<'config, 'fmt, 'fmt2> State<'config, 'fmt, 'fmt2> {
     fn require_new_line(&mut self) -> fmt::Result {
         let prefix = self
             .config
-            .ident_prefix
+            .indent_prefix
             .expect("require_new_line called with newlines disabled");
         self.write("\n")?;
         for _ in 0..self.indent_level {
@@ -114,13 +114,13 @@ impl<'config, 'fmt, 'fmt2> State<'config, 'fmt, 'fmt2> {
     }
 
     pub(crate) fn newlines_enabled(&self) -> bool {
-        self.config.ident_prefix.is_some()
+        self.config.indent_prefix.is_some()
     }
 
     pub(crate) fn on_single_line<R>(&mut self, f: impl FnOnce(&mut Self) -> R) -> R {
-        let indent_prefix = self.config.ident_prefix.take();
+        let indent_prefix = self.config.indent_prefix.take();
         let result = f(self);
-        self.config.ident_prefix = indent_prefix;
+        self.config.indent_prefix = indent_prefix;
         result
     }
 }
--- a/crates/apollo-compiler/src/ast/serialize.rs
+++ b/crates/apollo-compiler/src/ast/serialize.rs
@@ -100,6 +100,19 @@ impl<'config, 'fmt, 'fmt2> State<'config, 'fmt, 'fmt2> {
         Ok(())
     }
 
+    /// Panics if newlines are disabled
+    fn require_new_line(&mut self) -> fmt::Result {
+        let prefix = self
+            .config
+            .ident_prefix
+            .expect("require_new_line called with newlines disabled");
+        self.write("\n")?;
+        for _ in 0..self.indent_level {
+            self.write(prefix)?;
+        }
+        Ok(())
+    }
+
     pub(crate) fn newlines_enabled(&self) -> bool {
         self.config.ident_prefix.is_some()
     }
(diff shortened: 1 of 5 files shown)


SimonSapin:
Should we also prefer block strings for single-line descriptions? (As opposed to string `Value`s)

crates/apollo-compiler/test_data/ok/0015_supergraph.graphql
```diff
@@ -22,6 +22,10 @@ type Amazon {
 
 union Body = Image | Text
 
+"""
+  Usually made of pages
+
+"""
```
goto-bus-stop, line 28: could we put these in a separate test_data file specifically for string serialization tests?

crates/apollo-compiler/src/ast/serialize.rs
```diff
@@ -98,6 +98,19 @@ impl<'config, 'fmt, 'fmt2> State<'config, 'fmt, 'fmt2> {
         Ok(())
     }
 
+    /// Panics if newlines are disabled
+    fn require_new_line(&mut self) -> fmt::Result {
+        let prefix = self
+            .config
+            .ident_prefix
```
goto-bus-stop, line 105: Oh I was confused reading this, but it's meant to be `iNdent_prefix`

crates/apollo-compiler/src/ast/serialize.rs
```diff
@@ -829,6 +845,89 @@ fn serialize_string_value(state: &mut State, mut str: &str) -> fmt::Result {
...
+                break;
+            }
+        }
+    }
+    state.require_new_line()?;
+    state.write(TRIPLE_QUOTE)
+}
+
+fn prefer_block_string(value: &str) -> bool {
+    // If a value is multi-line, it likely looks nicer as a block string
```
goto-bus-stop, line 879: Maybe also if the value has `"` in it?
```graphql
"""Surrounded by "double" quotes"""
```

goto-bus-stop (commented):
I like the implementation, I don't really have an opinion on whether descriptions should always use block strings (If I was handwriting a schema i might use block strings by default for types and use line strings if all fields in a type have short descriptions ;p)

crates/apollo-compiler/src/ast/serialize.rs
SimonSapin: Eh, it’s been wrong all this time :sweat_smile:

crates/apollo-compiler/test_data/ok/0015_supergraph.graphql
SimonSapin: Done!

crates/apollo-compiler/src/ast/serialize.rs
lrlna: my 5 euros (inflation) is that we should do what graphql js does, otherwise each implementation starts having their own SDL display 😅

crates/apollo-compiler/src/ast/serialize.rs
```diff
@@ -799,6 +812,9 @@ fn curly_brackets_space_separated<T>(
 }
 
 fn serialize_string_value(state: &mut State, mut str: &str) -> fmt::Result {
+    if state.newlines_enabled() && prefer_block_string(str) && can_be_block_string(str) {
+        return serialize_block_string(state, str);
+    }
     // TODO: use block string when it would be possible AND would look nicer
```
goto-bus-stop, line 818: This comment is obsolete 🥳

SimonSapin:
It looks like graphql-js:

* [Always prefers block strings for descriptions](https://github.com/graphql/graphql-js/blob/0b7590f0a2b65e6210da2e49be0d8e6c27781af2/src/utilities/printSchema.ts#L306)
* [`StringValue` objects can contain a `isBLockString` boolean flag](https://github.com/graphql/graphql-js/blob/0b7590f0a2b65e6210da2e49be0d8e6c27781af2/src/language/printer.ts#L144-L147)… but [nothing ever sets it](https://github.com/search?q=repo%3Agraphql%2Fgraphql-js%20isBlockString&type=code)?
* Block string printing has [some heuristics](https://github.com/graphql/graphql-js/blob/0b7590f0a2b65e6210da2e49be0d8e6c27781af2/src/language/blockString.ts#L146-L153) to decide whether to add leading/trailing newlines

@IvanGoncharov did I miss something?

goto-bus-stop:
Did the new pushes change something or is it just a rebase?

SimonSapin:
Rebase + removing the obsolete comment

IvanGoncharov:
> [StringValue objects can contain a isBLockString boolean flag](https://github.com/graphql/graphql-js/blob/0b7590f0a2b65e6210da2e49be0d8e6c27781af2/src/language/printer.ts#L144-L147)… but [nothing ever sets it](https://github.com/search?q=repo%3Agraphql%2Fgraphql-js%20isBlockString&type=code)?

@SimonSapin The flag called, `block` is `isBLockString` just an alias in this particular destruction.
It is set in parser:
https://github.com/graphql/graphql-js/blob/0b7590f0a2b65e6210da2e49be0d8e6c27781af2/src/language/parser.ts#L713
This allows graphql-js to do round-trip where block string is parsed and when printed as block string.

> [Always prefers block strings for descriptions](https://github.com/graphql/graphql-js/blob/0b7590f0a2b65e6210da2e49be0d8e6c27781af2/src/utilities/printSchema.ts#L306)

Yes, note that not every string can be printed as block string.

> Block string printing has [some heuristics](https://github.com/graphql/graphql-js/blob/0b7590f0a2b65e6210da2e49be0d8e6c27781af2/src/language/blockString.ts#L146-L153) to decide whether to add leading/trailing newlines

Yes, in graphql-js we had a bug where by printing string as block string we were loosing some characters so I spend bunch of time figuring out different heuristics that preserved all characters.

crates/apollo-compiler/src/ast/serialize.rs
```diff
@@ -831,6 +842,89 @@ fn serialize_string_value(state: &mut State, mut str: &str) -> fmt::Result {
...
+                None // skip whitespace-only lines
+            }
+        });
+        each_line_indent_utf8_len.min().unwrap_or(0)
+    };
+    // If there is common indent `BlockStringValue` would remove it
+    // and incorrectly round-trip to a different value.
+    common_indent == 0
+}
+
```
IvanGoncharov, line 927: It is worth noting that each time the function `value.contains` is called, it iterates through the entire string. The same goes for each call of `value.split`. 

However, graphql-js optimizes this process by performing a single iteration through the string. This optimization is particularly useful for long descriptions. 
https://github.com/graphql/graphql-js/blob/0b7590f0a2b65e6210da2e49be0d8e6c27781af2/src/language/blockString.ts#L54.

crates/apollo-compiler/src/ast/serialize.rs
```diff
@@ -831,6 +842,89 @@ fn serialize_string_value(state: &mut State, mut str: &str) -> fmt::Result {
...
+fn prefer_block_string(value: &str) -> bool {
+    // If a value is multi-line, it likely looks nicer as a block string
+    value.contains('\n')
+}
+
+/// Is it possible to create a serialization that, when fed through
+/// [BlockStringValue](https://spec.graphql.org/October2021/#BlockStringValue()),
+/// returns exactly `value`?
+fn can_be_block_string(value: &str) -> bool {
```
IvanGoncharov, line 883: This function is missing a check for non-printable characters like here:
https://github.com/graphql/graphql-js/blob/0b7590f0a2b65e6210da2e49be0d8e6c27781af2/src/language/blockString.ts#L66-L79

Do you checking for them in some other place?

IvanGoncharov: https://spec.graphql.org/draft/#note-65771

SimonSapin: This non-normative note is also present [in the October 2021 version](https://spec.graphql.org/October2021/#note-65771) but in that case it is consistent with the grammar:

> [StringCharacter](https://spec.graphql.org/October2021/#StringCharacter)
> * SourceCharacter but not `"` or `\` or LineTerminator
> * `\u` EscapedUnicode
> * `\` EscapedCharacter
>
> [SourceCharacter](https://spec.graphql.org/October2021/#SourceCharacter)
> * U+0009
> * U+000A
> * U+000D
> * U+0020–U+FFFF

In the draft version of the spec [SourceCharacter](https://spec.graphql.org/draft/#SourceCharacter) was changed to “Any Unicode scalar value”, including non-printable ones. I think the note should have been removed at the same time.

SimonSapin:
New commit to make stylistic choices closer to graphql-js:
* Prefer block strings for descriptions
* For short single-line values serialize block strings as `"""example"""`, except for trailing characters that would make it incorrect

goto-bus-stop (approved):

PR #725, opened 2023-11-06
Improve error span for lowercase_enum_value

goto-bus-stop:
Before:
image

After:
image

Also removing a temporary string allocation while i'm there


SimonSapin (approved):
Nice!

PR #726, opened 2023-11-06
Skip schema AST conversion in ExecutableDocument::validate

SimonSapin:
In the example at https://github.com/apollographql/apollo-rs/pull/676, validation of a trivial (very small) query against a large schema goes from 1.5~2.1 ms to 0.2 ms.


goto-bus-stop (approved):
sweet

Issue #716, opened 2023-11-01
stack overflow on fragment cycle


qwerdenkerXD:
## Description

During validation aren't all fragment cycles detected.
This causes a stack overflow in latest beta `apollo-compiler@1.0.0-beta.5`

### Steps to reproduce

Validate the following query for your schema:
```graphql
query Introspection{
  __schema {
    types {
      ...cycle
    }
  }
}

fragment cycle on __Type {
  kind
  ...frag
}

fragment frag on __Type {
  kind
  ofType {
    ...cycle
  }
}
```

[Here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&code=use+apollo_compiler%3A%3A%7B%0D%0A++++ExecutableDocument%2C%0D%0A++++Schema%0D%0A%7D%3B%0D%0A%0D%0Afn+main%28%29+%7B%0D%0A++++let+schema%3A+%26str+%3D+r%23%22%0D%0A++++++type+Query+%7B%0D%0A++++++++useless%3A+Boolean%21%0D%0A++++++%7D%0D%0A++++%22%23%3B%0D%0A%0D%0A++++let+query%3A+%26str+%3D+r%23%22%0D%0A++++++query+Introspection%7B%0D%0A++++++++__schema+%7B%0D%0A+++++++++++types+%7B%0D%0A++++++++++++...cycle%0D%0A++++++++++%7D%0D%0A++++++++%7D%0D%0A++++++%7D%0D%0A%0D%0A++++++fragment+cycle+on+__Type+%7B%0D%0A++++++++kind%0D%0A++++++++...frag%0D%0A++++++%7D%0D%0A%0D%0A++++++fragment+frag+on+__Type+%7B%0D%0A++++++++kind%0D%0A++++++++ofType+%7B%0D%0A++++++++++...cycle%0D%0A++++++++%7D%0D%0A++++++%7D%0D%0A++++%22%23%3B%0D%0A%0D%0A++++let+schema%3A+Schema+%3D+Schema%3A%3Aparse%28schema%2C+%22schema%22%29%3B%0D%0A++++let+document+%3D+ExecutableDocument%3A%3Aparse%28%26schema%2C+query%2C+%22query%22%29%3B%0D%0A%0D%0A++++schema.validate%28%29.unwrap%28%29%3B%0D%0A++++document.validate%28%26schema%29%3B%0D%0A%7D) a code sample.

### Expected result

The validation should pass and in the diagnostics should be one like:
```json
[...]

PR #733, opened 2023-11-09
Fix validation stack overflow for fragment cycles inside nested fields

goto-bus-stop:
Previously fragment cycle detection only counted fragments that are
directly recursive, where the top-level selection for the fragment
included a fragment spread of itself. But spreading a fragment anywhere
inside a nested field is also problematic and prohibited by spec,
because even if the field you're spreading into is nullable, the client
can't know how deep that recursion goes, and it could be infinite.

Fixes https://github.com/apollographql/apollo-rs/issues/716

### DIFF
--- a/crates/apollo-compiler/test_data/diagnostics/0092_recursive_fragment_spread.txt
+++ b/crates/apollo-compiler/test_data/diagnostics/0092_recursive_fragment_spread.txt
@@ -1,12 +1,23 @@
 Error: `fragA` fragment cannot reference itself
-    ╭─[0092_recursive_fragment_spread.graphql:13:1]
+    ╭─[0092_recursive_fragment_spread.graphql:14:1]
     │
- 13 │ fragment fragA on Human { name, ...fragB }
+ 14 │ fragment fragA on Human { name, ...fragB }
     │ ───────┬──────  
     │        ╰──────── recursive fragment definition
     │ 
- 15 │ fragment fragC on Human { name, ...fragA }
+ 16 │ fragment fragC on Human { name, ...fragA }
     │                                 ────┬───  
     │                                     ╰───── refers to itself here
+────╯
+Error: `cycle1` fragment cannot reference itself
+    ╭─[0092_recursive_fragment_spread.graphql:19:1]
+    │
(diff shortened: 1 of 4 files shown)


crates/apollo-compiler/test_data/diagnostics/0092_recursive_fragment_spread.txt
```diff
@@ -1,12 +1,23 @@
...
     │
- 13 │ fragment fragA on Human { name, ...fragB }
+ 14 │ fragment fragA on Human { name, ...fragB }
     │ ───────┬──────  
     │        ╰──────── recursive fragment definition
     │ 
- 15 │ fragment fragC on Human { name, ...fragA }
+ 16 │ fragment fragC on Human { name, ...fragA }
     │                                 ────┬───  
     │                                     ╰───── refers to itself here
```
SimonSapin, line 10: The "refers to itself" message should probably be tweaked

SimonSapin: Should the diagnostic even contain a `Vec` of locations to show the complete cycle?

goto-bus-stop: This is easy to do currently. would be helpful to have more context lines :/
```
Error: `cycle` fragment cannot reference itself
    ╭─[q.graphql:9:1]
    │
  9 │ fragment cycle on __Type {
    │ ───────┬──────
    │        ╰──────── recursive fragment definition
    │
 11 │   ...frag
    │   ───┬───
    │      ╰───── `cycle` references `frag` here...
    │
 17 │     ...cycle
    │     ────┬───
    │         ╰───── `frag` circularly references `cycle` here
```

goto-bus-stop: It would be ideal here if, in TTY formatting, `cycle` in the final label uses the same colour as `fragment cycle` in the primary label 🙈

SimonSapin (approved):

PR #737, opened 2023-11-10
Introspection

yanns:
update to main


SimonSapin:
Thanks! Are you already using this?

We plan to focus on introspection again starting next week, and likely merge it as a module of the compiler crate.

yanns:
> Are you already using this?

Not yet, but I'm working on it, as an alternative to the current introspection based on js.
I'm in the lucky situation that I can try things on shadowed GraphQL requests, being able to compare responses without impacting users.

Issue #740, opened 2023-11-13
`fatal error: 'cassert' file not found` on macOS CI compiling libfuzzer-sys v0.4.7


SimonSapin:
Successful CI run three days ago on `main`, including v0.4.7 of the libfuzzer-sys crate: https://app.circleci.com/pipelines/github/apollographql/apollo-rs/1882/workflows/ce0f3ce4-41d1-4821-85ea-b6482a2af70f/jobs/12847?invite=true#step-106-8147_33

Re-run of the same job now fails to compile some C++ code in the build script of the same version of that crate: https://app.circleci.com/pipelines/github/apollographql/apollo-rs/1882/workflows/a434f1e4-eacb-4adf-acd2-9ad887228ea0/jobs/12904?invite=true#step-106-29330_68

```
error: failed to run custom build command for `libfuzzer-sys v0.4.7`
[…]
  cargo:warning=libfuzzer/FuzzerDefs.h:14:10: fatal error: 'cassert' file not found
[…]
  error occurred: Command "c++" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-gdwarf-2" "-fno-omit-frame-pointer" "-m64" "--target=x86_64-apple-darwin" "-mmacosx-version-min=10.7" "-Wall" "-Wextra" "-std=c++17" "-fno-omit-frame-pointer" "-w" "-o" "/Users/distiller/project/target/debug/build/libfuzzer-sys-82fe81b55dd69567/out/libfuzzer/FuzzerTracePC.o" "-c" "libfuzzer/FuzzerTracePC.cpp" with args "c++" did not execute successfully (status code exit status: 1).
```

Could something have changed in CircleCI’s base environment?

This is blocking PRs like https://github.com/apollographql/apollo-rs/pull/739 from merging.

PR #741, opened 2023-11-14
Upgrade XCode on CI

SimonSapin:
Fixes https://github.com/apollographql/apollo-rs/issues/740 somehow 🤷

See https://circleci.com/docs/using-macos/#supported-xcode-versions


goto-bus-stop (approved):
if it works it works!

PR #747, opened 2023-11-16
Add ariadne wrapper `ToDiagnostic` / `CliReport`

goto-bus-stop:
This is an attempt to provide a reusable API on top of ariadne that downstream users can use to get similar formatting to apollo-rs. There is especially a need for this as ariadne uses character indices while apollo-rs uses byte indices, so we need to convert the numbers in between.

This adds public API:
- `ToDiagnostic::report(&mut CliReport)` - a trait method that custom errors
  can implement to provide advanced diagnostics for printing to CLI
- `CliReport` - a builder for advanced diagnostics, provided to `ToDiagnostic::report`
- `Diagnostic<T>` - a wrapper for custom error types that provides
  pretty-printed `fmt` and `io::Write` output using the `ToDiagnostic`
  trait
- `ResultExt::to_diagnostic(SourceMap)` - extension method that wraps
  the error branch of a `Result` in a `Diagnostic<>`, so you don't need
  to do a verbose `.map_err` dance

This API requires that users store the error location inside their own
error type.

Inconveniences:
- If you're using an error enum, you have to store the location in every
  branch, and match in your `.to_report()` implementation
- This does not include `DiagnosticList`, which may be useful, as
  collecting multiple errors is a common use case and having to write
  your own `fmt` impls and a custom wrapper type for `Vec<YourError>` is
  busy-work

The things to do here:
- [x] Maintain the color/no-color settings from main, this is why tests break at the moment
- I'd like to rework at least executable validation to use this similarly to how build errors use them -> moved https://github.com/apollographql/apollo-rs/issues/756
- Should we fit schema coordinate / operation expression based error locations in this somehow?
  - No, this API will not have schema coordinates at all. The diagnostics that *use* this API can have coordinates/paths/expressions, though.
- validate usability in a third-party crate; document happy path
- [x] Consider if we need to provide a JSON error conversion function


crates/apollo-compiler/src/diagnostic.rs
```diff
@@ -0,0 +1,117 @@
+use crate::FileId;
+use crate::NodeLocation;
+use crate::SourceMap;
+use ariadne::ColorGenerator;
+pub use ariadne::ReportKind;
```
SimonSapin, line 5: If possible I’d prefer ariadne not to be a public dependency so semver-breaking changes don’t need to propagate

SimonSapin (commented):
I like the overall direction! Merging the two source maps into one is probably the right move

crates/apollo-compiler/src/diagnostic.rs
goto-bus-stop: Good point, I wanted to call this Severity rather than Kind anyways, and having our own enum that we can convert to ariadne when ready would allow for that.

crates/apollo-compiler/src/diagnostic.rs
```diff
@@ -0,0 +1,174 @@
...
+use std::fmt;
+use std::io;
+use std::ops::Range;
+use std::sync::Arc;
+use std::sync::OnceLock;
+
+type MappedSpan = (FileId, Range<usize>);
+
+/// Translate a byte-offset location into a char-offset location for use with ariadne.
+fn map_span(sources: &SourceMap, location: NodeLocation) -> Option<MappedSpan> {
```
SimonSapin, line 17: I’ve added these public APIs in https://github.com/apollographql/apollo-rs/pull/758 but they may belong more in this PR (or separately):
```rust
impl GraphQLLocation {
    /// Convert a `NodeLocation` to a line and column number
    pub fn from_node(sources: &SourceMap, location: Option<NodeLocation>) -> Option<Self> {…}
}
impl Node {
    pub fn line_column(&self, sources: &SourceMap) -> Option<GraphQLLocation> {…}
}
impl NodeStr {
    pub fn line_column(&self, sources: &SourceMap) -> Option<GraphQLLocation> {…}
}
```

crates/apollo-compiler/src/diagnostic.rs
```diff
@@ -0,0 +1,174 @@
...
+/// GraphQL source code.
+pub struct DiagnosticReport {
+    sources: SourceMap,
+    report: ariadne::Report<'static, MappedSpan>,
+}
+
+impl DiagnosticReport {
+    /// Returns a builder for creating diagnostic reports.
+    ///
+    /// Provide GraphQL source files and the main location for the diagnostic.
```
SimonSapin, line 105: Maybe mention that `sources` is typically from `document_or_schema.sources.clone()`, and `location` from `most_relevant_name_or_node.location()`?

SimonSapin (commented):
This looks good but I wonder if it’s maybe not the whole story. What does typically usage look like? Does each user call this in `Display` and `Debug` of their own error type that contains a `SourceMap`? Do they also use `Formatter::alternate` to select color v.s. no color?

(Or, BTW, should we replace that with color in `Debug` for panic messages going to CLI, and no color in `Display` because (almost?) every use of `.to_string()` wants `.to_string_no_color()` instead?)

Or is there some nice additional API we could provide on top? Maybe make `DiagnosticList` and `Diagnostic` generic over their `DiagnosticData` type? (And provide a public `ValidationError` type for our own use, maybe or maybe not opaque)

goto-bus-stop:
This now always uses colors and strips them out at the end, because `ariadne` adds/strips colors as soon as you add a label, but here we would normally add the labels first and then decide whether they should be colored when it's time to print.

I added `write` and `fmt` inherent methods that consume `self` so we only need one type. You have to pass in a `SourceMap` in order to render, and you have to use the `fmt` method inside your own `fmt::{Display,Debug}` implementation, but you can't pass in a `SourceMap` into your own `fmt::{Display,Debug}` implementation except for having it available on the object. Soo yes this way would require exposing `Diagnostic` (and `DiagnosticList` probably) for people's custom errors, probably with a trait, which feels heavyweight, or expecting users to stick a `SourceMap` in their errors wherever they are produced.

crates/apollo-compiler/src/diagnostic.rs
```diff
@@ -0,0 +1,304 @@
...
+    pub fn write(&self, w: impl std::io::Write) -> std::io::Result<()> {
+        let mut report = DiagnosticReport::builder(self.sources.clone(), self.error.location());
+        self.error.report(&mut report);
+        report.write(w)
+    }
+}
+
+impl<T: ToDiagnostic> fmt::Display for Diagnostic<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        struct StripColorAdaptor<'a, 'b> {
```
SimonSapin, line 228: This should call `DiagnosticReport::fmt` instead of duplicating its adaptors

crates/apollo-compiler/src/diagnostic.rs
```diff
@@ -0,0 +1,304 @@
...
+}
+
+/// A diagnostic report that can be printed to a CLI with pretty colours and labeled lines of
+/// GraphQL source code.
+///
+/// Custom errors can use this in their `Display` or `Debug` implementations to build a report and
+/// then write it out with [`fmt`].
+///
+/// [`fmt`]: DiagnosticReport::fmt
+pub struct DiagnosticReport {
```
SimonSapin, line 37: Ideas to consider:

* Rename to `CliReport` (since what characterizes Ariadne is monospace text output)
* Replace `SourceMap` parameter in the constructor with `&SourceMap` in `write()` and `fmt`
* Pass color config in the constructor and remove use of `anstream`

crates/apollo-compiler/src/diagnostic.rs
```diff
@@ -0,0 +1,304 @@
...
+                self.f.write_str(s).map_err(|_| io::ErrorKind::Other)?;
+                Ok(buf.len())
+            }
+
+            fn flush(&mut self) -> io::Result<()> {
+                Ok(())
+            }
+        }
+
+        if f.alternate() {
```
SimonSapin, line 262: I came up with this use of `alternate` but i don’t like it that much. What if instead we had `Debug` enable colors (for printed panic messages) and `Display` disable them (for `.to_string()` instead of `.to_string_no_color()`)?

crates/apollo-compiler/src/diagnostic.rs
goto-bus-stop: We need the `SourceMap` to translate locations when you add a label unfortunately.

crates/apollo-compiler/src/diagnostic.rs
```diff
@@ -0,0 +1,344 @@
...
+        let mut report =
+            CliReport::builder(self.sources.clone(), self.error.location()).with_color(color);
+        self.error.report(&mut report);
+        report
+    }
+
+    /// Pretty-print the diagnostic to a [`Write`].
+    ///
+    /// [`Write`]: std::io::Write
+    pub fn write(&self, color: bool, w: impl std::io::Write) -> std::io::Result<()> {
```
SimonSapin, line 288: At call sites it’s not obvious what this boolean means. What do you think of `enum Color { None, StderrIsTerminal }` ?

crates/apollo-compiler/src/diagnostic.rs
```diff
@@ -0,0 +1,344 @@
...
+    colors: ColorGenerator,
+    report: ariadne::ReportBuilder<'static, MappedSpan>,
+}
+
+impl CliReport {
+    /// Returns a builder for creating diagnostic reports.
+    ///
+    /// Provide GraphQL source files and the main location for the diagnostic.
+    /// Source files can be obtained from [`Schema::sources`] or [`ExecutableDocument::sources`].
+    pub fn builder(sources: SourceMap, location: Option<NodeLocation>) -> Self {
```
SimonSapin, line 131: A report typically only exists transiently to call its `write` or `fmt` method, right? Could it own a `&'_ SourceMap` instead of `SourceMap`? (Or maybe I shouldn’t worry about micro-things like this…)

Similarly for `Diagnostic<T>`

crates/apollo-compiler/src/diagnostic.rs
```diff
@@ -0,0 +1,344 @@
...
+            }
+            Some(Box::new(NoSourceFile))
+        }
+    }
+}
+
+/// A pretty-printable diagnostic.
+pub struct Diagnostic<T> {
+    pub sources: SourceMap,
+    pub location: Option<NodeLocation>,
```
SimonSapin, line 237: Is this field useful? Methods call `self.error.location()`

SimonSapin (approved):

SimonSapin:
I made some tweaks, added a changelog, and merged to stop myself from making more tweaks :sweat_smile:

Issue #742, opened 2023-11-14
Apply recursion limit in validation


goto-bus-stop:
The parser has a recursion limit to prevent stack overflows on deeply nested selection sets. Validation has recursion breaks for things that may circularly reference each other. But if you build a chain of thousands of named things that reference each other, you may theoretically be able to cause a stack overflow before the recursion break triggers (before it references back to the original definition).

qwerdenkerXD:
[...]

PR #748, opened 2023-11-17
fix(compiler): apply limits to recursive functions in validation

goto-bus-stop:
Fixes https://github.com/apollographql/apollo-rs/issues/742

The validation code already had protections for cycles. However if you provided a very long chain of types or fragments referencing each other, it could chase each reference and still cause a stack overflow. This does pretty much require malicious input, but the failure mode is quite bad, so it's worth protecting against.

Self-referential directives and input objects now also show the chain that was followed in diagnostics. The directives one can be a bit incomplete because there may be a different kind of node in between (eg there might be a directive applied to an enum used inside an input type used as an argument to the same directive... and only directive names are tracked in the chain.) I think we can improve that with other diagnostics work soon, this is already an improvement.


crates/apollo-compiler/README.md
```diff
@@ -87,14 +88,15 @@ let schema = Schema::parse(input, "document.graphql");
 
 /// In case of validation errors, the panic message will be nicely formatted
 /// to point at relevant parts of the source file(s)
-schema.validate().unwrap();
+schema.validate(ValidationOptions::default()).unwrap();
```
SimonSapin, line 91: This makes the public API feel quite heavy. Do we expect real use cases for changing the recursion limit from its default? As a precedent `serde_json::Deserializer`, doesn’t offer any way to change the amount of its own recursion limit.

goto-bus-stop: Yeah I see. I think having a diagnostic limit is potentially more useful than this (in a server it would be totally reasonable to abort after a handful of validation errors). But maybe there's a better way to provide that option. made it hardcoded now

crates/apollo-compiler/tests/validation/fragments.rs
```diff
@@ -0,0 +1,57 @@
+#[test]
+fn long_fragment_chains_do_not_overflow_stack() {
+    // Build a query that applies thousands of fragments
+    // Validating it would take a lot of recursion and blow the stack
```
SimonSapin, line 4: I take it you’ve manually reproduced the stack overflow when the limit is set too high?

Please also add a similar test that goes right under the limit and succeeds.

SimonSapin (approved):
Looks good! Let’s merge this with an additional test (see inline comment) and a changelog entry

goto-bus-stop:
Also added tests for input objects and directives. Directives in particular didn't cause a stack overflow even with thousands of definitions, but they did take over 60 seconds to validate, so the limit also helps with that.

I lowered the limit to 200 because fragments could still overflow with 500 (the cycles detection works with 500, but the recursion in `validate_fragment_definition` is more expensive as it cycles through a few different and bigger methods).
Thinking about the serde_json limit, 200 is still a lot!

crates/apollo-compiler/tests/validation/recursion.rs
```diff
@@ -0,0 +1,127 @@
...
+    // Build a query that applies 1K directives
+    // Validating it would take a lot of recursion and a lot of time
+    let schema = build_directive_chain(1_000);
+
+    let partial = apollo_compiler::Schema::parse_and_validate(schema, "directives.graphql")
+        .expect_err("must have recursion errors");
+
+    // The final 199 directives do not cause recursion errors because the chain is less than 200
+    // directives deep.
+    assert_eq!(partial.errors.len(), 801);
```
SimonSapin, line 146: Oh, do we restart following the whole chain for every step in it? Quadratic time is why it took 60 seconds… Maybe this limit should be much lower still

goto-bus-stop: We dooo. Fragment definitions are validated when they are used, but for directives and input objects each definition is validated.

Long fragment chains do happen in the wild, but long input object chains and directive chains might not, so we could have a low limit for those. The risk is lower for directives and input objects because they are in the schema and not likely to be provided over the internet by a malicious user.

SimonSapin (approved):

goto-bus-stop:
Lowered the limits more after trying many real-world inputs. I basically found no practical example of deeply nested input objects (for the purposes of schema validation, only non-null fields are chased, which are very rare) or nested directive applications (found no example at all, which makes sense to me, i've never seen someone take complicated arguments in directives aside from custom scalars). 32 is still an order of magnitude above what I've seen :)

For fragments I don't have great data, my suspicion is that fragment chains can be long in practice, but like...5 or 6 would already be quite long. I put that limit at 100 so it won't trigger even on extreme use cases but not grind to a halt.

Issue #709, opened 2023-10-19
It’s too easy to use a schema or document with errors


SimonSapin:
This is breaking change that we should either make before 1.0 leaves beta, or decide not to make.

## Status quo

Typical intended usage of apollo-compiler 1.0.0-beta.4 looks like this:

```rust
let schema = apollo_compiler::Schema::parse(schema_input, "schema.graphql");
schema.validate()?;
let doc = apollo_compiler::ExecutableDocument::parse(&schema, input, "doc.graphql");
doc.validate(&schema);
do_something_with(&schema, &doc)
```

However it’s very easy to forget validation:

```rust
let schema = apollo_compiler::Schema::parse(schema_input, "schema.graphql");
let doc = apollo_compiler::ExecutableDocument::parse(&schema, input, "doc.graphql");
do_something_with(&schema, &doc)
```

… and get a schema and document that look fine and contain some of the expected things, but in fact may have error and silently be missing other things. Specifically, two categories of errors can be encountered before running "full validation":

* Parse errors (invalid grammar or limit exceeded)
* So-called "build errors" during conversion to a high-level data structure. For example: inserting a definition in a name-indexed map can fail if the name was already used.

In either case an error is recorded and returned by `.validate()`. But in the second usage example they’ll be ignored and may cause unexpected results later, when they’re harder to debug.

Full validation as well is a property that may be useful to encode in the Rust type system. Maybe even `ExecutableDocument` should only be created with a fully valid schema?

## Proposal

Change `parse` methods to return a `Result` in order to force callers to handle the error case (even if with `unwrap`).

This result should not always be tied to that of full validation. Federation for example needs to parse subgraph schemas that are invalid per spec because their root `Query` type may be implicit. Federation code fills in the blanks by programmatically mutating a `Schema` before calling validation.

If each different state is a different Rust type we may need to pick names for up to 4 types just for schemas:

```rust
Schema::parse(…) -> Result<Schema, SchemaWithParseOrBuildError>
Schema::validate(self) -> Result<ValidSchema, InvalidSchema>
```

… and similarly for executable documents.

Unless `parse` runs full validation but only when it encounters a build or parse error? In which case both results could share a `InvalidSchema` error type.

SimonSapin:
Sketch of a possible API:

```rust
/// Immutable, validated. Execution would require this.
///
/// FIXME: keep this name? "Document" is made implicit
struct Executable {…}

/// Mutable, may or may not be valid
///
[...]

Issue #751, opened 2023-11-17
Consider re-introducing lints to apollo-rs


SimonSapin:
In apollo-compiler 0.11 validation returns a `Vec` that can contain:

* Any number of error-level diagnostics
* A warning for enum value definitions with non-uppercase names
* An "advice" for custom scalar definitions without an `@specifiedBy` directive

To find out if a document is valid, callers need to iterate the vec and find if any of the diagnostic return true for `.is_error()`.

In apollo-compiler 1.0 (in beta as of this writing) the API changes so that the library does that filtering and returns `Result::Err` for invalid documents. However as discussed starting at https://github.com/apollographql/apollo-rs/issues/709#issuecomment-1814923975 returning potential warnings and advice for a valid document makes the API awkward. Since they arguably don’t belong in `validate()` in the first place, they could be moved to a separate `lint()` method.

But with only two rules, this linter would not be very valuable and we don’t expect many users to go out of their way to call it explicitly.

An upcoming will therefore remove these two non-error rules from apollo-compiler entirely. This issue keeps a record of this regression of functionality. These two rule could be restored one day if we want to seriously work on having a GraphQL linter as part of apollo-rs.

goto-bus-stop:
Here is something that would be a useful "lint" while it's not a spec validation error: https://github.com/graphql/graphql-spec/pull/1053

SimonSapin:
Isn't the linked PR changing the spec to make it a validation error? It would only be a warning until the spec change is adopted (assuming it is)

goto-bus-stop:
yes! i meant "while" as in, "as long as" it's not an error

PR #752, opened 2023-11-18
API refactor to make it harder to ignore errors

SimonSapin:
Fixes https://github.com/apollographql/apollo-rs/issues/709

## BREAKING

- `ast::Document`, `Schema`, and `ExecutableDocument` not longer contain potential errors that users need to check separately.
- Instead, various constructors and methods now return a `Result`, with the `Err` case containing both both errors and a maybe-incomplete value.
- Change `validate` methods of `Schema` and `ExecutableDocument` to take ownership of `self`. On success they return the schema or document (unmodified) wrapped in a `Valid<_>` marker type, which is **immutable**.
- Change `ExecutableDocument` to require a `&Valid<Schema>` instead of `&Schema`, forcing callers to either run validation or opt out explicitly with `Valid::assert_valid`.
- Make `parse_mixed` and `to_mixed` validate both the schema and document. Rename them with a `_validate` suffix.
- Corresponding changes to all of the above in `Parser` method signatures
- Remove `ast::Document::check_parse_errors`: parse errors are now encoded in the return value of `parse`.
- Remove `ast::Document::to_schema_builder`. Use `SchemaBuilder::add_ast` instead.
- Move items from the crate top-level to `apollo_compiler::validation`:
  * `Diagnostic`
  * `DiagnosticList`
  * `FileId`
  * `NodeLocation`
- Move items from the crate top-level to `apollo_compiler::execution`:
  * `GraphQLError`
  * `GraphQLLocation`
- Remove warning-level and advice-level diagnostics. See https://github.com/apollographql/apollo-rs/issues/751

Highlight of signature changes:

```diff
+struct Valid<T>(T); // Implements `Deref` and `AsRef` but not `DerefMut` or `AsMut`
+
+struct WithErrors<T> {
+    partial: T, // Errors may cause components to be missing
+    errors: DiagnosticList,
+}

-pub fn parse_mixed(…) -> (Schema, ExecutableDocument)
+pub fn parse_mixed_validate(…)
+    -> Result<(Valid<Schema>, Valid<ExecutableDocument>), DiagnosticList>

  impl ast::Document {
-    pub fn parse(…) -> Self
+    pub fn parse(…) -> Result<Self, WithErrors<Self>>

-    pub fn to_schema(&self) -> Schema
+    pub fn to_schema(&self) -> Result<Schema, WithErrors<Schema>>

-    pub fn to_executable(&self) -> ExecutableDocument
+    pub fn to_executable(&self) -> Result<ExecutableDocument, WithErrors<ExecutableDocument>>

-    pub fn to_mixed(&self) -> (Schema, ExecutableDocument)
+    pub fn to_mixed_validate(
+        &self,
+    ) -> Result<(Valid<Schema>, Valid<ExecutableDocument>), DiagnosticList>
  }

  impl Schema {
-    pub fn parse(…) -> Self
-    pub fn validate(&self) -> Result<DiagnosticList, DiagnosticList>

+    pub fn parse_and_validate(…) -> Result<Valid<Self>, WithErrors<Self>>
+    pub fn parse(…) -> Result<Self, WithErrors<Self>>
+    pub fn validate(self) -> Result<Valid<Self>, WithErrors<Self>>
  }

  impl SchemaBuilder {
-    pub fn build(self) -> Schema
+    pub fn build(self) -> Result<Schema, WithErrors<Schema>>
  }

  impl ExecutableDocument {
-    pub fn parse(schema: &Schema, …) -> Self
-    pub fn validate(&self, schema: &Schema) -> Result<(), DiagnosticList>

+    pub fn parse_and_validate(schema: &Valid<Schema>, …) -> Result<Valid<Self>, WithErrors<Self>>
+    pub fn parse(schema: &Valid<Schema>, …) -> Result<Self, WithErrors<Self>>
+    pub fn validate(self, schema: &Valid<Schema>) -> Result<Valid<Self>, WithErrors<Self>>
  }
```

## Features

- **Add `parse_and_validate` constructors for `Schema` and `ExecutableDocument`:** when mutating isn’t needed after parsing, this returns an immutable `Valid<_>` value in one step.


SimonSapin:
TODO: add this escape hatch:

```rust
impl<T> Valid<T> {
    pub fn assert_valid(value: T) -> Self {
        Self(value)
    }
}
```

crates/apollo-compiler/src/validation/mod.rs
```diff
@@ -38,20 +38,32 @@ pub(crate) use validation_db::{ValidationDatabase, ValidationStorage};
...
 pub struct Valid<T>(pub(crate) T);
 
 impl<T> Valid<T> {
+    /// Construct a `Valid` document without actually running validation.
+    ///
+    /// The caller takes responsibility to ascertain that
+    /// the document is known through some other means to be valid.
+    /// For example, if it was loaded from some external storag
+    /// where it was only stored after validation.
+    pub fn assert_valid(document: T) -> Self {
```
goto-bus-stop, line 63: With `assert` i kind of expect it to work like `assert!`, what do you think about a different name like `assume_valid()`?

SimonSapin: I probably came up with this based on https://doc.rust-lang.org/core/panic/struct.AssertUnwindSafe.html, but I think the association with `assert!` and "can panic" is still better avoided. I’ll rename to `assume_valid`

goto-bus-stop (approved):

PR #754, opened 2023-11-20
Remove encoder, use compiler in smith instead

SimonSapin:
apollo-compiler now supports serialization for its document representations, which replaces the functionality of apollo-encoder entirely.


goto-bus-stop (approved):
lets gooo

Issue #691, opened 2023-10-12
Port PR #674 to 1.0: improve validation error summary


SimonSapin:
PR https://github.com/apollographql/apollo-rs/pull/674 was reverted in https://github.com/apollographql/apollo-rs/pull/690 (but kept in the [`compiler-v0.11`](https://github.com/apollographql/apollo-rs/commits/compiler-v0.11) branch) in order to help compiler v1.0 to be merged into main.

Merging these improvements into 1.0 with git has proven difficult, so we’ll need to port them manually.

PR #757, opened 2023-11-22
feat(compiler): add schema coordinates

goto-bus-stop:
Implements parsing and printing for the [Schema Coordinates RFC](https://github.com/graphql/graphql-wg/blob/main/rfcs/SchemaCoordinates.md).

Here, `SchemaCoordinate` is an enum of all the different kinds of coordinates. Each kind of coordinate is also its own type.

```rust
use apollo_compiler::name;
use apollo_compiler::coordinate::{SchemaCoordinate, FieldArgumentCoordinate};

let coord: SchemaCoordinate = "Type.field(argument:)".parse().unwrap();
assert_eq!(coord, SchemaCoordinate::FieldArgument(
    FieldArgumentCoordinate {
        ty: name!("Type"),
        field: name!("field"),
        argument: name!("argument"),
    },
));
```

```rust
use apollo_compiler::coord;

assert_eq!(coord!(@directive).to_string(), "@directive");
assert_eq!(coord!(@directive(arg:)).to_string(), "@directive(arg:)");
assert_eq!(coord!(Type).to_string(), "Type");
assert_eq!(coord!(Type.field).to_string(), "Type.field");
assert_eq!(coord!(Type.field(arg:)).to_string(), "Type.field(arg:)");
```

The goal of this API is:
 - easily tracking the current coordinate while walking a tree (eg with the `.argument()` method). Each kind of coordinate having its own `impl` is handy for this
 - look up a typed coordinate in the schema
   - not yet implemented but imagine an `fn lookup(&Schema, FieldArgumentCoordinate) -> Option<Node<InputValueDefinition>>`, so the return value can be typed. I'm not 100% convinced of how important this is as you may often have just a `SchemaCoordinate`, which can return multiple types.
   - Backed out for now, we can do it later, typed lookup is certainly possible.
 - Printing for use in error messages (https://github.com/apollographql/apollo-rs/issues/691)

Some other options:
 - `TypeCoordinate` could just be an alias of `NamedType` 🤪 
 - Just have a `SchemaCoordinate` enum without all the specific types--I like this less for how it's used in validation (different types with simple methods for pushing/popping are ideal)
 - A more flat text representation, so not having 3 separate refcounted strings like here. Just representing a schema coordinate as its source text is actually fairly efficient (like literally storing the string `Type.field(arg:)`). A schema coordinate could then be pointer-sized, here it's 3 pointers + a discriminant. A downside is that you need to allocate when you append a field or argument or something.

### DIFF
--- a/crates/apollo-compiler/CHANGELOG.md
+++ b/crates/apollo-compiler/CHANGELOG.md
@@ -45,7 +45,6 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   Schema coordinates are a compact, human-readable way to uniquely point to an item defined in a schema.
   - `string.parse::<SchemaCoordinate>()` parses a coordinate from a string.
   - Coordinates have a `Display` impl that writes them out with schema coordinate syntax.
-  - `schema.lookup(&coord)` looks up the item pointed to by a coordinate in a schema.
   - The `coord!()` macro creates a static coordinate at compile time from spec syntax.
 
 ## Fixes
--- a/crates/apollo-compiler/CHANGELOG.md
+++ b/crates/apollo-compiler/CHANGELOG.md
@@ -41,6 +41,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - **Add `NodeStr::from(Name)` - [goto-bus-stop], [pull/773]**
 - **Convenience accessors for `ast::Selection` enum - [SimonSapin], [pull/777]**
   `as_field`, `as_inline_fragment`, and `as_fragment_spread`; all returning `Option<&_>`.
+- **Add schema coordinates - [goto-bus-stop], [pull/757]:**
+  Schema coordinates are a compact, human-readable way to uniquely point to an item defined in a schema.
+  - `string.parse::<SchemaCoordinate>()` parses a coordinate from a string.
+  - Coordinates have a `Display` impl that writes them out with schema coordinate syntax.
+  - `schema.lookup(&coord)` looks up the item pointed to by a coordinate in a schema.
+  - The `coord!()` macro creates a static coordinate at compile time from spec syntax.
 
 ## Fixes
 - **Fix serializing single-line strings with leading whitespace - [goto-bus-stop], [pull/774]**
--- a/crates/apollo-compiler/src/coordinate.rs
+++ b/crates/apollo-compiler/src/coordinate.rs
@@ -17,23 +17,33 @@ use std::str::FromStr;
 #[macro_export]
 macro_rules! coord {
     ( @ $name:ident ) => {
-        $crate::coordinate::DirectiveCoordinate($crate::name!($name))
+        $crate::coordinate::DirectiveCoordinate {
+            directive: $crate::name!($name),
+        }
     };
     ( @ $name:ident ( $arg:ident : ) ) => {
-        $crate::coordinate::DirectiveArgumentCoordinate($crate::name!($name), $crate::name!($arg))
+        $crate::coordinate::DirectiveArgumentCoordinate {
+            directive: $crate::name!($name),
+            argument: $crate::name!($arg),
+        }
     };
     ( $name:ident ) => {
-        $crate::coordinate::TypeCoordinate($crate::name!($name))
+        $crate::coordinate::TypeCoordinate {
+            ty: $crate::name!($name),
+        }
     };
     ( $name:ident . $field:ident ) => {
-        $crate::coordinate::FieldCoordinate($crate::name!($name), $crate::name!($field))
+        $crate::coordinate::FieldCoordinate {
+            ty: $crate::name!($name),
+            field: $crate::name!($field),
+        }
     };
     ( $name:ident . $field:ident ( $arg:ident : ) ) => {
-        $crate::coordinate::FieldArgumentCoordinate(
-            $crate::name!($name),
-            $crate::name!($field),
-            $crate::name!($arg),
-        )
+        $crate::coordinate::FieldArgumentCoordinate {
+            ty: $crate::name!($name),
+            field: $crate::name!($field),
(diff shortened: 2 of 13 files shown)


crates/apollo-compiler/src/coordinate.rs
```diff
@@ -0,0 +1,286 @@
...
+/// ```
+/// use apollo_compiler::name;
+/// use apollo_compiler::coordinate::TypeCoordinate;
+///
+/// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+/// assert_eq!(TypeCoordinate(name!("Type")).to_string(), "Type");
+/// # Ok(()) }
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct TypeCoordinate(pub Name);
```
SimonSapin, line 52: I’d prefer named fields for all of these structs, especially for those with 2+ fields of the same type (but let’s keep them consistent with each other)

crates/apollo-compiler/src/coordinate.rs
```diff
@@ -0,0 +1,286 @@
...
+}
+
+/// A schema coordinate targeting a type definition: `Type`.
+///
+/// # Example
+/// ```
+/// use apollo_compiler::name;
+/// use apollo_compiler::coordinate::TypeCoordinate;
+///
+/// # fn main() -> Result<(), Box<dyn std::error::Error>> {
```
SimonSapin, line 47: The `fn main` boilerplate isn’t necessary unless using the `?` operator, is it?

SimonSapin (commented):
Are there other kinds of schema coordinates that diagnostics could use too? Like enum value definition, or `implements` of an interface by an object type, etc.

I tend to prefer structured data over stringly-typed, but if the main purpose is to be formatted into a string anyway :shrug: I don’t think 3× refcount traffic is very significant compared to allocating a new string. Do you think `size_of::<SchemaCoordinate>()` is something to worry about? We may care less (to some extent) about performance of reporting errors than about validation performance of valid schemas/documents.

@sachindshinde I think I already asked but sorry I don’t remember the conclusion :sweat_smile: Is there overlap between this and federation-next’s referencers/positions? Any change you’d suggest here?

goto-bus-stop:
Oh hmm, a `FieldCoordinate` can also refer to an enum value. The RFC actually calls it a type attribute rather than a field: https://github.com/graphql/graphql-wg/blob/main/rfcs/SchemaCoordinates.md#typeattribute

So you can have `Query.products` referring to a field and `ProductType.book` referring to an enum value. that makes the typed lookup by coordinate idea a bit less interesting.

goto-bus-stop:
I think for pointing out a specific `implements`, or pointing out a directive application, we would use both a coordinate + a string (or another coordinate) for the more specific thing in the error message. It does mean that the coordinates for error locations would often be a parent node of where the error actually occurred, but it's not gonna be super far away...

For example, if you do something like
```graphql
type Example {
  name @deprecated(reason: false)
}
```
you could get an error with its location at coordinate `Example.name`, and the message formatting using two coordinates:
```
Expected type String, but Boolean was provided to `@deprecated(reason:)` on `Example.name`
```
(would need some copy workshopping to make it read well)

goto-bus-stop:
Something else you can't point to is a schema definition, so we can't *require* a coordinate for diagnostics on those

fuzz/fuzz_targets/coordinate.rs
```diff
@@ -0,0 +1,20 @@
+#![no_main]
+use apollo_compiler::coordinate::SchemaCoordinate;
+use libfuzzer_sys::arbitrary::Arbitrary;
+use libfuzzer_sys::arbitrary::Unstructured;
+use libfuzzer_sys::fuzz_target;
+use log::debug;
+
+fuzz_target!(|data: &[u8]| {
+    let s = String::arbitrary(&mut Unstructured::new(data)).unwrap();
+
```
SimonSapin, line 10: libfuzzer can call `Arbitrary` itself to convert the argument to the expected type:

```suggestion
fuzz_target!(|s: &str| {
```

crates/apollo-compiler/src/coordinate.rs
```diff
@@ -0,0 +1,684 @@
...
+        }
+    };
+}
+
+mod sealed {
+    pub trait Sealed {}
+}
+
+/// Provides [`schema.lookup(&coord)`][Schema::lookup] for any schema coordinate.
+pub trait SchemaLookup: sealed::Sealed {
```
SimonSapin, line 71: This trait only exists to support the `Schema::lookup` method, right? I feel it and the method don’t bring a lot of value, compared to separate `lookup` inherent methods on the various coordinate types. Anything that could be generic can use the enum instead.

crates/apollo-compiler/CHANGELOG.md
```diff
@@ -20,6 +20,16 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+
+- **Add schema coordinates - [goto-bus-stop], [pull/757]:**
+  Schema coordinates are a compact, human-readable way to uniquely point to an item defined in a schema.
+
+  * `string.parse::<SchemaCoordinate>()` parses a coordinate from a string.
+  * Coordinates have a `Display` impl that writes them out with schema coordinate syntax.
+  * `schema.lookup(&coord)` looks up the item pointed to by a coordinate in a schema.
+  * The `coord!()` macro creates a static coordinate at compile time from spec syntax.
+
+# [1.0.0-beta.9](https://crates.io/crates/apollo-compiler/1.0.0-beta.9) - 2023-11-30
```
SimonSapin, line 32: Looks like wrong rebase or merge

goto-bus-stop:
As discussed in checkin, we keep lookup out for now. Replaced `coordinate: String` uses in diagnostics with either a specific coordinate type or `SchemaCoordinate`.

SimonSapin (approved):

PR #758, opened 2023-11-23
Add schema introspection support and execution-related APIs

SimonSapin:
Changelog:

- Add data structure in `apollo_compiler::execution` for a GraphQL response, its data, and errors.
  All (de)serializable with `serde`.
- Add [`coerce_variable_values()`] in that same module.
- Add `apollo_compiler::execution::SchemaIntrospection`
  providing full execution for the [schema introspection] parts of an operation
  and separating the rest to be executed separately.
  In order to support all kinds of introspection queries this actually includes
  a full execution engine where users provide objects with resolvable fields.
  At this time this engine is not exposed in the public API.
  If you’re interested in it [let us know] about your use case!
- Add `ExecutableDocument::insert_operation` convenience method.

[`coerce_variable_values()`]: https://spec.graphql.org/October2021/#sec-Coercing-Variable-Values
[schema introspection]: https://spec.graphql.org/October2021/#sec-Schema-Introspection
[let us know]: https://github.com/apollographql/apollo-rs/issues/new?assignees=&labels=triage&projects=&template=feature_request.md

----

Replaces and closes https://github.com/apollographql/apollo-rs/pull/645: 
* The execution engine and introspection resolvers are copied from there. `async` was removed (since unused by introspection, the only user)
* The public API was reworked to be part of apollo-compiler
* The logic splitting an operation into introspection and non-introspection parts was rewritten two produce valid documents. This logic needs a lot more testing.


SimonSapin:
I pushed a rework of error handling. `introspection_split.rs` still needs work to reject non-top-level introspection, but the rest of the PR can be considered ready for review.

crates/apollo-compiler/tests/introspection.rs
```diff
@@ -0,0 +1,136 @@
+use apollo_compiler::execution::coerce_variable_values;
+use apollo_compiler::execution::JsonMap;
+use apollo_compiler::execution::Response;
+use apollo_compiler::execution::SchemaIntrospectionQuery;
+use apollo_compiler::ExecutableDocument;
+use apollo_compiler::Schema;
+use expect_test::expect;
+use expect_test::expect_file;
+
+#[test]
```
lrlna, line 10: sometime later we should port over the graphql-js introspection & execution tests

lrlna (approved):
We should probably merge this in without making too many more changes, and then incrementally improve after we get a chance to try it out in the router (like you said before, Simon).

Ideally we'd try integrating it in the router, see if our users have any feedback, implement tests from graphql-js implementations and make this API a 1.0 with the compiler. But yea, regardless, let's get this in.

crates/apollo-compiler/src/execution/engine.rs
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    let mut response_map = JsonMap::with_capacity(grouped_field_set.len());
+    for (&response_key, fields) in &grouped_field_set {
+        // Indexing should not panic: `collect_fields` only creates a `Vec` to push to it
+        let field_name = &fields[0].name;
+        let Ok(field_def) = schema.type_field(object_type_name, field_name) else {
+            // TODO: Return a `validation_bug`` field error here?
+            // The spec specifically has a “If fieldType is defined” condition,
+            // but it being undefined would make the request invalid, right?
```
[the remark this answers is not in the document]
SimonSapin: I don’t think so. This is about [field errors](https://spec.graphql.org/October2021/#sec-Handling-Field-Errors) that happen during execution. An undefined field makes an operation invalid, so execution shouldn’t even start.

This comment is because I was surprised to see the spec use "if" there, whereas it uses "assert" in other places ([example](https://spec.graphql.org/October2021/#sel-IANNFHCDABABGqD)) for things expected of valid documents.

SimonSapin:
Alright, this is now fully ready for review! Please also take a look at `introspection_split.rs` and its tests.

crates/apollo-compiler/tests/introspection_split.rs
```diff
@@ -0,0 +1,382 @@
...
+}
+
+#[test]
+fn test_mutation() {
+    let doc = r#"
+        mutation TheOperation { __schema { description } }
+    "#;
+    let expected = expect![[r#"
+        GraphQLError {
+            message: "Schema introspection field __schema is only supported at the root of a query",
```
lrlna, line 122: to triple check, we only split on query _operations_? sometimes people use the word "query" to also mean mutations and subscriptions, which would make this error inaccurate.

crates/apollo-compiler/tests/introspection_split.rs
```diff
@@ -0,0 +1,382 @@
...
+    assert_split(doc, expected);
+}
+
+#[test]
+fn test_nested_in_other_operation() {
+    let doc = r#"
+        query TheOperation { value }
+        query OtherOperation { value next { __schema { description } } }
+    "#;
+    let expected = expect!["None"];
```
lrlna, line 163: This is not meant to produce an error like with the cases above?

crates/apollo-compiler/src/execution/engine.rs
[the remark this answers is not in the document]
lrlna: gotcha gotcha

crates/apollo-compiler/tests/introspection_split.rs
SimonSapin: Yes, “query” is used here to mean `OperationType::Query`. Conversely, `__schema` and `__type` are *not* supported:

* Anywhere in a mutation
* Anywhere in a subscription
* In a nested field (not at the root) of a query

crates/apollo-compiler/tests/introspection_split.rs
SimonSapin: No. You’d get an "unsupported" error if you select the other operation, but the document (however weird) is valid per spec.

crates/apollo-compiler/tests/introspection_split.rs
SimonSapin: I rephrased the error messages (see latest commit) to avoid this ambiguity.

PR #768, opened 2023-12-01
Add `Valid::assume_valid_ref`

SimonSapin:


SimonSapin:
federation-next needs this to upgrade to beta 8+

goto-bus-stop (approved):
Makes sense. though would it make sense for fed-next to clone the document anyways, if it has to do modifications?

goto-bus-stop:
Ahhh right I get it, thanks!

PR #772, opened 2023-12-04
__typename only exists on objects, intrefaces, and unions

SimonSapin:
`Schema::type_field` incorrectly returned the definition of __typename for other kinds of types.

This turned out not to be observable from parser input since some other error like "field selection of scalar type `String` must not have subselections" triggers before `type_field` would be called. But might as well make `type_field` more correct now that the patch is written.


goto-bus-stop (approved):
SGTM

PR #774, opened 2023-12-07
fix(parser,compiler): fix string serialization bugs found by fuzzing

goto-bus-stop:
Two minor fixes:
- Leading whitespace on a single-line string can not be serialized as a block string
- `\\"""` in a block string should be lexed as `\` followed by `\"""`, previously the first backslash escaped the second backslash and caused the `"""` to be interpreted as the end of the string.


lrlna (approved):
hell ya!

Issue #538, opened 2023-04-26
represent parameterised grammar (const values)


lrlna:
We don't correctly diagnose incorrectly parameterised values on the parser side of things. For example, `Directive` in a `VariableDefinition` must not have a `Variable` as its `Argument`'s `Value`. [spec](https://spec.graphql.org/October2021/#sec-Grammar-Notation.Parameterized-Grammar-Productions). We should produce this diagnostic in the parser.

Perhaps the most straightforward way to do this is to have two versions of the the `Value` enum - one with `Variable` and the other one without, and then propagate that up to `Arguments`, `Argument`, `Directives` and `Directive`. 

Would be useful to have that enum variant type representation in rust here, but alas not possible in the language yet :/

SimonSapin:
Would it be worth making const-ness a type parameter?

```rust
enum Value<C: Constness> {
    Null,
    Boolean(bool),
    // …
    Variable(C::Variable),
}

trait Constness {
    type Variable;
}

struct NotConst;
impl Constness for NotConst {
    type Variable = Name;
}

struct Const;
impl ConstNess for Const {
    type Variable = Never;
}
enum Never {}
```

Or without the indirection:

```rust
enum MaybeConstValue<Variable> {
    Null,
    Boolean(bool),
    // …
    Variable(Variable),
}

type Value = MaybeConstValue<Name>;
type ConstValue = MaybeConstValue<Never>;
enum Never {}
```

goto-bus-stop:
With the `Never` pattern you don't need to match on `ConstValue::Variable(_)` right?

The constness has to propagate through a few levels (Directives -> Arguments -> Value), and users will have to write functions that are generic over constness. So I think having a clearly named `Constness` trait that they can use as a bound is a good idea.

If you have a function that operates on a not-const value you can't put a const value in. I guess we would want a From impl or maybe even a safe method on maybe-const types that does a transmute internally from const to not-const as we know the Variable member will never be active.

SimonSapin:
I tried it, with `Never` you’d need a match arm like this:

```rust
ConstValue::Variable(never) => match never {}
```

… which type-checks for any expected type. "All" of the zero arms of the inner `match` for each variant of `Never` have a value of the expected type.

It’s only with [the unstable `!` type](https://doc.rust-lang.org/std/primitive.never.html) that the `Variable` match arm could be skipped entirely.

SimonSapin:
[...]

PR #777, opened 2023-12-10
fix(parser): emit syntax errors for variables in const values

SimonSapin:
https://github.com/apollographql/apollo-rs/issues/538 tracks making such invalid values impossible to represent in Rust data structures, so that users matching on the enum values don’t need to handle impossible cases. But that isn’t needed to make the parser emit syntax errors.


crates/apollo-parser/test_data/parser/err/0052_cont_value.graphql
```diff
@@ -0,0 +1,14 @@
+query(
```
goto-bus-stop, line 1: file name says `cont` instead of `const`

crates/apollo-parser/src/parser/grammar/enum_.rs
```diff
@@ -26,7 +25,7 @@ pub(crate) fn enum_type_definition(p: &mut Parser) {
     }
 
     if let Some(T![@]) = p.peek() {
-        directive::directives(p);
+        directive::directives(p, Constness::NotConst);
```
goto-bus-stop, line 28: should be const on enum type definition
```suggestion
        directive::directives(p, Constness::Const);
```

crates/apollo-parser/src/parser/grammar/value.rs
```diff
@@ -3,21 +3,38 @@ use crate::{
...
+    Const,
+    NotConst,
+}
+
 /// See: https://spec.graphql.org/October2021/#Value
 ///
-/// *Value*
-///     Variable
+/// *Value[Const]*
+///     [if not Cont] Variable
```
goto-bus-stop, line 15:
```suggestion
///     [if not Const] Variable
```

goto-bus-stop (changes requested):
Great 😁  I think one of them is wrong, otherwise LGTM

crates/apollo-parser/src/parser/grammar/enum_.rs
SimonSapin: Good catch!

goto-bus-stop (approved):

PR #782, opened 2023-12-13
feat(xtask): add xtask lint

goto-bus-stop:
Runs fmt and clippy with strict settings so when linting passes locally,
it will also pass on CI.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -30,12 +30,31 @@ struct Xtask {
 pub enum Command {
     /// Perform code generation for the parser
     Codegen(codegen::Codegen),
+    /// Run clippy
+    Lint,
+}
+
+fn run_clippy() -> Result<()> {
+    let sh = Shell::new()?;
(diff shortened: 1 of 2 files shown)


lrlna (approved):
let's do it!

xtask/src/main.rs
```diff
@@ -30,12 +30,31 @@ struct Xtask {
...
+
+    cmd!(sh, "cargo fmt --all -- --check").run()?;
+
+    cmd!(
+        sh,
+        "cargo clippy --all-targets --all-features -- -D warnings"
+    )
+    .run()?;
+
+    cmd!(sh, "cargo clippy --benches").run()?;
```
SimonSapin, line 48: Isn’t this already included in `--all-targets` above?

goto-bus-stop: yes thanks! https://github.com/apollographql/apollo-rs/pull/783

PR #794, opened 2024-01-04
chore: Apply Clippy suggestions from Rust 1.75

goto-bus-stop:


crates/apollo-compiler/src/execution/input_coercion.rs
```diff
@@ -448,7 +448,7 @@ fn coerce_argument_value(
                     ));
                     return Err(PropagateNull);
                 }
-                // `map` converts `&(k, v)` to `(&k, &v)`
+                #[allow(clippy::map_identity)] // `map` converts `&(k, v)` to `(&k, &v)`
```
SimonSapin, line 451: This false positive is already fixed upstream: https://github.com/rust-lang/rust-clippy/issues/11764 so this `allow` will be unnecessary eventually but it’s fine to have it in the mean time

SimonSapin (approved):

PR #798, opened 2024-01-08
fix(compiler): InputValueDefinition::is_required() returns false if there is a default value

goto-bus-stop:
Now `argument.is_required() == true` only if the type is non-null and there is no
default value, meaning a value must be provided when it's used. This is how the
[spec](https://spec.graphql.org/draft/#sec-Required-Arguments) describes required arguments.

This seems more useful as the previous implementation can already be achieved
by checking `argument.ty.is_non_null()`.

It's kind of a fix, kind of just a change, kind of breaking...


SimonSapin (approved):
Do we know of any callers, such as in router or federation-next? Do they need to change?

goto-bus-stop:
I didn't find any!

PR #811, opened 2024-01-22
Add constness to input value generation

### LINKED ISSUE #810: fuzz:parser invalid input values
When executing `cargo fuzz run parser` documents are created with input values with variables e.g.

```
interface s {
  U0(s0: ID = $FQAAAAAAAAAAAAAAAAAAAAAAA, A1: A0, A2: A0): A0
  M1: A0
}
```

This is invalid as per https://spec.graphql.org/October2021/#Value
Variables should not be used in a const context. This means that within a graphql schema variables should not be used, only within an operation.

BrynCooke:
As per the spec: https://spec.graphql.org/October2021/#Value Values should not be variables in a const context. 

The TLDR of this is that within schema definitions variables should not be used on arguments. Within operations variables can should be used.

Fixes #810

### DIFF
--- a/crates/apollo-smith/src/argument.rs
+++ b/crates/apollo-smith/src/argument.rs
@@ -77,10 +77,10 @@ impl TryFrom<apollo_parser::cst::Argument> for Argument {
 
 impl<'a> DocumentBuilder<'a> {
     /// Create an arbitrary vector of `Argument`
-    pub fn arguments(&mut self) -> ArbitraryResult<Vec<Argument>> {
+    pub fn arguments(&mut self, constness: Constness) -> ArbitraryResult<Vec<Argument>> {
         let num_arguments = self.u.int_in_range(0..=4)?;
         let arguments = (0..num_arguments)
-            .map(|_| self.argument())
+            .map(|_| self.argument(constness))
             .collect::<ArbitraryResult<Vec<_>>>()?;
 
         Ok(arguments)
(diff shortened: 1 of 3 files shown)


crates/apollo-smith/src/argument.rs
```diff
@@ -102,7 +103,7 @@ impl<'a> DocumentBuilder<'a> {
     /// Create an arbitrary `Argument`
     pub fn argument(&mut self) -> ArbitraryResult<Argument> {
         let name = self.name()?;
-        let value = self.input_value()?;
+        let value = self.input_value(Constness::NonConst)?;
```
goto-bus-stop, line 106: Arguments should only take `const` values when applied on schema elements, so this should be passed in from outside

goto-bus-stop (commented):
Nice! I think there's some more places where this should be applied. constness should be propagated through directive applications and arguments, as you can do `@skip(if: $variable)` with variables in an operation, but not in a directive application in the schema.

crates/apollo-smith/src/argument.rs
BrynCooke: I've added it to the signature for `argument(s)` although we don't actually use them right now. I think we'll need to take a step back and look at the design again before we add actually use them as we will need to ensure that parameters are added in a consistent way across applications.

goto-bus-stop: oh i see!

goto-bus-stop (approved):

lrlna (approved):

PR #816, opened 2024-01-25
Fix field merging validation when multiple sub-selections partially overlap

goto-bus-stop:
**Current situation**
This is an offending query:

```rust
{
  root { ... on A { unrelated } }
  root { ... on B { overlapping } }
  root { ... on C { overlapping } }
}
```

Here, assume `B.overlapping` selects an `Int!`, and `C.overlapping` selects an `Int`. Clearly `root.overlapping` can either be `Int!` or `Int`, and so this must be an error. But previous versions of apollo-compiler do not report anything.

**Cause**
The validation is implemented by drilling in from the top down. The first field it encounters is `root`. `root` is selected multiple times so we need to see if its subselections can merge. In `main`, we do this by comparing every subselection to the *first* subselection (`... on A { unrelated }`). Both the `B` and `C` subselections can be merged with `A`, as they don't select any conflicting fields.

**Fix**
Comparing all selections of the same field, to the first occurrence of the same field is not enough, as other occurrences can conflict with each other but not with the first field. This PR considers all possible combinations of fields.

### DIFF
--- a/crates/apollo-compiler/src/validation/selection.rs
+++ b/crates/apollo-compiler/src/validation/selection.rs
@@ -198,7 +198,7 @@ pub(crate) fn same_response_shape(
 
             for fields_for_name in grouped_by_name.values() {
                 // 9. Given each pair of members subfieldA and subfieldB in fieldsForName:
-                for (subfield_a, subfield_b) in pair_combinations(&fields_for_name) {
+                for (subfield_a, subfield_b) in pair_combinations(fields_for_name) {
                     // 9a. If SameResponseShape(subfieldA, subfieldB) is false, return false.
                     same_response_shape(db, file_id, *subfield_a, *subfield_b)?;
                 }
--- a/crates/apollo-compiler/src/validation/selection.rs
+++ b/crates/apollo-compiler/src/validation/selection.rs
@@ -5,7 +5,15 @@ use crate::validation::{FileId, ValidationDatabase};
 use crate::{ast, schema, Node};
 
 use super::operation::OperationValidationConfig;
-/// TODO(@goto-bus-stop) test pathological query with many of the same field
+
+/// Return all possible unordered combinations of 2 elements from a slice.
+fn pair_combinations<T>(slice: &[T]) -> impl Iterator<Item = (&T, &T)> {
+    slice
+        .iter()
+        .enumerate()
+        // Final element will zip with the empty slice and produce no result.
+        .flat_map(|(index, element)| std::iter::repeat(element).zip(&slice[index + 1..]))
+}
 
 /// A field and the type it selects from.
 #[derive(Debug, Clone, Copy)]
@@ -188,12 +196,9 @@ pub(crate) fn same_response_shape(
             ));
             let grouped_by_name = group_fields_by_name(merged_set);
 
-            for (_, fields_for_name) in grouped_by_name {
+            for fields_for_name in grouped_by_name.values() {
                 // 9. Given each pair of members subfieldA and subfieldB in fieldsForName:
-                let Some((subfield_a, rest)) = fields_for_name.split_first() else {
-                    continue;
-                };
-                for subfield_b in rest {
+                for (subfield_a, subfield_b) in pair_combinations(&fields_for_name) {
                     // 9a. If SameResponseShape(subfieldA, subfieldB) is false, return false.
                     same_response_shape(db, file_id, *subfield_a, *subfield_b)?;
                 }
@@ -305,15 +310,12 @@ pub(crate) fn fields_in_set_can_merge(
     let mut diagnostics = vec![];
 
     for (_, fields_for_name) in grouped_by_name {
-        let Some((field_a, rest)) = fields_for_name.split_first() else {
-            continue; // Nothing to merge
-        };
-        let Ok(parent_a) = schema.type_field(field_a.against_type, &field_a.field.name) else {
-            continue; // Can't do much if we don't know the type
-        };
-
         // 2. Given each pair of members fieldA and fieldB in fieldsForName:
-        for field_b in rest {
+        for (field_a, field_b) in pair_combinations(&fields_for_name) {
+            let Ok(parent_a) = schema.type_field(field_a.against_type, &field_a.field.name) else {
+                continue; // Can't do much if we don't know the type
+            };
+
             // 2a. SameResponseShape(fieldA, fieldB) must be true.
             if let Err(diagnostic) = same_response_shape(db, file_id, *field_a, *field_b) {
                 diagnostics.push(diagnostic);
@@ -438,4 +440,20 @@ pub(crate) fn validate_selections(
     }
 
     diagnostics
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn pair_combinations_test() {
+        let pairs = pair_combinations::<i64>(&[1, 2, 3, 4]).collect::<Vec<_>>();
+        assert_eq!(
+            pairs,
+            &[(&1, &2), (&1, &3), (&1, &4), (&2, &3), (&2, &4), (&3, &4)]
+        );
+        let pairs = pair_combinations(&["a", "a", "a"]).collect::<Vec<_>>();
+        assert_eq!(pairs, &[(&"a", &"a"), (&"a", &"a"), (&"a", &"a")]);
+    }
 }
--- a/crates/apollo-compiler/test_data/diagnostics/0113_partially_overlapping_fragments.txt
+++ b/crates/apollo-compiler/test_data/diagnostics/0113_partially_overlapping_fragments.txt
@@ -1,0 +1,12 @@
+Error: operation must not select different types using the same field name `overlapping`
+    ╭─[0113_partially_overlapping_fragments.graphql:27:16]
+    │
+ 24 │     ... on B { overlapping }
+    │                ─────┬─────  
+    │                     ╰─────── `overlapping` has type `Int!` here
+    │ 
+ 27 │     ... on C { overlapping }
+    │                ─────┬─────  
+    │                     ╰─────── but the same field name has type `Int` here
+────╯
+
--- a/crates/apollo-compiler/test_data/diagnostics/0113_partially_overlapping_fragments.graphql
+++ b/crates/apollo-compiler/test_data/diagnostics/0113_partially_overlapping_fragments.graphql
@@ -1,0 +1,29 @@
+type A {
+  unrelated: String
+  overlapping: Int
+}
+type B {
+  overlapping: Int!
+}
+type C {
+  overlapping: Int
+}
+union Union = A | B | C
+type Query {
+  root: Union
+}
+
+# There is a conflicting type due to `B.overlapping` and `C.overlapping` both being selected
+# to `root.overlapping`, but neither overlap with the initial selection set for `root`.
+
+{
+  root {
+    ... on A { unrelated }
+  }
+  root {
+    ... on B { overlapping }
+  }
+  root {
+    ... on C { overlapping }
+  }
+}
--- a/crates/apollo-compiler/test_data/serializer/diagnostics/0113_partially_overlapping_fragments.graphql
+++ b/crates/apollo-compiler/test_data/serializer/diagnostics/0113_partially_overlapping_fragments.graphql
@@ -1,0 +1,36 @@
+type A {
+  unrelated: String
+  overlapping: Int
+}
+
+type B {
+  overlapping: Int!
+}
+
+type C {
+  overlapping: Int
+}
+
+union Union = A | B | C
+
+type Query {
+  root: Union
+}
+
+query {
+  root {
+    ... on A {
+      unrelated
+    }
+  }
+  root {
+    ... on B {
+      overlapping
+    }
+  }
+  root {
+    ... on C {
+      overlapping
+    }
+  }
+}


SimonSapin (approved):

lrlna (approved):

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -188,12 +196,9 @@ pub(crate) fn same_response_shape(
...
             let grouped_by_name = group_fields_by_name(merged_set);
 
-            for (_, fields_for_name) in grouped_by_name {
+            for fields_for_name in grouped_by_name.values() {
                 // 9. Given each pair of members subfieldA and subfieldB in fieldsForName:
-                let Some((subfield_a, rest)) = fields_for_name.split_first() else {
-                    continue;
-                };
-                for subfield_b in rest {
+                for (subfield_a, subfield_b) in pair_combinations(fields_for_name) {
```
SimonSapin, line 201: Do we expect real-world cases where `fields_for_name` can grow very large? Checking all pairs this way takes quadratic time. Trying to build a merged selection set just to throw it away seems wasteful but I think could be done in linear time.

SimonSapin: This kinda sounds like https://tech.new-work.se/graphql-overlapping-fields-can-be-merged-fast-ea6e92e0a01 which i still have as TODO to read. So maybe for later

goto-bus-stop: That was my thinking, but now comparing some crafted inputs to graphql-js, it looks like the JS implementation has some changes to speed it up dramatically while my implementation falls over. I had assumed that graphql-js was using the naive implementation. I'll see about implementing either the other algorithm or the graphql-js changes tomorrow, so we don't introduce a new easy DOS vector

goto-bus-stop:
I *think* i now have a faithful implementation of https://tech.new-work.se/graphql-overlapping-fields-can-be-merged-fast-ea6e92e0a01, except the caching. I *think* that it's valid for the output type check and the same field name/arguments check to be ran linearly, just like we did before this change, because of the grouping work that's done beforehand and because the recursions are now independent. Both checks assert that the elements they compare are the *same*, and `a == b && a == c` implies `b == c`. Previously that was also true, but the grouping worked differently, and it could miss out on combinations because of that.

Without caching, we end up grouping most selection sets in exactly the same way twice. It should be simple to add but i want to add more tests to make sure I did it right.

lrlna:
more tests sounds good! i'd also run this against the harness @garypen wrote to ensure correctness. do you think that with these changes we need to wait to rip out js validation out of the router and keep running the two implementations in parallel?

goto-bus-stop:
I ran it through the new harness, not comparatively, but just to make sure that it doesn't report anything weird. I think the biggest risk with this change is that there could still be a false negative, like before, so I would like to continue to run it side by side in prod for a few days and make sure our previous false negatives have all gone away.

lrlna:
that sounds good then - take it away!

crates/apollo-compiler/tests/validation/field_merging.rs
```diff
@@ -0,0 +1,1422 @@
...
+                │
+             20 │ fragment J on T {
+                │               ┬  
+                │               ╰── type condition here
+            ────╯
+        "#]],
+    );
+}
+
+mod return_types {
```
SimonSapin, line 852: Why not a module in a separate file? This file is already kinda long

SimonSapin (commented):
Initial comments before I review the actual algorithm

crates/apollo-compiler/src/executable/mod.rs
```diff
@@ -161,6 +161,34 @@ pub(crate) enum BuildError {
         field_name: Name,
         path: SelectionPath,
     },
+
+    #[error(
+        "{} can only have one root field",
+        subscription_name_or_anonymous(name)
+    )]
+    SubscriptionUsesMultipleFields {
```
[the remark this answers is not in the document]
goto-bus-stop: Yeah I wasn't sure about adding this here, it seemed like the most appropriate for now for new-style errors, but it definitely isn't a 100% fit.

crates/apollo-compiler/test_data/diagnostics/0075_merge_conflicting_args.txt
```diff
@@ -10,30 +10,6 @@ Error: operation must not provide conflicting field arguments for the same field
     │ 
     │ Help: Fields with the same response name must provide the same set of arguments. Consider adding an alias if you need to select fields with different arguments.
 ────╯
-Error: operation must not provide conflicting field arguments for the same field name `doesKnowCommand`
```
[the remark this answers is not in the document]
goto-bus-stop: The amount of duplicate error reports is reduced, yes. I should add it to the changelog!

crates/apollo-compiler/tests/validation/field_merging.rs
```diff
@@ -0,0 +1,1422 @@
...
+      }
+      fragment A on Type {
+        x: a
+      }
+      fragment B on Type {
+        x: b
+      }
+    "#,
+        expect![[r#"
+            Error: cannot find fragment `A` in this document
```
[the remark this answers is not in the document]
goto-bus-stop: totally right, I remember noticing this at some point but then forgot to address it by the time i committed 🙈

crates/apollo-compiler/test_data/diagnostics/0074_merge_identical_fields.txt
```diff
@@ -8,15 +8,15 @@ Error: operation must not select different types using the same field name `name
...
     ╭─[0074_merge_identical_fields.graphql:19:3]
     │
  18 │   name: nickname
     │   ───────┬──────  
-    │          ╰──────── `name` has type `String` here
+    │          ╰──────── field `name` is selected from field `nickname` here
  19 │   name
     │   ──┬─  
-    │     ╰─── but the same field name has type `String!` here
+    │     ╰─── but the same field `name` is also selected from field `name` here
```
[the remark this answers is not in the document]
goto-bus-stop: I've tried a slightly different approach here, also to address the "but" issue. Now explaining why it's invalid in the help message and unconditionally referring to the output name / response name / response key as the "alias" (you can only get into this situation with at least one alias as far as I understand, since we already know that the parent types overlap)
```
Error: cannot select multiple fields into the same alias `fido`
   ╭─[query.graphql:3:3]
   │
 2 │   fido: name
   │   ─────┬────  
   │        ╰────── `fido` is selected from `Dog.name` here
 3 │   fido: nickname
   │   ───────┬──────  
   │          ╰──────── `fido` is selected from `Dog.nickname` here
   │ 
   │ Help: Both fields may be present on the schema type, so it's not clear which one should be used to fill the response
───╯
```

crates/apollo-compiler/test_data/diagnostics/0078_merge_conflict_nested_fragments.txt
```diff
@@ -1,13 +1,13 @@
...
     │
  25 │   y: c
     │   ──┬─  
-    │     ╰─── field `y` is selected from field `c` here
+    │     ╰─── but the same field `y` is also selected from field `c` here
     │ 
  28 │   y: d
     │   ──┬─  
-    │     ╰─── but the same field `y` is also selected from field `d` here
+    │     ╰─── field `y` is selected from field `d` here
```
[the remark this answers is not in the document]
goto-bus-stop: Tried to address this by making the label text order-independent. otherwise we would have to sort the fields by source location, which I guess is not impossible, but i think it's unnecessary

crates/apollo-compiler/test_data/diagnostics/0113_partially_overlapping_fragments.txt
```diff
@@ -0,0 +1,12 @@
+Error: operation must not select different types using the same field name `overlapping`
+    ╭─[0113_partially_overlapping_fragments.graphql:27:16]
+    │
+ 24 │     ... on B { overlapping }
+    │                ─────┬─────  
+    │                     ╰─────── `overlapping` has type `Int!` here
+    │ 
+ 27 │     ... on C { overlapping }
+    │                ─────┬─────  
+    │                     ╰─────── but the same field name has type `Int` here
```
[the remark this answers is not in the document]
goto-bus-stop: maybe this would be clarified by using the selection path, and saying that it is selected from both `B.overlapping: Int!` and `C.overlapping: Int`, without explicitly talking about "fields" and "response" anythings.

crates/apollo-compiler/tests/validation/field_merging.rs
```diff
@@ -0,0 +1,1394 @@
...
+                 13 │     x: c
+                    │     ──┬─  
+                    │       ╰─── `x` is selected from `Type.c` here
+                    │ 
+                 20 │   x: b
+                    │   ──┬─  
+                    │     ╰─── `x` is selected from `Type.b` here
+                    │ 
+                    │ Help: Both fields may be present on the schema type, so it's not clear which one should be used to fill the response
+                ────╯
```
goto-bus-stop, line 644: This test is specifically meant to check that we don't produce too many warnings, and I think this may be one too many (but it may not be). It would be more obvious to see if this is correct as is, if it included a `SelectionPath`

SimonSapin:
I made sure the XING blog post has a copy at https://web.archive.org/web/20240208084612/https://tech.new-work.se/graphql-overlapping-fields-can-be-merged-fast-ea6e92e0a01?gi=3e85e33327df

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -1,112 +1,229 @@
...
+            .all(|(left, right)| same_value(left, right)),
+        (ast::Value::Object(left), ast::Value::Object(right)) if left.len() == right.len() => {
+            // This check could miss out on keys that exist in `right`, but not in `left`, if `left` contains duplicate keys.
+            // We assume that that doesn't happen. GraphQL does not support duplicate keys and
+            // that is checked elsewhere in validation.
+            left.iter().all(|(key, value)| {
+                right
+                    .iter()
+                    .find(|(other_key, _)| key == other_key)
+                    .is_some_and(|(_, other_value)| same_value(value, other_value))
```
SimonSapin, line 201: Why would this not miss keys in right but not left, regardless of duplicates?

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -1,112 +1,229 @@
...
+                original_value: original_arg.map(|arg| (*arg.value).clone()),
+                conflicting_location: field_b.field.location(),
+                conflicting_coordinate: field_b.coordinate().with_argument(arg.name.clone()),
+                conflicting_value: redefined_arg.map(|arg| (*arg.value).clone()),
+            }
+        };
+
+    // Check if fieldB provides the same argument names and values as fieldA (order-independent).
+    for arg in &field_a.field.arguments {
+        let Some(other_arg) = field_b.field.argument_by_name(&arg.name) else {
```
SimonSapin, line 160: I don’t expect this to ever be a problem in non-adversarial documents, but this takes O(n*m) time. Maybe convert both argument lists to maps if either has more than like 20 entries?

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -1,112 +1,229 @@
...
+            .zip(right.iter())
+            .all(|(left, right)| same_value(left, right)),
+        (ast::Value::Object(left), ast::Value::Object(right)) if left.len() == right.len() => {
+            // This check could miss out on keys that exist in `right`, but not in `left`, if `left` contains duplicate keys.
+            // We assume that that doesn't happen. GraphQL does not support duplicate keys and
+            // that is checked elsewhere in validation.
+            left.iter().all(|(key, value)| {
+                right
+                    .iter()
+                    .find(|(other_key, _)| key == other_key)
```
SimonSapin, line 200: Same as above with arguments, this takes O(n*m) time

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -163,205 +275,274 @@ pub(crate) fn same_response_shape(
...
+
+/// A boolean that turns on after the first check.
+struct OnceBool(std::cell::Cell<bool>);
+impl OnceBool {
+    fn new() -> Self {
+        Self(false.into())
+    }
+
+    /// Returns `false` the first time it is called, then returns `true` forever.
+    fn check(&self) -> bool {
```
SimonSapin, line 292: Maybe renaming this `already_done`?

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -163,205 +275,274 @@ pub(crate) fn same_response_shape(
...
                 }
             }
 
-            Ok(())
+            let mut nested_selection_sets = fields_for_name
+                .iter()
+                .map(|selection| &selection.field.selection_set)
+                .filter(|set| !set.selections.is_empty())
+                .peekable();
+            // TODO cache
```
SimonSapin, line 348: Is a follow-up planned or is this a "maybe some day" TODO? Or is the comment outdated and the relevant cache already there?

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -163,205 +275,274 @@ pub(crate) fn same_response_shape(
...
+            let selection_for_name = validator.lookup(fields_for_name.clone());
+            for fields_for_parents in selection_for_name.group_by_common_parents(validator.schema) {
+                // 2bi. fieldA and fieldB must have identical field names.
+                // 2bii. fieldA and fieldB must have identical sets of arguments.
+                // The same arguments check is reflexive so we don't need to check all
+                // combinations.
+                let Some((field_a, rest)) = fields_for_parents.split_first() else {
+                    continue;
+                };
+                for (field_a, field_b) in std::iter::repeat(field_a).zip(rest.iter()) {
```
SimonSapin, line 382:
```suggestion
                for field_b in rest {
```

(Same as above)

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -163,205 +275,274 @@ pub(crate) fn same_response_shape(
...
+            let mut map = IndexMap::new();
+            for selection in &self.selections {
+                match map.entry(selection.field.response_key().clone()) {
+                    Entry::Vacant(entry) => {
+                        entry.insert(vec![*selection]);
+                    }
+                    Entry::Occupied(mut entry) => {
+                        entry.get_mut().push(*selection);
+                    }
+                }
```
SimonSapin, line 414:
```suggestion
                map.entry(selection.field.response_key().clone())
                    .or_default()
                    .push(*selection);
```

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -163,205 +275,274 @@ pub(crate) fn same_response_shape(
...
-                    original_value: Some((*arg.value).clone()),
-                    redefined_selection: loc_b,
-                    redefined_value: None,
-                },
-            ));
-        };
+    /// Returns potentially overlapping groups of fields. Fields overlap if they are selected from
+    /// the same concrete type or if they are selected from an abstract type (future schema changes
+    /// can make any abstract type overlap with any other type).
+    fn group_by_common_parents(&self, schema: &schema::Schema) -> &Vec<Vec<FieldSelection<'doc>>> {
```
SimonSapin, line 423: Should this return a Vec of (hash?)set instead of a Vec of Vec?

crates/apollo-compiler/src/validation/selection.rs
goto-bus-stop: Yeah this is outdated!

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -1,112 +1,229 @@
...
-                    seen,
-                    inline.type_condition.as_ref().unwrap_or(against_type),
-                    &inline.selection_set,
-                ),
-                ast::Selection::FragmentSpread(spread) => {
-                    if seen.contains(&spread.fragment_name) {
-                        return vec![];
+/// Check if a selection set contains a fragment cycle, meaning we can't run recursive
+/// validations on it.
+fn contains_any_fragment_cycle(
```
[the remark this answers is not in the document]
goto-bus-stop: Removed this as loops should be handled by cache hits; plus now that recursions go through the `FieldsInSetCanMerge` struct, it was easy to add a limit to additionally catch the case where there is an actually very deep selection.

crates/apollo-compiler/src/validation/selection.rs
goto-bus-stop: I added an `ArgumentLookup` type for this, it's a good shout. It could also potentially happen in arguments validation when matching values to the argument definition, however there `m` is decided by the schema, so `O(m*n)` doesn't grow exponentially.

goto-bus-stop:
Should add a test for the recursion limit here

crates/apollo-compiler/tests/validation/recursion.rs
```diff
@@ -105,6 +117,7 @@ fn long_fragment_chains_do_not_overflow_stack() {
     .expect_err("must have recursion errors");
 
     let expected = expect_test::expect![[r#"
+        Error: too much recursion
```
goto-bus-stop, line 120: This test now hits a recursion limit in two places, once in fragment validation and once in field merging validation, and emitting two errors for it is correct enough in my opinion

lrlna:
Now that the recursion tests are in, shall we merge?

PR #817, opened 2024-01-29
chore(compiler): explicitly use `IndexMap::{swap,shift}_remove`

goto-bus-stop:
The indexmap crate has deprecated the `::remove` method.

I think `shift_remove()` is the correct one to use in both cases: we
want to maintain order even for orphan type extensions.

Fixes a deprecation warning that is preventing linting from passing on CI.


goto-bus-stop:
Taking Simon's verbal approval of using `shift_remove` (without seeing the PR) as approval of this change 😁 😇

PR #820, opened 2024-01-31
chore(parser): reduce intermediate string allocations

goto-bus-stop:
By threading the appropriate input lifetime through lexer and parser methods as initially pointed out in https://github.com/apollographql/apollo-rs/pull/814, we can avoid copying strings in a few places.

Especially nice is that we can avoid copying the same strings multiple times when peeking at tokens, and we can avoid an eager copy that is almost always just thrown away in the implementation of `.expect()`.


crates/apollo-parser/src/parser/mod.rs
```diff
@@ -103,9 +103,9 @@ pub struct Parser<'a> {
 /// Defaulting to around a quarter of that, to keep a comfortable safety margin.
 const DEFAULT_RECURSION_LIMIT: usize = 500;
 
-impl<'a> Parser<'a> {
+impl<'input> Parser<'input> {
```
goto-bus-stop, line 106: Renamed the lifetime so it's clear what it refers to when you're looking at a return value somewhere a few pages down in the impl block

goto-bus-stop:
Reverted the change making `Token` copyable, it was kinda worse to read and on top of that it was actually slower to copy tokens than to use references by about 10% according to benchmarks. Now the ownership of tokens is the same as before, only the lifetime of the source text inside the tokens has changed.

SimonSapin (approved):
Nice!

PR #828, opened 2024-02-09
fix(parser): ensure all loops advance parsing, fuzz with arbitrary bytes

goto-bus-stop:
This addresses an issue where the parser could get stuck in a loop if the token limit was reached at a specific point. Most egregiously, the parser would get in an infinite loop if the token limit was reached in the middle of a fragment spread: `{ ... <LIMIT> fragmentName }`.

To catch cases like this, this PR introduces `p.peek_while()`, which replaces `while let Some() = p.peek()` in the parser. In debug mode `peek_while()` asserts that parsing was advanced by the iteration. My current solution to cases that did **not** advance parsing is to change error reports to use `p.err_and_pop()`, which consumes the token that caused the error. This probably isn't always the most correct thing for error-tolerance, since the token may be a closing token or something that could be interpreted more optimally, but we're not yet diligent about that to begin with and running into an infinite loop is worse :)

`p.peek_while()` requires you return `ControlFlow`, which lets you break out of the loop when the peeked token did not match an expected token.

`p.peek_while_kind()` supports a simpler case where you expect one specific token kind to come up every time. If that token kind is spotted, you *must* parse at least that token. You can't break out of the loop early, it will end when the condition doesn't match anymore.

`p.parse_separated_list()` supports the various separated lists with optional prefix in the spec: `& Interface`, `| Union`, and `| DirectiveLocation`. Using this also makes directive location parsing no longer recursive (it was recursive but didn't have a recursion limit until now! 😱 )

Root operation type parsing is no longer recursive.

This also adds a fuzz test using completely arbitrary strings as input to the parser--all our other fuzz tests generate a document with apollo-smith, which is useful in its own way, but doesn't do a great job of finding issues in error edge cases like this. The new fuzz tests has a token limit as well. I found a few more cases that could panic with this.
- my own non-recursive root operation type parsing 🤪 
- stray top-level string value at the token limit
- token limit reached in the middle of parsing a type reference


crates/apollo-parser/src/parser/grammar/document.rs
```diff
@@ -13,16 +14,19 @@ use crate::{
...
             p.recursion_limit.current, 0,
             "unbalanced limit increment / decrement"
         );
 
-        match node {
+        match kind {
             TokenKind::StringValue => {
-                let def = p.peek_data_n(2).unwrap();
-                select_definition(def, p);
+                if let Some(def) = p.peek_data_n(2) {
```
goto-bus-stop, line 25: Could return `None` if the token limit was reached

crates/apollo-parser/src/parser/grammar/operation.rs
```diff
@@ -1,46 +1,6 @@
...
-use crate::{Parser, SyntaxKind, TokenKind, S, T};
-
-/// RootOperationTypeDefinition is used in a SchemaDefinition. Not to be confused
-/// with OperationDefinition.
-///
-/// See: https://spec.graphql.org/October2021/#RootOperationTypeDefinition
-///
-/// *RootOperationTypeDefinition*:
-///    OperationType **:** NamedType
-pub(crate) fn root_operation_type_definition(p: &mut Parser, is_operation_type: bool) {
```
goto-bus-stop, line 12: this belongs with the schema definition, not with operation definitions, in my opinion

crates/apollo-parser/src/parser/grammar/selection.rs
```diff
@@ -52,42 +54,43 @@ pub(crate) fn field_set(p: &mut Parser) {
...
-            TokenKind::Name => {
-                field::field(p);
-                has_selection = true;
-            }
-            _ => {
-                if !has_selection {
-                    p.err("expected at least one Selection in Selection Set");
+                // If there is no token,
+                None => {
+                    p.err_and_pop("expected an Inline Fragment or a Fragment Spread");
```
goto-bus-stop, line 78: changing this from `p.err()` to `p.err_and_pop()` fixes the infinite loop.

crates/apollo-parser/test_data/parser/err/0054_root_operation_type_with_extra_brackets.graphql
```diff
@@ -0,0 +1,5 @@
+schema {
+  query: Query
+  { mutation: Mutation
```
goto-bus-stop, line 3: the recursive root operation type parsing implementation allowed this

crates/apollo-parser/src/parser/grammar/ty.rs
```diff
@@ -85,6 +87,7 @@ fn parse<'a>(p: &mut Parser<'a>) -> Result<(), Token<'a>> {
 /// *NamedType*:
 ///     Name
 pub(crate) fn named_type(p: &mut Parser) {
+    // TODO(@goto-bus-stop) can we make this error instead if no name is found?
```
SimonSapin, line 90: Yes, I don’t know if the parser as a whole has inputs where this makes an observable difference, but this function silently not doing anything seems bad on principle. I’ve filed https://github.com/apollographql/apollo-rs/issues/829 so we don’t forget to look into it.

SimonSapin (approved):

fuzz/fuzz_targets/parser_limited.rs
```diff
@@ -0,0 +1,21 @@
...
+    let parser = panic::catch_unwind(|| Parser::new(data));
+    let parser = match parser {
+        Err(err) => {
+            panic!("error {err:?}");
+        }
+        Ok(p) => p.token_limit(500),
+    };
+
+    // This will have errors--we just need to make sure it does not run forever.
+    let _tree = parser.parse();
```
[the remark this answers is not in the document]
goto-bus-stop: now with peek_while, i guess it's more relevant that it doesn't _crash_ :)

to catch an actual infinite loop I think you would have to watch for it and kill manually.

crates/apollo-parser/src/parser/mod.rs
```diff
@@ -428,6 +429,63 @@ impl<'input> Parser<'input> {
...
+            if kind != expect {
+                break;
+            }
+
+            let before = self.current_token.clone();
+            run(self);
+            debug_assert!(
+                before != self.current_token,
+                "peek_while_kind() iteration must advance parsing"
+            );
```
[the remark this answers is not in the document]
goto-bus-stop: they both compare the whole token ... position could maybe work, but Eof is zero-width, so if we did consume a token but it was Eof that would have to be handled specially.

crates/apollo-parser/src/parser/grammar/document.rs
goto-bus-stop: Then you would get a `Some(Eof)` here. that's something i forgot to mention in the PR description: a token limit immediately cuts off the stream, while end of file is its own token. I considered emitting an Eof token when a limit is reached, and I do think that could make sense, but I felt that removing the unwraps from the parser was worthwhile anyways--`peek_data_n(2)` would not panic if there was always an Eof token still in the queue, but relying on that detail implicitly throughout the codebase is not great. So in this PR I did not add an Eof token when the limit is reached.

crates/apollo-parser/src/parser/grammar/selection.rs
goto-bus-stop: I don't have a good answer for this, I will merge and file an issue

PR #831, opened 2024-02-13
Exclude the fuzz crate from default-members

SimonSapin:
… so that libfuzzer-sys is only compiled when needed. This is an attempt to work around a Windows CI failure where `g++.exe` isn’t found anymore:
https://app.circleci.com/pipelines/github/apollographql/apollo-rs/2214/workflows/e6b564ef-82af-4f25-95a5-9155e1f6c926/jobs/16157

`cargo fuzz run` still works as before. Presumably it looks for `[package.metadata] cargo-fuzz = true` in `Cargo.toml` of all members, not just default ones.


goto-bus-stop (approved):
lgtm but what da hell is this
image

goto-bus-stop:
the DoubleEndedIterator fix is on main already (i meant to PR for it but pushed to main on accident)

SimonSapin:
Looks like the explanation is in the "Install Rustup" step:

```
warning: it looks like you have an existing installation of Rust at:
warning: C:\ProgramData\chocolatey\bin
warning: It is recommended that rustup be the primary Rust installation.
warning: Otherwise you may have confusion unless you are careful with your PATH
warning: If you are sure that you want both rustup and your already installed Rust
warning: then please reply `y' or `yes' or set RUSTUP_INIT_SKIP_PATH_CHECK to yes
warning: or pass `-y' to ignore all ignorable checks.
error: cannot install while Rust is installed
warning: continuing (because the -y flag is set and the error is ignorable)

lrlna (approved):

SimonSapin:
https://github.com/apollographql/apollo-rs/pull/832 fixed CI. We can revisit later to make things easier to build on Windows.

PR #835, opened 2024-02-14
fix(compiler): accept enum value (unquoted string) as input for custom scalars

### LINKED ISSUE #834: Enum values provided to scalars
Validation does not currently accept something like this:

```graphql
scalar Currency
type Query {
  convertToUSD(amount: Int!, currency: Currency!): Int!
}

query {
  convertToUSD(amount: 100, currency: EUR)
}
```

which is accepted by graphql-js and (rarely) used by customers. I don't love it, but we already accept enum values nested inside lists or objects as input for custom scalars, so it makes sense to accept this too.

goto-bus-stop:
Validation does not currently accept something like this:

```graphql
scalar Currency
type Query {
  convertToUSD(amount: Int!, currency: Currency!): Int!
}

query {
  convertToUSD(amount: 100, currency: EUR)
}
```

which is accepted by graphql-js and (rarely) used by customers. I don't love it, but we already accept enum values nested inside lists or objects as input for custom scalars, so it makes sense to accept this too.

Fixes #834

### DIFF
--- a/crates/apollo-compiler/src/validation/value.rs
+++ b/crates/apollo-compiler/src/validation/value.rs
@@ -159,6 +159,9 @@ pub(crate) fn value_of_correct_type(
             _ => diagnostics.push(unsupported_type(arg_value, ty)),
         },
         ast::Value::Enum(value) => match &type_definition {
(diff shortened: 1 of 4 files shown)


lrlna (approved):
thank youuu

PR #838, opened 2024-02-16
Remove Salsa database for 1.2× ~ 2× validation speed

SimonSapin:
See changelog


lrlna (approved):
we did talk about this for a while, i suppose it's time!

SimonSapin:
I think it definitely can, when used a certain way. But that’s hard to reconcile with using apollo-compiler 1.0’s arbitrarily mutable data structures as input.

PR #840, opened 2024-03-13
Add unstable error name getter to validation errors

goto-bus-stop:
This is not a proper feature! Error codes for validation errors are desired but need careful consideration. This is just a debugging aid so we can get information on validation bugs that were found by folks running the Apollo Router.


lrlna (approved):
excellent, let's go!

PR #845, opened 2024-03-26
Bug fix - implement missing parser validation for interface and object having missing fields

tinnou:
## Description

The parser is a bit lenient around letting types being defined without fields.

The example below should fail parsing but doesn't today:
```
type User {
}
```

This PR implements the missing validation to bail parsing for object types and interfaces when defined as empty (within curly braces).

The parser already implemented this validation for enum and input types.


goto-bus-stop (approved):
Looks perfect! 🥳 thank you!

PR #847, opened 2024-03-26
Implement missing schema validation - types must have one or more (fields, enum values, members, input values)

tinnou:
## Description 

Per the [GraphQL spec]( https://spec.graphql.org/draft/#sel-FAHZhCFDBAACDA4qe), object, interface, enum, union, and input types must define one or more fields, enum values, members, input values respectively across all extensions.

This PR implements the necessary validation for each of the aforementioned types.

As an example, previously, the following SDL would incorrectly pass schema validation:
```
directive @foo on OBJECT
type Query {
  foo: User
}
type User 
extend type User @foo
```
`User` does not define any fields, even across its extensions.


goto-bus-stop (approved):
Thank you!

PR #848, opened 2024-03-28
feat(compiler): serialize with initial indentation

SimonSapin:
This helps use indentation in something that is partly but not entirely GraphQL syntax, such as the debugging representation of an Apollo Router query plan. Example: `document.serialize().initial_indent_level(2).to_string()`


goto-bus-stop (approved):
👍🏻

PR #853, opened 2024-04-12
Add a graphql-js compatible-ish error serialization for Apollo Router

goto-bus-stop:
With some additions to the error structures to pass through the data required.

This PR adds a `.unstable_to_json_compat()` function to diagnostics that returns best-effort error messages in the graphql-js format. Some things like "Did you mean?" are omitted intentionally. There are minor differences in a handful of other messages, but much fewer than before, and hopefully it'll be obvious to users that the meaning is the same. One example difference is the `UnsupportedValueType` message, which in graphql-js is a customisable type-specific validation, but in apollo-rs is a fixed check based only on what the spec knows about builtin scalars, so the error is a bit less specific.

The new API is only meant for use in the Apollo Router, temporarily, exempt from semver, and any other users will not be supported.

### DIFF
--- a/crates/apollo-compiler/CHANGELOG.md
+++ b/crates/apollo-compiler/CHANGELOG.md
@@ -17,6 +17,16 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## Maintenance
 ## Documentation-->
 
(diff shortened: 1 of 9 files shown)


lrlna (approved):
thank you for doing this one!

PR #861, opened 2024-05-24
Convert SourceLocation to Range<GraphQLLocation>, not just starting line/column

dylan-apollo:
Needed for diagnostics in https://github.com/apollographql/router/pull/5247 .

Seems like maybe we want to have a single struct somewhere instead which returns start and end to save a bit of extra work in calling both methods? But not sure how to align that with the need of `GraphQLLocation` to be the spec's execution error location.


goto-bus-stop:
Maybe instead of `_start` and `_end`, we could have a method that returns a `(GraphQLLocation, GraphQLLocation)` tuple? This would address the double-lookup and I think it would be extremely rare to *only* need the end but not the start.

The methods could be called `get_line_column` and `get_line_column_range` for getting the start only or the start+end range respectively.

SimonSapin:
+1 for `get_line_column` and `get_line_column_range`. The latter could even return `std::ops::Range<GraphQLLocation>` (or `Option` of)

trevor-scheer:
Having the range available would be a nice convenience for the language server. The type that's most interesting for my use case is the [`lsp_types::Range`](https://docs.rs/lsp-types/latest/lsp_types/struct.Range.html) which says 0 indexed with an exclusive end, so that sounds compatible with the suggested `Range<GraphQLLocation>` approach. I am indifferent to tuple or struct, I think in this case the ordering of the tuple is very clear. A struct would match the `lsp_types::Range` type more closely, but in either case I will be mapping the type.

dylan-apollo:
I've reworked this PR to a completely new proposal that uses _new_ structs for representing both line/col, decoupled from the GraphQL error representation.

However, given that we're going to need an exclusive end & zero-indexing for LSP (and therefore composition), this may need further tweaking to get everything just right, maybe with `Range<LineColumn>` instead of the new dedicated struct. Interested to hear any ideas on how to best handle exclusive ends.

dylan-apollo:
Also, a quick prototype of these changes in `router`: https://github.com/apollographql/router/pull/5263

crates/apollo-compiler/src/name.rs
```diff
@@ -184,9 +185,9 @@ impl Name {
         }
     }
 
-    /// If this string contains a location, convert it to line and column numbers
-    pub fn line_column(&self, sources: &SourceMap) -> Option<GraphQLLocation> {
-        GraphQLLocation::from_node(sources, self.location())
+    /// If this string contains a location, convert it to starting line and column numbers
+    pub fn line_column_range(&self, sources: &SourceMap) -> Option<Range<LineColumn>> {
```
SimonSapin, line 189: "starting" in this comment looks wrong?

crates/apollo-compiler/src/node.rs
```diff
@@ -107,9 +108,10 @@ impl<T: ?Sized> Node<T> {
         self.location().map(|l| l.file_id()) == Some(FileId::BUILT_IN)
     }
 
-    /// If this node contains a location, convert it to line and column numbers
-    pub fn line_column(&self, sources: &SourceMap) -> Option<GraphQLLocation> {
-        GraphQLLocation::from_node(sources, self.location())
+    /// If this node contains a location, convert it to the line and column numbers of the
+    /// start of the node.
+    pub fn line_column_range(&self, sources: &SourceMap) -> Option<Range<LineColumn>> {
```
SimonSapin, line 113: Same as above re "start of the node"

crates/apollo-compiler/src/parser.rs
```diff
@@ -362,12 +362,23 @@ impl SourceFile {
...
+/// A specific location within the source of a file, like an offset.
+///
+/// See also [`crate::node::LineColumnRange`].
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub struct LineColumn {
+    /// The line number, starting at 0 for the first line (not 1 like [`GraphQLError`]).
+    pub line: usize,
+    /// The column number, starting at 0 for the first character of the line.
+    pub column: usize,
+}
```
SimonSapin, line 380: It seems awkward to have two different structs in the public API that are the same except off by one. @goto-bus-stop what do you think?

dylan-apollo: Yeah, the challenge is to draw a distinction between a generic location in source code and the definition of location within a GraphQL Error at runtime. They serve quite different purposes, in practice, but they sure look similar 🤔

SimonSapin: I’m questioning whether the distinction needs to exist at all in this library’s public API. Arguably 1-indexing is "more generic" since that’s how humans usually talk about line numbers. So apollo-compiler could have one line-column struct with a generic name that is also used in GraphQL errors, and conversion to 0-indexed for LSP purpose could happen in the language server. (The redundant +1 then -1 computation is extremely cheap.)

dylan-apollo: Cheap, yes, but very easy to make mistakes on, since we need to do them in several different repos today. I have already fixed several bugs relating to a + or -1 being missing in one line or another 😬. Feels bad to do it in more places than necessary.

If we have two canonical types, the + and -1 only need to be in a single place, and we can reference the correct type for each use-case from the rest.

SimonSapin: If we have both in the same library, especially because this is error prone, I think they should not be called "generic" and "GraphQL" but something like `ZeroIndexedLineColumn` and `OneIndexedLineColumn`. Kinda unpleasant to use but makes the difference very obvious.

That said, couldn’t the language server define an `LspLineColumn` type in one place and only do +1 or -1 in `From` conversions to or from that type?

goto-bus-stop: I guess it is a bit annoying for lsps because you are probably dealing with `lsp_types::Position` and `apollo_compiler::LineColumn`, and you can't implement `From<>` on types from external crates. You'd probably use a helper function instead of a `From<>` impl.
But I'd also prefer having just one type for line/column in apollo-compiler... We already have multiple location types (indices and line/column)

lrlna: I would rather that the compiler defines the line/column range that the lsp would use. I don't want them to deal with incrementing/decrementing it themselves, especially since it sounds like we've had bugs with several implementations already. Does it need to be different structs though? If `LineColumn` is used in context of a `Diagnostic` or `GraphQLError`, it's has an all together different meaning then when used in a context of a `Node`. Or is the worry that when looking at a struct definition of `Node` and `Diagnostic` and see `LineColumn` the two are actually not identical in range?

dylan-apollo: For me, the main benefit of two structs is using type checking to enforce offset checking. Meaning, if you do have a function that converts from `LineColumn` to `LspLineColumn` (or whatever), that function should _always_ work for those types. If the info about whether it's zero- or one-indexed is not in the struct itself (but instead some parent struct like `Diagnostic`), then you wouldn't be able to lean on the compiler to ensure that conversion is right.

Worth pointing out as well, I think, that there are _many_ location types spread around our various repos right now. I'd love to start consolidating those however possible 😁. There are 3 similar types just in `federation-rs`! I'm not sure if we want all of those crates to depend on this one, though.

dylan-apollo: That is assuming that only the LSP needs 0-indexed positions. I'm not sure that's true. It also _feels_ bad for this crate to be constantly converting _to_ 1-indexed for every location just for the consumer (LSP) to be subtracting again. I guess in practice it's a very quick operation, but my brain doesn't like it 😆.

goto-bus-stop: Yeah. I think it's preferable to have just one. An LSP needs 0-indexed but other use cases (including graphql server error responses) need 1-indexed. Someone always has to convert, and IMO conversion is preferable to having two APIs that return different indexes (even if it's a different type). apollo-rs provides graphql JSON error types so we have a requirement in the crate to *at least* have a 1-indexed type. I think we need a *strong* reason to add another one...

dylan-apollo: Since the main hold-up is about the second struct, I removed it and switched everything back to 1-indexed. So now the only change in this PR is returning `Range` from a few places instead of single locations.

I'll move the zero-indexed struct to a different shared crate (`apollo-composition`, probably) so that everyone that needs to convert can re-use that one.

SimonSapin (approved):

PR #866, opened 2024-06-12
apollo-smith: more varied fields generation

Geal:
This changes the field generation algorithm in apollo-smith to allow more variety, because the previous implementation was too biased towards the first field specified in a type.
This also adds an example to generate a random query from a schema.

One thing that comes out of these tests is that we need the input data for Arbitrary to be large enough, otherwise once it has no more bytes to consume, it returns 0 every time and the selections get no more variations

### DIFF
--- a/crates/apollo-smith/Cargo.toml
+++ b/crates/apollo-smith/Cargo.toml
@@ -31,3 +31,4 @@ thiserror = "1.0.37"
 
 [dev-dependencies]
 expect-test = "1.4"
+rand = "0.8.5"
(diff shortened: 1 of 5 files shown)


lrlna (approved):

lrlna:
This does make the snapshots fail, and I think that failure is legit. We should have _some_ fields in the operation's selection set and the fragment's selection set

Geal:
you're right, I messed up the lower bound of the range. I set it to at least 1 selection instead of 2 and updated the snapshots. We do not lose much by doing that

PR #868, opened 2024-06-14
New memory representation for Name/NodeStr

SimonSapin:
This is the result of a series of experiments. I’m not sure I like it, but let’s discuss it.

## TL;DR

`NodeStr` is no more. GraphQL string values use a plain `String` (the `Value` enum is already in a `Node` with the same source span) and descriptions use `Node<str>`. `Name` has a dual representation based on either `Arc<str>` or `&'static str` and can cheaply convert to that.

## Motivation

Apollo Router has found a performance bottleneck when using GraphQL type and field names in OpenTelemetry. [`opentelemetry::StringValue`](https://docs.rs/opentelemetry/latest/opentelemetry/struct.StringValue.html) can be converted cheaply from `String`, `Arc<str>`, or `&'static str` but `apollo_compiler::NodeStr` (usually) contains none of these so converting requires a new heap allocation. This happens in a hot path often enough that the allocation (and later deallocation) costs add up.

## `Name` and `NodeStr` in apollo-compiler 1.0.0-beta.17

`Name` is used for GraphQL names. It dereferences to `NodeStr` and is a thin wrapper that adds checking for GraphQL name syntax. (ASCII alphanumeric or underscore, not starting with a digit)

In addition to names, `NodeStr` is also used for GraphQL string `Value`s and descriptions. It contains an immutable `str` value and optional source location (file ID and text range) for diagnostics. 

Internally, `NodeStr` contains a single `ptr::NonNull<()>` with a tag in its lowest bit. The pointer is usually to a ([`triomphe::ThinArc`](https://docs.rs/triomphe/latest/triomphe/struct.ThinArc.html)) shared heap allocation that contains:

* Atomic reference counter
* Optional source location
* String length
* String bytes (variable length, inline following the header)

The pointer can instead (discriminated with the pointer tag) represent `&'static &'static str`. This is a double indirection, because `&'static str` is a wide pointer that stores the string length next to its address but we didn’t want to double the size of `NodeStr`. In this case there is no location information. This is used by the `name!("example")` macro which does GraphQL name syntax checking at compile-time then calls `NodeStr::from_static(&"example")` with an extra `&` which gets promoted to a constant.

A quality of this design is that `size_of::<NodeStr>()` is only 8 bytes.

## First idea: [`String::leak`](https://doc.rust-lang.org/std/string/struct.String.html#method.leak)

GraphQL type and field names in a schema rarely change. What if we accepted a small memory leak to turn them into `&'static str`?

However that’s not enough to use the existing static representation of `NodeStr`:

* `String::leak` doesn’t give us the double `&'static &'static str` indirection. We could potentially allocate `Box<&'static str>` and also leak that, but even then…
* For diagnostics we want type and field names in the schema to carry source location, which the current static `NodeStr` doesn’t.

## Overall direction for a new memory representation

If we’re reconsidering the design of `NodeStr`, we might as well pick one that makes conversion to `opentelemetry::StringValue` cheap in all cases. This means a dual representation based on either `Arc<str>` or `&'static str`. (The standard library `Arc` allocates space for a weak reference counter that we don’t use, but we can accept that cost. Also see below for a `Weak` come-back.)

This in turn means moving the source location outside of the allocation, to be inline in the `NodeStr` struct. At first this seems like a loss: if a string is cloned many times, its source location (or even space to record the lack thereof) is duplicated. But maybe that doesn’t happen that much? On the contrary, inline location could allow strings at different locations to share a heap allocation (hint hint more foreshadowing).

A straightforward `(enum { Arc<str>, &'static str }, Option<NodeLocation>)` has `size_of() == 40` bytes, which feels like too much.

## Layout optimizations

`rowan::TextRange` inside `NodeLocation` uses 32-bit offsets, so apollo-parser is already limited to 4 GiB inputs `NodeLocation` can be decomposed into `{ file_id, start_offset: u32, end_offset: u32 }`.

Both `Arc<str>` and `&'static str` are represented as `{ data: *const u8, len: usize }`. The length is kind of already represented in the location is the difference between end and start offsets. Accessing the length to build a `&str` is much more common than accessing the end offset, so we can store `{ start_offset: u32, len: u32 }` and compute the end offset on demand. When a synthetic string doesn’t have a source location, we can set `start_offset` to zero but still use `len`.

Finally, an actual Rust `enum` would allocate 64 bits to the 1-bit discriminant in order preserve memory alignment. Like before, we can stash that bit somewhere else:

* `&'static str` has `align_of() == 1` (it could be a sub-slice of some other string) so we can’t use the lower pointer bit for the tag
* The upper bit of `start_offset` would work, but bring the existing 4 GiB file input limit to 2 GiB. This could be ok but we can avoid it
* The upper bit of `len` would move that 2 GiB limit to the length of a single string sounds fine, but extra steps in the common code path of rebuilding a `&str` can easily be avoided
* `FileId` contains a `NonZeroI64`, we can easily spare a bit there

We end up with this representation which has `size_of() == 24` bytes:

```rust
struct {
    ptr: NonNull<u8>,
    len: u32,
    start_offset: u32,
    tagged_file_id: TaggedFileId,
}
```

## Consequences and tradeoffs

* Remember when the length of the string is kind of already in the source location? Only kind of. This is the case for `Name` whose value appear exactly in the GraphQL source, but string values and descriptions are quoted strings that can use backslash-escape sequences. Both of these make the source span longer than the decoded value.

  This PR chooses to remove `NodeStr` entirely and move the special representation directly into `Name`. GraphQL string values use a plain `String` (the `Value` enum is already in a `Node` with the same source span) and descriptions use ~~`Node<String>`~~ `Node<str>`

* Clippy started complaining because `executable::BuildError` was now had `size_of() == 264` bytes. I boxed its largest variants to bring that to 104 bytes, but the impact doesn’t stop there. Every `Name` will now incur an extra 16 bytes of inline size. On the other hand each heap allocation for `Name` will have 16 fewer bytes, so I’d guess the overall impact on memory use of a running Router is negligible.

* `FileId` contained `NonZeroI64` with -1 and -2 reserved for special values. `TaggedFileId` was easier to make without negative values, so now `NonZeroU64` is used and "normal" IDs start at 3 instead of 1. This causes a diff for every snapshot test that prints file IDs, but this is a one-time transition.

## Optional: leaking string interner

The above gives cheap conversion to `opentelemetry::StringValue` already, so we could stop there. But the original `String::leak` idea is more doable now. But if we’re gonna leak we should definitely deduplicated the leaked strings.

~~This PR includes (in a separate commit)~~ (**edit**: this inclusion was reverted for now) an opt-in process-wide string interner based on `OnceLock<RwLock<hashbrown::RawTable<&'static str>>>` and `String::leak`. It’s enabled by configuring `Parser::new().intern_and_leak_names()`. When *not* enabled, the parser does not leak any new string but still looks in the interner for strings already there. So `intern_and_leak_names` could for example be enabled when parsing a schema but not when parsing executable document, still allowing the latter to reuse already-interned field names.

Leaking is enabled for parsing `built_in.graphql`, so strings like `Int` or `__typename` are always interned.

## Going further: weak reference string interner

Leaking memory seems risky. Another possibility is to store `std::sync::Weak<str>` in the interner and give out `Arc<str>` on lookup. This would allow strings not references anymore to be de-allocated.

Before inserting into the hash table, the interner would check if it’s about to grow with `table.capacity() == table.len()`. In that case (since we’re about to do an `O(n)` reallocation of the table anyway) it would scan the table to remove entries with `Weak::strong_count() == 0` since those have been deallocated and can’t be used anymore.

A potential risk is atomic contention: atomic operations in `Arc` fast when uncontended but if many threads are all trying to increment and decrement the refcount of the one `Arc<str>` representing `"Query"` then they will start waiting on each other.


SimonSapin:
I’d like to do the change from `Node<String>` to `Node<str>` but it requires a change upstream: https://github.com/Manishearth/triomphe/pull/92

SimonSapin:
In today’s discussion there was clear reluctance about intentionally leaking memory, so I’ve reverted the addition of the leaking string interner. A future PR can consider adding a non-leaking interner based on `Arc` and `Weak`.

SimonSapin:
With a few small changes to make https://github.com/apollographql/router/pull/5477 compile and a changelog entry, this is now ready for review

crates/apollo-compiler/src/name.rs
```diff
@@ -0,0 +1,511 @@
...
+        self.as_static_str().ok_or_else(|| {
+            let manually_drop = self.as_arc().unwrap();
+            Arc::clone(&manually_drop)
+        })
+    }
+
+    /// Returns whether the given string is a valid GraphQL name.
+    ///
+    /// <https://spec.graphql.org/October2021/#Name>
+    pub const fn valid_syntax(value: &str) -> bool {
```
lrlna, line 261: i really like when functions returning a bool start with `is_` so _nobody_ is confused

```suggestion
    pub const fn is_valid_syntax(value: &str) -> bool {
```

crates/apollo-compiler/src/name.rs
```diff
@@ -0,0 +1,511 @@
...
+        } else {
+            None
+        }
+    }
+
+    /// If this `Name` was created with [`new_static`][Self::new_static]
+    /// or the [`name!`][crate::name] macro, return the string with `'static` lifetime.
+    ///
+    /// Otherwise, return a clone of the `Arc` used internally for this `Name`.
+    pub fn to_static_str_or_cloned_arc(&self) -> Result<&'static str, Arc<str>> {
```
lrlna, line 251: yeaaaaaa i think this needs to be a enum for a returned type. A bit annoying with the various `str`s, I know. Just doesn't feel right to use `Result` in a public API for something that isn't one.

crates/apollo-compiler/src/name.rs
```diff
@@ -0,0 +1,511 @@
...
+    const fn char_is_name_start(byte: u8) -> bool {
+        byte.is_ascii_alphabetic() || byte == b'_'
+    }
+
+    /// <https://spec.graphql.org/October2021/#NameContinue>
+    const fn char_is_name_continue(byte: u8) -> bool {
+        byte.is_ascii_alphanumeric() || byte == b'_'
+    }
+
+    pub fn to_component(&self, origin: ComponentOrigin) -> ComponentName {
```
lrlna, line 304: would this be an `into_component`, since we are doing a conversion from types that don't implement `Copy`?

crates/apollo-compiler/src/name.rs
```diff
@@ -0,0 +1,511 @@
...
+        } else {
+            Err(InvalidNameError {
+                name: value.to_owned(),
+                location,
+            })
+        }
+    }
+
+    /// <https://spec.graphql.org/October2021/#NameStart>
+    const fn char_is_name_start(byte: u8) -> bool {
```
lrlna, line 295:
```suggestion
    const fn is_char_name_start(byte: u8) -> bool {
```

crates/apollo-compiler/src/name.rs
```diff
@@ -0,0 +1,511 @@
...
+        byte.is_ascii_alphabetic() || byte == b'_'
+    }
+
+    /// <https://spec.graphql.org/October2021/#NameContinue>
+    const fn char_is_name_continue(byte: u8) -> bool {
```
lrlna, line 300:
```suggestion
    const fn is_char_name_continue(byte: u8) -> bool {
```

crates/apollo-compiler/src/name.rs
```diff
@@ -0,0 +1,511 @@
...
+    const fn parts_from_static(static_str: &'static str) -> (NonNull<u8>, u32, bool) {
+        let len = Self::new_len(static_str);
+        let ptr = static_str.as_ptr().cast_mut();
+        // SAFETY: `&'static str` is always non-null
+        let ptr = unsafe { NonNull::new_unchecked(ptr) };
+        (ptr, len, TAG_STATIC)
+    }
+
+    /// Returns start_offset
+    fn with_location(value: &str, location: &NodeLocation) -> u32 {
```
lrlna, line 167: `with_` is usually a constructor method. should we call this `start_offset`?
```suggestion
    fn start_offset(value: &str, location: &NodeLocation) -> u32 {
```

crates/apollo-compiler/src/name.rs
```diff
@@ -0,0 +1,511 @@
...
+        self
+    }
+}
+
+impl ToCliReport for InvalidNameError {
+    fn location(&self) -> Option<NodeLocation> {
+        self.location
+    }
+    fn report(&self, report: &mut CliReport) {
+        report.with_label_opt(self.location, "cannot be parsed as a GraphQL name");
```
lrlna, line 503: capitalising to indicate somehow it's a token
```suggestion
        report.with_label_opt(self.location, "cannot be parsed as a GraphQL Name");
```

crates/apollo-compiler/src/validation/mod.rs
```diff
@@ -25,8 +25,11 @@ pub(crate) mod variable;
 
 use crate::ast::Name;
 use crate::diagnostic::{CliReport, Diagnostic, ToCliReport};
-use crate::executable::BuildError as ExecutableBuildError;
 use crate::executable::VariableDefinition;
+use crate::executable::{
+    BuildError as ExecutableBuildError, ConflictingFieldArgument, ConflictingFieldName,
+    ConflictingFieldType,
+};
```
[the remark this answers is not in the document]
SimonSapin: In the router repo we configure rustfmt with [`imports_granularity = Item`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=import#imports_granularity) and [`group_imports = StdExternalCrate`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=group#group_imports). I tend to write new files this way, but haven’t configured rustfmt in this repo so it defaults to `Preserve` config.

I’d be in favor of adopting and enforcing a style in this repo too. The problem is those options are unstable in rustfmt, so they require a Nightly toolchain. If we have them in `rustfmt.toml` then stable rustfmt emits warnings that they’re ignored. What makes this especially bad is that `cargo fmt` calls rustfmt separately for every source file, so warnings are repeated enough time to flood the terminal. In the router repo for now we don’t configure `rustfmt.toml`, but have an xtask command that runs `cargo +nightly fmt -- --config […]`

crates/apollo-compiler/src/name.rs
```diff
@@ -0,0 +1,511 @@
...
+    Static(&'static str),
+}
+
+#[derive(Clone, Eq, PartialEq, thiserror::Error)]
+#[error("`{name}` is not a valid GraphQL name")]
+pub struct InvalidNameError {
+    name: String,
+    location: Option<NodeLocation>,
+}
+
```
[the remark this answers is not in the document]
SimonSapin: Despite having "invalid" in its name this is not an error returned by `Schema::validate` or `ExecutableDocument::validate`, only by `Name::new` or other constructors.

We made `Name` a distinct Rust type so that `validate()` does not need to check name validity at all: https://github.com/apollographql/apollo-rs/issues/710

crates/apollo-compiler/src/name.rs
```diff
@@ -0,0 +1,511 @@
...
+    assert_send_and_sync::<(UnpackedRepr, u32, TaggedFileId)>();
+};
+
+unsafe impl Send for Name {}
+
+unsafe impl Sync for Name {}
+
+impl Name {
+    /// Create a new `Name` parsed from the given source location
+    pub fn new_parsed(value: &str, location: NodeLocation) -> Result<Self, InvalidNameError> {
```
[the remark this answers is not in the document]
SimonSapin: There are ~~two~~ three independent axes for constructors:
* Stored as `Arc<str>` v.s. stored as `&'static str`
* With a source location (aka "parsed") v.s. without
* Checking syntax validity and returning a `Result`, or "unchecked" (used when already checked at compile-time or by apollo-parser)

All combinations are possible as far as `Name` is concerned (though static parsed is only useful with a leaking string interner). In a later commit I replaced the "parsed" constructors with a `with_location` builder-like chaining method to reduce the number of constructors without limiting possibilities.

But you’re right that `new_something` and `from_something` are still used a bit inconsistently. One argument is that `from_` only seems to say anything about the input parameter type, but static is also a possible internal representation inside of `Name`.

crates/apollo-compiler/src/name.rs
```diff
@@ -0,0 +1,511 @@
...
+        Self {
+            ptr,
+            len,
+            start_offset: 0,
+            tagged_file_id: TaggedFileId::pack(tag, FileId::NONE),
+            phantom: PhantomData,
+        }
+    }
+
+    fn parts_from_arc(value: &str) -> (NonNull<u8>, u32, bool) {
```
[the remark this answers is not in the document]
SimonSapin: This is inlined in a later commit

crates/apollo-compiler/src/name.rs
SimonSapin: `into_` is used when moving ownership of `self` away from the caller

crates/apollo-compiler/src/name.rs
SimonSapin: `char` (should be `byte` since i changed the argument type) is the subject of `is` not the object. I went with just `is_name_start`, it’s a private function and the type of argument is already apparent in the rest of the signature.

crates/apollo-compiler/src/name.rs
SimonSapin: This is inlined in a later commit

SimonSapin:
Sorry! At some point I tried something in a separate branch and forgot to merge back. I didn’t realize I was not pushing to this PR anymore and asked for review on something quite a bit behind from what I thought. All pushed now.

Remaining to bikeshed:

* Names of constructors (`new_*` v.s. `from_*`)
* Name of the `StaticOrArc` enum
* Name of the `to_static_str_or_cloned_arc` method that returns it
* Import path of `Name`, `InvalidNameError`, and `StaticOrArc`. Is three items the point where they deserve their own public module? (The `name!` macro would belong there too but `#[macro_export]` currently always exports at the crate root)

goto-bus-stop:
I haven't thought a lot about this but I think it makes sense to have an `apollo_compiler::name::{Name, InvalidNameError, StaticOrArc}` module and a reexport of `Name` at the root.

SimonSapin:
Per today’s discussion I removed the `StaticOrArc` enum and split the conversion into two separate `as_static_str()` and `to_cloned_arc()` methods returning options. It means conversion to OpenTelementry string needs an unwrap: https://github.com/apollographql/router/pull/5352#issuecomment-2176472531

crates/apollo-compiler/src/validation/mod.rs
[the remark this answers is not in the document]
lrlna: ahhhhh that makes sense. i got confused on what's formatted which way and where. It's not an issue, and doesn't really need the time to be put into it at the moment. Can figure it out at a later point!

crates/apollo-compiler/src/name.rs
```diff
@@ -0,0 +1,516 @@
...
+            // However it’s a `*const` and we already have a `NonNull` with the same address in `self`
+        }
+        Self { ..*self }
+    }
+}
+
+impl Drop for Name {
+    fn drop(&mut self) {
+        if let Some(arc) = &mut self.as_arc() {
+            // SAFETY: neither the dropped `ManuallyDrop` nor `self.ptr` is not used again
```
goto-bus-stop, line 313:
```suggestion
            // SAFETY: neither the dropped `ManuallyDrop` nor `self.ptr` is used again
```
right?

goto-bus-stop (approved):

crates/apollo-compiler/src/node.rs
```diff
@@ -267,39 +334,86 @@ impl fmt::Debug for FileId {
...
         } else {
             panic!()
         }
     }
 }
+
+impl TaggedFileId {
+    pub(crate) const fn pack(tag: bool, id: FileId) -> Self {
+        let tag_and_id = if tag {
+            debug_assert!((id.id.get() & TAG) == 0);
```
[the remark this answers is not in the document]
SimonSapin: Done!

PR #878, opened 2024-07-09
fix(compiler): remove potential hash collision bug from `FieldsInSetCanMerge`

goto-bus-stop:
In #816, to avoid cloning a big `Vec` for use as a hash key, I instead used the hash of the Vec as a key. I guess it's possible to get a hash collision that way and produce the wrong result.

Here I avoid the clone by allocating field selections into an arena allocator. The allocator outlives `FieldsInSetCanMerge` so we can pass around references to slices, and we can store the reference in the key and the value at the same time.

We already didn't deallocate the `Vec`s until the `FieldsInSetCanMerge` struct is dropped so the memory usage is the same.


lrlna (approved):
I see we have gotten to a bump allocator stage of compiler work hehehe

SimonSapin (approved):

PR #880, opened 2024-07-11
Mutation-based generator of arbitrary executable documents

SimonSapin:
For a given valid schema, use a byte string typically [provided by cargo-fuzz](https://rust-fuzz.github.io/book/cargo-fuzz/tutorial.html) as a source of entropy to deterministically generate a valid executable document:

```rust
pub fn arbitrary_valid_executable_document(
    schema: &Valid<Schema>,
    arbitrary_bytes: &[u8],
) -> Valid<executable::ExecutableDocument>
```

Instead of using the `arbitrary` crate, parts of it are forked for reasons documented at the top of the new `entropy.rs` file.

At a high level, it creates a minimal operation then keeps adding selections to it until entropy is exhausted. As a result, providing a longer byte string tends to create larger documents. Hopefully this helps fuzzers better explore the space of possible documents, compared to apollo-smith which uses code like `u.int_in_range(1..=5)` to decide how many items to generate regardless of remaining entropy.

With this approach I believe we could enumerate all possible valid documents up to a certain "size" with bounded amounts of entropy. We’re not quite there yet for at least a few reasons:

* All generated field selections get a unique alias, so [field merging](https://spec.graphql.org/draft/#sec-Field-Selection-Merging) is trivial. We can easily sometimes re-use response keys, but maintaining "fields can merge" validity is tricky.
* Every time the generator decides create a named fragment spread selection it creates a new named fragment definition. If re-using fragments we need to be careful not to introduce [cycles](https://spec.graphql.org/draft/#sec-Fragment-Spreads-Must-Not-Form-Cycles) but this may accidentally be already solved: for borrow-checking reasons we remove fragments from the map while modifying them, and this may be exactly the subset that would introduce a cycle if reused.
* String values are generated like names (limited character set, and typically short). I think this is ok for the purpose of testing query planner as string values in executable documents are typically not meaningful for planning purpose. (It’s a different story for generating subgraph schemas with federation directives)


SimonSapin:
This PR is draft because I’m unsure if this generator should live inside apollo-rs. I think eventually yes, but at first we’ll likely want to tweak it a lot and going through a crates.io release for every change would be a significant obstacle.

PR #881, opened 2024-07-15
Introduce a faster hasher

o0Ignition0o:
This changeset introduces ahash, a hasher that is faster than the one provided by the stdlib.

It exposes a collections::fast module, that can be used by other libraries that rely on apollo-rs types. This makes it easier for the apollo-rs to swap implementations in the future if needs be.

### DIFF
--- a/crates/apollo-compiler/src/collections.rs
+++ b/crates/apollo-compiler/src/collections.rs
@@ -1,9 +1,7 @@
-pub mod fast {
-    use indexmap::IndexMap as IM;
-    use indexmap::IndexSet as IS;
+use indexmap::IndexMap as IM;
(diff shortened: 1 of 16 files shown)


crates/apollo-compiler/src/collections.rs
```diff
@@ -0,0 +1,7 @@
+use indexmap::IndexMap as IM;
```
SimonSapin, line 1:
```suggestion
//! Type aliases for hashing-based collections configured with a specific hasher,
//! as used in various places thorough the API

use indexmap::IndexMap as IM;
```

SimonSapin (approved):

PR #883, opened 2024-07-19
fix(compiler): fix validation for undefined variables nested in values

goto-bus-stop:
Previously, an operation like this would pass validation:
```graphql
query {
  field(listArg: [$undefinedVar])
}
```
It turns out this is not very common as we haven't seen this cause a problem
on hundreds of thousands of operations. But it does happen :)

This moves the UndefinedVariable rule to value validation: we are already
recursing into the value here, and we're actually also already checking
that the variable exists, so this is a simpler place to do it than in
the `variable.rs` module.

This fixes the immediate problem. There are two things that I may be able to
address here before we merge it time willing...
- [ ] We are validating the type of variables passed directly to arguments in `variable.rs`, but the type of a variable used inside a nested value inside `value.rs`, and both ways are different? Is this correct according to spec or should it be changed?
- [ ] There is a TODO in this code that seems valid and this might be a nice time to do it

### DIFF
--- a/crates/apollo-compiler/test_data/diagnostics/0101_mismatched_variable_usage.txt
+++ b/crates/apollo-compiler/test_data/diagnostics/0101_mismatched_variable_usage.txt
@@ -64,4 +64,15 @@ Error: variable `$intArg` of type `Int` cannot be used for argument `nonNullIntA
     │                                              ──────┬─────  
     │                                                    ╰─────── variable `$intArg` of type `Int` is declared here
 ────╯
+Error: expected value of type Boolean, found a variable
+    ╭─[0101_mismatched_variable_usage.graphql:51:50]
+    │
+ 51 │     nonNullBooleanListField(nonNullBooleanListArg: [$intArg])
+    │                                                     ───┬───  
+    │                                                        ╰───── provided value is a variable
+    │ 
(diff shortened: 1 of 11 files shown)


crates/apollo-compiler/test_data/diagnostics/0101_mismatched_variable_usage.txt
```diff
@@ -64,4 +64,15 @@ Error: variable `$intArg` of type `Int` cannot be used for argument `nonNullIntA
...
+    ╭─[0101_mismatched_variable_usage.graphql:51:50]
+    │
+ 51 │     nonNullBooleanListField(nonNullBooleanListArg: [$intArg])
+    │                                                     ───┬───  
+    │                                                        ╰───── provided value is a variable
+    │ 
+ 63 │     nonNullBooleanListField(nonNullBooleanListArg: [Boolean]!): Boolean
+    │                                                    ─────┬────  
+    │                                                         ╰────── expected type declared here as Boolean
+────╯
```
goto-bus-stop, line 77: I added this to show that the error message is very different depending on if you use the variable directly or nested inside a value... I think they should be treated the same

SimonSapin (approved):

Issue #731, opened 2023-11-09
Many lexer errors do not have a location


goto-bus-stop:
## Description

Many errors emitted by the lexer do not set the `error.index` property, so they point to the start of the document by default.

### Steps to reproduce

Input a GraphQL document with unexpected characters:
```graphql
type Query {
  // Some invalid characters
  user: User
}
```

### Expected result

Syntax errors point to the correct position

### Actual result

```
rror: syntax error: Unexpected character "/"
   ╭─[q.graphql:1:1]
   │
 1 │ type Query {
   │ ┬
   │ ╰── Unexpected character "/"
───╯
Error: syntax error: Unexpected character "/"
   ╭─[q.graphql:1:1]
   │
 1 │ type Query {
   │ ┬
   │ ╰── Unexpected character "/"
───╯
```

SimonSapin:
I’d be in favor of eventually removing the `Error::new` constructor that sets the index to zero, making the index always provided by the caller.

PR #886, opened 2024-07-21
fix(parser): add missing location information for some lexer errors

PhoebeSzmucer:
Fixes https://github.com/apollographql/apollo-rs/issues/731

I _think_ these are all the missing locations, but there are a few more places that do `Error::new` - as far as I can tell though these errors are getting location information filled in at a later stage (via `self.done()`).

I saw the original issue proposed removing `Error::new` completely - I'm happy to do it, but it would be a break, and I'm not sure what the policy on those is in this repo.

### DIFF
--- a/crates/apollo-parser/src/lexer/mod.rs
+++ b/crates/apollo-parser/src/lexer/mod.rs
@@ -264,7 +264,11 @@ impl<'a> Cursor<'a> {
                         return self.done(token);
                     }
                     c if !c.is_ascii_hexdigit() => {
-                        self.add_err(Error::new("invalid unicode escape sequence", c.to_string()));
+                        self.add_err(Error::with_loc(
+                            "invalid unicode escape sequence",
+                            c.to_string(),
+                            0,
+                        ));
                         state = State::StringLiteral;
 
                         continue;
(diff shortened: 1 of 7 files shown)


crates/apollo-parser/src/lexer/mod.rs
```diff
@@ -254,12 +255,11 @@ impl<'a> Cursor<'a> {
...
                     '"' => {
-                        self.add_err(Error::new(
+                        return Err(Error::with_loc(
                             "incomplete unicode escape sequence",
-                            c.to_string(),
+                            self.current_str().to_string(),
+                            token.index,
                         ));
-                        token.data = self.current_str();
-                        return self.done(token);
```
goto-bus-stop, line 262: This is a slight behaviour change, because we would previously still add the string token (knowing it's invalid, but that is intentional in apollo-parser). Was there some issue with using `.add_err()` here?

goto-bus-stop (commented):
Thanks for the contribution!

goto-bus-stop:
Also, totally fine to remove `Error::new` entirely. We are still working on the apollo-rs v1 release and we are regularly making breaking changes in apollo-compiler until then.

crates/apollo-parser/src/lexer/mod.rs
PhoebeSzmucer: Ah yes you're right. Sorry, I didn't mean to change the behaviour, I just felt like this was more consistent with the rest of the code. I will revert this part.

I will also remove `Error::new`.

PhoebeSzmucer:
@goto-bus-stop I removed `Error::new` entirely. There is only `Error::with_loc` left - I wonder if it should become the new `Error::new` 🤔

goto-bus-stop (approved):
LGTM, thank you!

PR #896, opened 2024-08-20
fix(compiler): fix validation for fragment spread of interface without implementers

goto-bus-stop:
This `Intf` has no implementations. As written in the spec, doing a `... on Intf` fragment spread should never work, as the set of possible types is empty and can never intersect with the parent type. However, implementations like graphql-js and graphql-go have an early check, accepting the fragment if the type condition is equal to the parent type. This tests reproduces that.

We may want to align with graphql-js and graphql-go rather than the spec here for compatibility? Though it's not something that's likely to happen in the real world.

Ref https://github.com/graphql/graphql-spec/issues/1109

### DIFF
--- a/crates/apollo-compiler/test_data/diagnostics/0091_recursive_interface_definition.graphql
+++ b/crates/apollo-compiler/test_data/diagnostics/0091_recursive_interface_definition.graphql
@@ -4,13 +4,17 @@ interface A implements B {
 interface B implements A {
   name: String
 }
+type Impl implements A & B {
+  name: String
+}
+type Query {
+  get: A
+}
+
 fragment recursive on A {
   name
 }
 
-type Query {
-  get: A
-}
 query {
   get { ...recursive }
 }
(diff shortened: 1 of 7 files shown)


crates/apollo-compiler/test_data/diagnostics/0091_recursive_interface_definition.graphql
```diff
@@ -4,13 +4,17 @@ interface A implements B {
 interface B implements A {
   name: String
 }
-fragment recursive on A {
-  name
+type Impl implements A & B {
+  name: String
 }
```
goto-bus-stop, line 9: Unrelated change, just fixing the schema so it doesn't rely on having an interface with 0 implementers

duckki:
I agree that Apollo's validator is spec-compliant currently, while others are not :)

At the same time, this could be an oversight in the spec. It's good to be able to test queries before writing an implementation (despite the weirdness of conditioning on the parent type).

lrlna (approved):

PR #899, opened 2024-08-27
Fixes and additions to schema introspection, for comparative fuzzing

SimonSapin:
We fuzz many introspection queries then expect apollo-compiler and graphql-js to return equivalent responses. Some of these changes help writing the fuzz targets, others fix bugs found by fuzzing.


goto-bus-stop (approved):
lgtm

PR #900, opened 2024-08-27
refactor(parser): Use peek_token() to streamline peek operations

### LINKED ISSUE #830: Simplify `peek_token()` calls in the parser
We have a few places where we call `peek()` to check an upcoming token's kind, and then `peek_data().unwrap()` to check its value. For example, to identify what type of definition is coming up: `type`, `union`, or something else. This is one example:
https://github.com/apollographql/apollo-rs/blob/6c9adc4c076d23a83de740ac31717ca921ce6161/crates/apollo-parser/src/parser/grammar/object.rs#L27-L33

The `unwrap()` call is valid in those cases, but a bit icky. For it to be valid, the current token **must not** change between those calls, and this is not statically verifiable. A mistake in a refactor could disconnect the `peek()` and `peek_data()` calls and then the unwrap could panic.

I'd like to see these refactored to use `let Some(token) = peek_token()` where possible, and then use `token.kind` / `token.data` directly, so there is no way that it could ever panic.

surajk-m:
Closes #830 

This PR `refactors` parser to use `peek_token()` instead of separate `peek()` and `peek_data()` calls.

_Changes:_

- Replaced instances of `peek()` / `peek_data()` call pairs.
- Removed `unwrap()` by working directly with the `peek_token()`. This ensures that the `token's kind` and `data` are always in `sync` and that no intermediate changes can occur between the checks.

### DIFF
--- a/crates/apollo-parser/src/parser/grammar/name.rs
+++ b/crates/apollo-parser/src/parser/grammar/name.rs
@@ -8,14 +8,16 @@ use crate::S;
 /// *Name*:
 ///     [_A-Za-z][_0-9A-Za-z]
 pub(crate) fn name(p: &mut Parser) {
-    match p.peek() {
-        Some(TokenKind::Name) => {
+    if let Some(token) = p.peek_token() {
+        if token.kind() == TokenKind::Name {
+            let name_data = token.data();
             let _g = p.start_node(SyntaxKind::NAME);
-            validate_name(p.peek_data().unwrap(), p);
+            validate_name(name_data, p);
             p.bump(SyntaxKind::IDENT);
+            return;
         }
-        _ => p.err("expected a Name"),
     }
+    p.err("expected a Name");
 }
 
 pub(crate) fn validate_name(name: &str, p: &mut Parser) {
--- a/crates/apollo-parser/src/parser/grammar/document.rs
+++ b/crates/apollo-parser/src/parser/grammar/document.rs
@@ -36,12 +36,18 @@ pub(crate) fn document(p: &mut Parser) {
                 }
             }
             TokenKind::Name => {
-                let def = p.peek_data().unwrap();
-                select_definition(def, p);
+                if let Some(def) = p.peek_data() {
+                    select_definition(def, p);
+                } else {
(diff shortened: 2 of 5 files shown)


SimonSapin:
Hi, and thanks for submitting this PR!

Issue #830 specifically pertains to the pattern where `peek()` is followed by `peek_data().unwrap()`, it’s the `unwrap()` part and its potential panic that we try to avoid. Using `peek()` or `peek_data()` without `unwrap()` is fine and should not be replaced with more verbose code. 

There are a few other locations with the `peek_data().unwrap()` pattern is used:

* `crates/apollo-parser/src/parser/grammar/fragment.rs`
* `crates/apollo-parser/src/parser/grammar/document.rs`
* `crates/apollo-parser/src/parser/grammar/value.rs`
* `crates/apollo-parser/src/parser/grammar/name.rs`

Would you be able to update those locations as well and undo the cases where `peek_token()` isn't needed?

surajk-m:
Hi @SimonSapin,

Thanks for reviewing my PR! I'll revert the changes where `peek_token()` isn't needed and will update the remaining instances of `peek_data().unwrap()` pattern used in the mentioned files.

crates/apollo-parser/src/parser/grammar/name.rs
```diff
@@ -8,14 +8,16 @@ use crate::S;
...
+            let name_data = token.data();
             let _g = p.start_node(SyntaxKind::NAME);
-            validate_name(p.peek_data().unwrap(), p);
+            validate_name(name_data, p);
             p.bump(SyntaxKind::IDENT);
+            return;
         }
-        _ => p.err("expected a Name"),
     }
+    p.err("expected a Name");
```
lrlna:
```suggestion
    match p.peek_token() {
        Some(token) if token.kind() == TokenKind::Name => {
            let name_data = token.data();
            let _g = p.start_node(SyntaxKind::NAME);
            validate_name(name_data, p);
            p.bump(SyntaxKind::IDENT);
        }
        _ => p.err("expected a Name"),
    }
```

PR #901, opened 2024-08-28
fix: Parser fails to parse object type def with no fields

trevor-scheer:
According to the spec, a syntactically valid object type definition can have no fields (if the curlies are omitted).

See the second option:
https://spec.graphql.org/draft/#ObjectTypeDefinition

This removes the parse expectation that the only `TokenKind::Name` which can follow a type's name is `implements`, since it can also be the beginning of a following definition or extension.

Credit to @swcollard for spotting this bug during LSP QA!


crates/apollo-parser/test_data/parser/ok/0044_input_type_definition_without_fields.graphql
lrlna, line 1: the nittiest of nits: input type definition has "input values" not "fields", so this file should say `without_input_values` 🙈

lrlna (approved):
Can you also add diagnostics tests to `apollo-compiler` under `test_data/diagnostics` to make sure the compiler catches these types with no fields/input values and _no_ extensions? I'd want to make sure there is an diagnostic for cases like these:

```graphql
"A type with no fields"
type AnObjectTypeWithoutFields
```

trevor-scheer:
@lrlna it looks like these validation tests are already covered by 0114-0118. Those tests _do_ have extensions, but they don't contribute fields/input values so they seem more comprehensive than what I would've added. If you think it'd be beneficial to copy those with explicitly no empty extensions (like what you suggested verbatim) I'd be happy to add!

lrlna:
@trevor-scheer ah! i missed them. those cover the cases i was thinking about

lrlna (approved):

PR #908, opened 2024-09-09
An adopted schema extension disables implicit root operations + Prep releases

SimonSapin:
`adopt_orphan_extensions` is a non-standard mode so we don’t have a spec to precisely define its behavior, but I think this behavior is more consistent. See new test for an example.

Also prepare releases for:
* apollo-parser@0.8.2
* apollo-compiler@1.0.0-beta.22
* apollo-smitch@0.12.0 (only needed to update the pinned compiler dependency)

### DIFF
--- a/crates/apollo-compiler/tests/extensions.rs
+++ b/crates/apollo-compiler/tests/extensions.rs
@@ -93,6 +93,25 @@ fn test_invalid_orphan_extensions_schema_def_with_duplicate_root_operation() {
 }
 
 #[test]
+fn test_orphan_schema_extension_disables_implicit_root_types() {
+    let input = r#"
+        extend schema { query: Query }
+        type Query { viruses: [Virus] }
+        type Virus { mutations: [Mutation] }
+        type Mutation { something: String }
+    "#;
+
+    let schema = Schema::builder()
+        .adopt_orphan_extensions()
+        .parse(input, "schema.graphql")
+        .build()
+        .unwrap();
+
+    assert!(schema.schema_definition.mutation.is_none());
+    schema.validate().unwrap();
+}
+
+#[test]
 fn test_orphan_extensions_kind_mismatch() {
     let input = r#"
     extend type T @dir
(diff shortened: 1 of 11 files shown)


crates/apollo-parser/CHANGELOG.md
```diff
@@ -16,6 +16,18 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 
 ## Documentation -->
+
+# [0.8.2](https://crates.io/crates/apollo-parser/0.8.2) - 2024-09-09
+
+## Fixes
+- **Allow object type definition syntax without fields - [trevor-scheer], [pull/901].**
+  Object types must have at least one field during validation,
+  but at the syntax level field definitions can all be in `extend type` extensions,
+  with the main `type` definition not including `{}` curly braces or fields.
```
trevor-scheer, line 26: I think this description could be a bit more clear.

It is syntactically valid for an object type to have no fields when the `{}` curly braces are omitted. During validation, this is also valid as long as an extension for the type also exists which does include field definitions.

dariuszkuc (approved):

crates/apollo-compiler/tests/extensions.rs
```diff
@@ -92,6 +92,25 @@ fn test_invalid_orphan_extensions_schema_def_with_duplicate_root_operation() {
...
+        type Mutation { something: String }
+    "#;
+
+    let schema = Schema::builder()
+        .adopt_orphan_extensions()
+        .parse(input, "schema.graphql")
+        .build()
+        .unwrap();
+
+    assert!(schema.schema_definition.mutation.is_none());
```
trevor-scheer, line 110: So in this mode, if I've explicitly defined a root operation on an extension then we shouldn't implicitly provide the others, seems reasonable.

And in the case where we're in this mode, AND I've provided an extension, AND the extension's roots are empty, then we do still provide the implicit 👍

trevor-scheer (approved):

SimonSapin:
Accidentally merged, so I reverted main and moved to https://github.com/apollographql/apollo-rs/pull/909

PR #909, opened 2024-09-09
An adopted schema extension disables implicit root operations + Prep releases

SimonSapin:
`adopt_orphan_extensions` is a non-standard mode so we don’t have a spec to precisely define its behavior, but I think this behavior is more consistent. See new test for an example.

Also prepare releases for:
* apollo-parser@0.8.2
* apollo-compiler@1.0.0-beta.22
* apollo-smitch@0.12.0 (only needed to update the pinned compiler dependency)

### DIFF
--- a/crates/apollo-parser/CHANGELOG.md
+++ b/crates/apollo-parser/CHANGELOG.md
@@ -21,9 +21,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## Fixes
 - **Allow object type definition syntax without fields - [trevor-scheer], [pull/901].**
-  Object types must have at least one field during validation,
-  but at the syntax level field definitions can all be in `extend type` extensions,
-  with the main `type` definition not including `{}` curly braces or fields.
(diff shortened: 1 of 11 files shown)


crates/apollo-parser/CHANGELOG.md
```diff
@@ -16,6 +16,18 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## Maintenance
 
 ## Documentation -->
+
+# [0.8.2](https://crates.io/crates/apollo-parser/0.8.2) - 2024-09-09
+
+## Fixes
+- **Allow object type definition syntax without fields - [trevor-scheer], [pull/901].**
```
trevor-scheer, line 23: Copied from previous PR:

I think this description could be a bit more clear.

It is syntactically valid for an object type to have no fields when the {} curly braces are omitted. During validation, this is also valid as long as an extension for the type also exists which does include field definitions.

trevor-scheer (approved):

PR #910, opened 2024-09-16
feat(compiler): add compact Debug impl for LineColumn

goto-bus-stop:
This felt nicer in a case like the below:

https://github.com/apollographql/router/blob/e9dc61255e87294f624a0ecc68ac7c83fab5a7f0/apollo-federation/src/sources/connect/validation/snapshots/validation_tests%40rest.graphql.snap#L10-L25

It's not a big deal though. If anyone isn't that enthusiastic about it
we can just not :)


SimonSapin (approved):
I like it!

PR #911, opened 2024-09-16
Validation adds/removes built-in scalar defs based on usage

SimonSapin:
We choose to reflect in the Rust API the behavior expected of introspection:

https://spec.graphql.org/draft/#sec-Scalars.Built-in-Scalars
> When returning the set of types from the `__Schema` introspection type, all referenced built-in scalars must be included. If a built-in scalar type is not referenced anywhere in a schema (there is no field, argument, or input field of that type) then it must not be included.

Because our public API allows arbitrary mutation of a Rust `Schema`, we need a full schema scan at some point to determine usage. Validation is a good fit as it already does this scan.

In [the full GraphQL grammar](https://spec.graphql.org/draft/#sec-Document-Syntax)
the `NamedType` rule is used in:

* `TypeCondition`: not in a schema
* `RootOperationDefinition`, `UnionMemberTypes`: must refer to an object type
* `ImplementsInterfaces`: must refer to an interface type, not a scalar
* `Type` / `NonNullType`:
  * … in `VariableDefinition`: not in a schema
  * … in `FieldDefinition`: object/interface field "return" type, **can be a scalar**
  * … in `InputValueDefinition`: **can be a scalar**
    * … in `InputFieldsDefinition` (fields of input types)
    * … in `ArgumentsDefinition` for directive arguments or object/interface field arguments

Based on this, I think `validate_input_value_definitions` and `validate_field_definitions` are the only two places where validation code needs to record potential references to built-in scalar types.

### DIFF
--- a/crates/apollo-compiler/test_data/introspection/response_full.json
+++ b/crates/apollo-compiler/test_data/introspection/response_full.json
@@ -926,16 +926,6 @@
         },
         {
           "kind": "SCALAR",
-          "name": "Float",
-          "description": "The `Float` scalar type represents signed double-precision fractional values as specified by [IEEE 754](https://en.wikipedia.org/wiki/IEEE_floating_point).",
-          "fields": null,
-          "inputFields": null,
-          "interfaces": null,
-          "enumValues": null,
-          "possibleTypes": null
-        },
-        {
-          "kind": "SCALAR",
           "name": "String",
           "description": "The `String` scalar type represents textual data, represented as UTF-8 character sequences. The String type is most often used by GraphQL to represent free-form human-readable text.",
           "fields": null,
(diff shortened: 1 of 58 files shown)


crates/apollo-compiler/test_data/introspection/response_full.json
```diff
@@ -924,16 +924,6 @@
...
-        {
-          "kind": "SCALAR",
-          "name": "Float",
-          "description": "The `Float` scalar type represents signed double-precision fractional values as specified by [IEEE 754](https://en.wikipedia.org/wiki/IEEE_floating_point).",
-          "fields": null,
-          "inputFields": null,
-          "interfaces": null,
-          "enumValues": null,
-          "possibleTypes": null
-        },
```
lrlna, line 936: Would have imagined a Float _was used_ in this schema, but I guess not?

lrlna (approved):

goto-bus-stop (approved):

PR #912, opened 2024-09-16
Fix value locations for object field values and add a couple location-specific tests

dylan-apollo:
Previously, object field values would include the name of the field in the span, so in the example test, the span was for `field: "value"` rather than just `"value"`

The tests currently fail because some diagnostics need to point at the whole field, not just the value of it.


dylan-apollo:
I'm not sure what the right API is to allow accessing the entire field's (not just the value's) location, as is wanted in some diagnostics.

SimonSapin (approved):
I think the change for AST construction is correct. Validation could recompose the span for the name/value pair but I'm not even sure it should: the error is in the value('s type) only

Discussed this with @goto-bus-stop who agreed

Issue #916, opened 2024-09-26
misleading validation error message in selections under a String


Geal:
Found when testing something else: https://github.com/apollographql/router/pull/6035#discussion_r1768865666

With the query `query Invalid { me { name { reviews { body } } } }` where `name` is a String, we get the following errors:

```json
{
  "errors":[
    {
      "message":"Field \"me\" of type \"Query\" must have a selection of subfields. Did you mean \"me { ... }\"?",
      "locations":[{"line":1,"column":17}],
      "extensions":{"code":"GRAPHQL_VALIDATION_FAILED"}},
    {
       "message":"Field \"name\" must not have a selection since type \"String\" has no subfields",
       "locations":[{"line":1,"column":22}],
       "extensions":{"code":"GRAPHQL_VALIDATION_FAILED"}
    }
  ]
}
```

The second error is expected, but why would we get the first one? Is that because the compiler considers the entire `name` selection invalid, removes it from its view, and then the `me` selection appears empty?

goto-bus-stop:
Repro in https://github.com/apollographql/apollo-rs/pull/917

PR #917, opened 2024-09-26
Add failing test for #916

goto-bus-stop:
It also reports this, which doesn't make sense:
```
Error: interface, union and object types must have a subselection set
   ╭─[0119_nested_subselection_issue_916.graphql:8:17]
   │
 8 │ query Invalid { me { name { reviews { body } } } }
   │                 ────────────────┬───────────────
   │                                 ╰───────────────── `Query.me` is an object type `User` and must select fields
───╯
```


goto-bus-stop:
The reason appears to be that the `executable::Field`'s subselection gets emptied during ast -> executable conversion

PR #922, opened 2024-10-29
Silence clippy::result_large_err warning

SimonSapin:
This lint seems to be more aggressive in Rust 1.82 than in previous versions as it now warns about functions returning `Result<Schema, WithErrors<Schema>>` or `Result<ExecutableDocument, WithErrors<ExecutableDocument>>`.

The lint is correct that a return value of e.g. 224 bytes has some cost, but parsing and validation functions are typically not called very often so we accept this cost rather than make the API more awkward by boxing schemas and documents.


goto-bus-stop (approved):
Thank youu

Issue #852, opened 2024-04-09
Validate that `Value::Variable` is not used in const locations


SimonSapin:
In the GraphQL spec, [the `Value` grammar](https://spec.graphql.org/October2021/#sec-Input-Values) has a `Const` parameter that disallows referencing a variable in type system locations, where variables don’t exist. As of apollo-compiler 1.0.0-beta.15, we have a single `Value` Rust enum with a `Variable` variant. Since https://github.com/apollographql/apollo-rs/pull/777 apollo-parser correctly emits parse errors when a variable is used in a const context. But since apollo-compiler data structures are mutable, it is still possible construct a wrong value programatically.

These is [discussion of encoding constness in the type system](https://github.com/apollographql/apollo-rs/issues/538) but that may or may not be the best way forward. Until that happens (if it does), validation should emit a diagnostic if `Value::Variable` is use incorrectly. This is in the same spirit as having both https://github.com/apollographql/apollo-rs/pull/847 and https://github.com/apollographql/apollo-rs/pull/845.

PR #925, opened 2024-11-04
Make variables in const contexts validation errors

SimonSapin:
Fixes https://github.com/apollographql/apollo-rs/issues/852

GraphQL values can be of different kinds, including variables with the `$x` syntax. Variables are not allowed in some contexts, such as schemas. The parser already rejects them as syntax errors, but the Rust API allows mutating a document and representing variable values in contexts where they are not allowed.

It was already the case that would usually cause a validation error like ``variable `$x` is not defined``. This PR ensures and tests that this is the case for every relevant context. It also makes related drive-by fixes:

* Validate the default value of input fields and arguments
* Validate the default value of variables once at their definition, not at every usage of the variable. This fixes duplicate diagnostics for a default value of incorrect type for a variable correctly used multiple times.


goto-bus-stop (approved):
great!

Issue #928, opened 2024-11-25
Default value validation is a breaking change, do it again in compiler 2.0


SimonSapin:
PR https://github.com/apollographql/apollo-rs/pull/925 includes a change to start  validating the default value of input fields and arguments. This turns out to break enough existing schemas (Router rejects them as invalid where it previously accepted them) that we consider it a breaking change. After some discussion, our plan is:

1. Revert default value validation for now
2. Publish a new compiler beta without it to unblock current work
3. Finish up compiler 1.0 stable and use it in Router before its 1.x LTS
4. Stricter validation, which we consider a breaking change, will be in compiler 2.0which will be used by router 2.0. _**This is the part tracked by this issue**_
5. If needed later, we can backport LTS fixes to compiler 1.x

SimonSapin:
For reference, graphql-js only recently started validating default input values in https://github.com/graphql/graphql-js/pull/3814. Version 16.x used by router-bridge does not.

SimonSapin:
Revert PR https://github.com/apollographql/apollo-rs/pull/929

tninesling:
Resolved by #1082

PR #929, opened 2024-11-25
Revert default value validation of input fields and arguments

SimonSapin:
This turns out to break enough existing schemas (Router rejects them as invalid where it previously accepted them) that we consider it a breaking change.

See https://github.com/apollographql/apollo-rs/issues/928

### DIFF
--- a/crates/apollo-compiler/CHANGELOG.md
+++ b/crates/apollo-compiler/CHANGELOG.md
@@ -21,11 +21,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## Fixes
 - **Validate against reserved names starting with `__` in schemas - [SimonSapin], [pull/923].**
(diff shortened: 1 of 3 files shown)


crates/apollo-compiler/CHANGELOG.md
```diff
@@ -21,7 +21,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## Fixes
 - **Validate against reserved names starting with `__` in schemas - [SimonSapin], [pull/923].**
-- **Validate the default value of input fields and arguments - [SimonSapin], [pull/925].**
+- **Fix duplicate diagnostic for variable with invalid default value - [SimonSapin], [pull/925].**
```
goto-bus-stop, line 24: maybe add this PR to the changelog entry too? if someone looks at 925 alone they might get the wrong impression

goto-bus-stop (approved):
👍🏻

PR #931, opened 2024-11-29
chore: Retire Projects Tagger

abernix:
We no longer necessitate triggering an action that adds issues to Project boards, and we're sunsetting the endpoint that we were previously calling.

Just like I removed it in https://github.com/apollographql/router/pull/6113


goto-bus-stop (approved):

abernix:
I think the lint failure must be unrelated
to this PR.

SimonSapin:
Indeed, I’ll make another PR for this

PR #942, opened 2025-01-09
API docs and stabilization pass

SimonSapin:
I went through the entire public API of apollo-compiler (except the `excecution` module which will get special attention soon) to add documentation and make various tweaks, with the goal of soon releasing a semver-stable version.

Overall I’m fairly happy with where we were already! In terms of API "shape"

[ROUTER-953]: https://apollographql.atlassian.net/browse/ROUTER-953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/apollo-compiler/src/schema/mod.rs
```diff
@@ -1,4 +1,48 @@
...
+//!   (For example, some directives can be applied to an object type extensions to affect
+//!   fields defined in the same extension but not other fields of the object type.)
+//!   See [`Component`].
+//!
+//! [type system extensions]: https://spec.graphql.org/draft/#sec-Type-System-Extensions
+//!
+//! In some cases like [`SchemaDefinition`], this module and the [`ast`] module
+//! define different Rust types with the same names.
+//! In other cases like [`Directive`] there is no data structure difference needed,
+//! so this modules reuses and publicly re-exports some Rust types from the [`ast`] module.
```
goto-bus-stop, line 23:
```suggestion
//! so this module reuses and publicly re-exports some Rust types from the [`ast`] module.
```

crates/apollo-compiler/src/coordinate.rs
```diff
@@ -30,6 +32,16 @@ use std::str::FromStr;
 /// assert_eq!(coord!(Type.field(arg:)).to_string(), "Type.field(arg:)");
 /// assert_eq!(coord!(EnumType.ENUM_VALUE).to_string(), "EnumType.ENUM_VALUE");
 /// ```
+///
+/// All possible return types of this macro implement
+/// [`From`]`<`[`SchemaCoordinate`]`>` so they can be converted `.into()`:
```
goto-bus-stop, line 37:
```suggestion
/// [`From`]`<`[`SchemaCoordinate`]`>` so they can be converted using `.into()`:
```

goto-bus-stop (approved):
I haven't checked that it's exhaustive, but I have checked that the docs changes are great 😁 🥳

PR #946, opened 2025-01-14
Prepare stable release of apollo-compiler@1.25.0 (and apollo-smith@0.15.0)

SimonSapin:


SimonSapin:
I have a branch of Apollo Router that compiles and passes test with this new apollo-compiler version

lrlna (approved):
HELL YA!

PR #947, opened 2025-01-16
Prepare apollo-parser@0.8.4

SimonSapin:
The rowan update was not on crates.io yet

### DIFF
--- a/crates/apollo-parser/CHANGELOG.md
+++ b/crates/apollo-parser/CHANGELOG.md
@@ -17,6 +17,14 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## Documentation -->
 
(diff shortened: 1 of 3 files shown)


goto-bus-stop:
shld we update the declared dependency in apollo-compiler/Cargo.toml too? compiler *must* be used with a new apollo-parser version to work, right?

goto-bus-stop (approved):
please see note ^ but we can go ahead with or without

SimonSapin:
Compiler uses `rowan::TextRange` inside `SourceSpan` but it’s entirely separate from parser’s use of rowan. The negative consequence of unmatched versions is only increased compile time.

SimonSapin:
It wouldn’t be difficult to make compiler use a custom `struct { start_offset: u32, end_offset: u32 }` instead and remove the dependency

goto-bus-stop:
ahh okay great!

PR #953, opened 2025-03-04
Add WebAssembly demo

SimonSapin:
image

This replace compiling the library with wasm target on CI, which recently started failing with:

```
error: the wasm*-unknown-unknown targets are not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support
   --> /home/circleci/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.15/src/lib.rs:342:9
    |
342 | /         compile_error!("the wasm*-unknown-unknown targets are not supported by \
343 | |                         default, you may need to enable the \"js\" feature. \
344 | |                         For more information see: \
345 | |                         https://docs.rs/getrandom/#webassembly-support");
    | |________________________________________________________________________^

error[E0433]: failed to resolve: use of undeclared crate or module `imp`
   --> /home/circleci/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.15/src/lib.rs:398:9
    |
398 |         imp::getrandom_inner(dest)?;
    |         ^^^ use of undeclared crate or module `imp`
```

The only available fix is to enable the `js` feature when compiling getrandom to wasm, but https://docs.rs/getrandom/0.2.15/getrandom/index.html#webassembly-support recommends specifically _against_ doing so in a library crate. The new demo therefore serves as a non-library crate to enable it, as well as a more end-to-end test (if manual) than just compiling the library.

Randomness is used by the `ahash` crate as a DoS mitigation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,8 +106,11 @@ jobs:
           rust_channel: << parameters.rust_channel >>
           platform: << parameters.platform >>
       - run:
-          name: Build to wasm target
-          command: rustup target add wasm32-unknown-unknown && cargo build --package apollo-parser --package apollo-compiler --target wasm32-unknown-unknown
+          name: Install wasm-pack
+          command: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+      - run:
+          name: Build wasm demo
+          command: wasm-pack build --target web examples/validation-wasm-demo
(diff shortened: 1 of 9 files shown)


lrlna (approved):
Thank **you**!

PR #960, opened 2025-03-18
change type signature for impl ariadne::Cache

lrlna:
fn fetch() and fn display() changed signatures in the latest patch release of ariadne (0.51.0).

* `fn fetch(&mut self, file_id: &FileId) -> Result<&ariadne::Source, Box<dyn fmt::Debug + '_>>` --> `fn fetch(&mut self, file_id: &FileId) -> Result<&ariadne::Source, impl fmt::Debug>`
* `fn display<'a>(&self, file_id: &'a FileId) -> Option<Box<dyn fmt::Display + 'a>> ` --> `fn display<'a>(&self, file_id: &'a FileId) -> Option<impl fmt::Display + 'a>`

### DIFF
--- a/crates/apollo-compiler/src/diagnostic.rs
+++ b/crates/apollo-compiler/src/diagnostic.rs
@@ -296,7 +296,7 @@ impl ariadne::Cache<FileId> for Cache<'_> {
             static EMPTY: OnceLock<ariadne::Source> = OnceLock::new();
             Ok(EMPTY.get_or_init(|| ariadne::Source::from(String::new())))
         } else {
-            Err(Box::new(NotFound(*file_id)))
+            Err(NotFound(*file_id))
         }
     }
 
@@ -316,9 +316,9 @@ impl ariadne::Cache<FileId> for Cache<'_> {
 
         if *file_id != FileId::NONE {
             let source_file = self.0.get(file_id)?;
-            Some(Box::new(Path::SourceFile(source_file.clone())))
+            Some(Path::SourceFile(source_file.clone()))
         } else {
-            Some(Box::new(Path::NoSourceFile))
+            Some(Path::NoSourceFile)
         }
     }
 }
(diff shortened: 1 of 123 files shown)


crates/apollo-compiler/src/diagnostic.rs
```diff
@@ -300,24 +300,25 @@ impl ariadne::Cache<FileId> for Cache<'_> {
...
-                    self.0.path().display().fmt(f)
+    fn display<'a>(&self, file_id: &'a FileId) -> Option<impl fmt::Display + 'a> {
+        enum Path {
+            SourceFile(Arc<SourceFile>),
+            NoSourceFile,
+        }
+        impl fmt::Display for Path {
+            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+                match self {
+                    Path::SourceFile(source_file) => source_file.path().display().fmt(f),
```
lrlna, line 311: this is adding an extra white space around path for some reason 🤔 
so instead of this:
```
Error: `typeFragment1` contains too much nesting
    ╭─[overflow.graphql:11:11]
```

we get this:
```
Error: `typeFragment1` contains too much nesting
    ╭─[ overflow.graphql:11:11 ]
```

SimonSapin: I think this is another change in Ariadne. We changed snapshots in router: https://github.com/apollographql/router/pull/7012

crates/apollo-compiler/src/diagnostic.rs
```diff
@@ -300,24 +300,25 @@ impl ariadne::Cache<FileId> for Cache<'_> {
...
+                    Path::NoSourceFile => f.write_str("(no source file)"),
                 }
             }
+        }
+
+        if *file_id != FileId::NONE {
             let source_file = self.0.get(file_id)?;
-            Some(Box::new(Path(source_file.clone())))
+            Some(Box::new(Path::SourceFile(source_file.clone())))
```
goto-bus-stop, line 319:
```suggestion
            Some(Path::SourceFile(source_file.clone()))
```
we probably don't need to box this anymore which is nice

lrlna: just spotted that this morning too. removed the Box in fn fetch as well 🙏

goto-bus-stop (approved):
\ 😁 /

PR #963, opened 2025-04-03
fix: validate subscription with conditional selections

dariuszkuc:
Adds validation for subscription operations that specify `@skip`/`@include` conditionals on root selections.

See: https://github.com/graphql/graphql-spec/pull/860


SimonSapin (commented):
Should this PR wait until https://github.com/graphql/graphql-spec/pull/860 is merged?

crates/apollo-compiler/tests/validation/operation.rs
```diff
@@ -218,3 +218,99 @@ type Product {
         "{errors}"
     );
 }
+
+#[test]
+fn it_validates_subscription_cannot_specify_multiple_fields() {
```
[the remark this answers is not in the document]
dariuszkuc: was wondering where are all the validation tests

dariuszkuc:
> Should this PR wait until [graphql/graphql-spec#860](https://github.com/graphql/graphql-spec/pull/860) is merged?

`graphql-js` PR was already merged so I'd argue that we should merge this ASAP.

crates/apollo-compiler/src/validation/operation.rs
```diff
@@ -46,6 +46,24 @@ pub(crate) fn validate_subscription(
                 },
             );
         }
+
+        // first rule validates that we only have single selection
+        let has_conditional_selection =
+            operation.selection_set.selections.iter().find(|selection| {
```
SimonSapin: Sorry for the back-and-forth, but now this lacks spec algorithm’s recursion and fails to catch something like `subscription { ... { field @skip } }`. I’ll move the check into `expand_selections`

crates/apollo-compiler/src/validation/selection.rs
```diff
@@ -67,13 +67,28 @@ impl<'a> FieldSelection<'a> {
...
+                    .find(|d| matches!(d.name.as_str(), "skip" | "include"))
+                {
+                    diagnostics.push(
+                        conditional_directive.location(),
+                        executable::BuildError::SubscriptionUsesConditionalSelection {
+                            name: operation.name.clone(),
+                        },
+                    );
+                }
+            }
```
goto-bus-stop, line 91: I don't think this should be part of `expand_selections`. That function is for expanding selections, not for doing validation. We can just check this in `validate_subscription` with the list returned by `expand_selections`

crates/apollo-compiler/src/validation/operation.rs
```diff
@@ -1,51 +1,117 @@
...
+    document: &'doc ExecutableDocument,
+    selections: &'doc executable::SelectionSet,
+    mut f: impl FnMut(&'doc executable::Selection),
+) -> Result<(), RecursionLimitError> {
+    fn walk_selections_inner<'doc>(
+        document: &'doc ExecutableDocument,
+        selection_set: &'doc executable::SelectionSet,
+        seen: &mut HashSet<&'doc Name>,
+        f: &mut dyn FnMut(&'doc executable::Selection),
+    ) -> Result<(), RecursionLimitError> {
```
goto-bus-stop, line 27: The implementation here doesn't actually raise a RecursionLimitError right now, but that will be part of a follow-up PR

goto-bus-stop (approved):

SimonSapin (approved):

PR #966, opened 2025-04-15
Harden stack overflow protection

goto-bus-stop:
This adds stack overflow protection to the existing "Variable Is Used" and "Fragment Is Used" validations, and the newly introduced subscription validation #963. Additionally, this significantly improves performance of the "Fragment Is Used" validation, which used to walk almost the entire document AST for each fragment definition, and now only does it once.

You need *a lot* of fragments to cause these to overflow, so it's only a problem for users who don't have a token limit configured or have it set much higher than the Router default.

### DIFF
--- a/crates/apollo-compiler/src/validation/mod.rs
+++ b/crates/apollo-compiler/src/validation/mod.rs
@@ -1089,6 +1089,60 @@ const DEFAULT_RECURSION_LIMIT: usize = 32;
 #[non_exhaustive]
 struct RecursionLimitError {}
 
+/// Track recursion depth to prevent stack overflow.
+#[derive(Debug)]
+struct DepthCounter {
+    value: usize,
+    high: usize,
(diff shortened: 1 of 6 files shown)


crates/apollo-compiler/src/validation/mod.rs
```diff
@@ -1089,6 +1089,60 @@ const DEFAULT_RECURSION_LIMIT: usize = 32;
 #[non_exhaustive]
 struct RecursionLimitError {}
 
+/// Track recursion depth to prevent stack overflow.
+#[derive(Debug)]
+struct DepthCounter {
```
goto-bus-stop, line 1094: This struct really is just an integer with extra steps, but I wanted to mimick the pattern that we have with RecursionStack in the other recursion-limited functions.

SimonSapin (approved):

PR #967, opened 2025-04-18
Simplify directive mutation

tninesling:
Implements mutable versions of some functions used to access directives and their arguments. This improves the ergonomics for some of the composition work we're doing in https://github.com/apollographql/router/pull/7235.

### DIFF
--- a/crates/apollo-compiler/src/ast/impls.rs
+++ b/crates/apollo-compiler/src/ast/impls.rs
@@ -680,6 +680,8 @@ impl Directive {
     }
 
     /// Returns the value of the argument named `name`, as specified in the directive application.
+    /// If there are other [`Node`] pointers to the same argument, this method will clone the
+    /// argument using [`Node::make_mut`].
     ///
     /// Returns `None` if the directive application does not specify this argument.
     ///
--- a/crates/apollo-compiler/src/ast/impls.rs
+++ b/crates/apollo-compiler/src/ast/impls.rs
@@ -468,6 +468,13 @@ impl DirectiveDefinition {
         self.arguments.iter().find(|argument| argument.name == name)
     }
 
+    /// Returns the definition of an argument by a given name.
+    pub fn argument_by_name_mut(&mut self, name: &str) -> Option<&mut Node<InputValueDefinition>> {
+        self.arguments
+            .iter_mut()
+            .find(|argument| argument.name == name)
+    }
+
     serialize_method!();
 }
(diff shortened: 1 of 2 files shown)


crates/apollo-compiler/src/node.rs
```diff
@@ -144,6 +144,12 @@ impl<T: ?Sized> std::ops::Deref for Node<T> {
     }
 }
 
+impl<T: Clone> std::ops::DerefMut for Node<T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.make_mut()
+    }
+}
```
SimonSapin, line 151: The rest of this PR looks great but this change feels a bit "too magical" since it can make it invisible when a subtree stops being shared. Maybe that’s not so bad: our nodes don’t use interior mutability so sharing v.s. not wouldn’t change behavior and is "only" a performance concern. But it’s a departure from usual Rust conventions.

@goto-bus-stop what do you think?

tninesling: That's fair. I think this probably does obfuscate the copy-on-write behavior a bit too much. This is trying to resolve some ergonomic issues we're having when we try to mutate parts of the schema, but it probably isn't the right solution. When we get a handle to some field, and we want to mutate something inside, it becomes `.make_mut()` all the way down.

We're specifically trying to mutate schemas in-place, so I am slightly worried about the possibility that we run into a case where we try to make some update and an unexpected clone causes that not to stick (admittedly, hiding the `make_mut` in this way just makes that worse). Is there a way for us to get a non-Arc-ed schema expressly for in-place mutation and then freeze it afterwards?

crates/apollo-compiler/src/ast/impls.rs
```diff
@@ -551,6 +569,14 @@ impl DirectiveList {
         self.get_all(name).next()
     }
 
+    /// Returns the first mutable directive with the given name, if any.
+    ///
+    /// This method is best for non-repeatable directives.
+    /// See also [`get_all_mut`][Self::get_all_mut] for repeatable directives.
+    pub fn get_mut(&mut self, name: &str) -> Option<&mut Node<Directive>> {
```
tninesling, line 576: I realized this is technically changing the public API because `DirectiveList` implements `DerefMut` for the underlying `Vec`, which has a slightly different signature than this one.

goto-bus-stop (approved):

PR #974, opened 2025-07-02
feat(compiler, parser): support executable document descriptions

goto-bus-stop:
Support for parsing and serializing descriptions on:
- (longhand) operation definitions
- fragment definitions
- variable definitions

As described by the latest GraphQL spec draft: https://github.com/graphql/graphql-spec/pull/1170

Adding a description on a query shorthand (bare braced selection set) raises a parse error.

This is a breaking change for apollo-compiler due to the addition of new fields:
- `apollo_compiler::ast::OperationDefinition::description`
- `apollo_compiler::ast::FragmentDefinition::description`
- `apollo_compiler::ast::VariableDefinition::description`
- `apollo_compiler::executable::Operation::description`
- `apollo_compiler::executable::Fragment::description`

For most users, adding `description: None` anywhere they create one of these structures is good enough.

**Todos**
- [ ] Formatting improvements for variable definitions with descriptions (steal what we do for argument definitions)


crates/apollo-compiler/test_data/serializer/diagnostics/0125_query_shorthand_description.graphql
```diff
@@ -0,0 +1,14 @@
+{
```
lrlna, line 1: because this is in `diagnostics`, i am assuming you wanted to test the shorthand with a description, which is not valid with the new spec. i think this is missing that description 😅

crates/apollo-compiler/test_data/diagnostics/0125_query_shorthand_description.graphql
lrlna, line 1: is this missing a matching .txt file?

crates/apollo-compiler/test_data/serializer/ok/0118_executable_descriptions.graphql
lrlna, line 1: should we also have this test case in the parser's test_data directory for posterity?

crates/apollo-compiler/test_data/diagnostics/0125_query_shorthand_description.graphql
lrlna: also, should we have this test case in the parser's test_data directory for posterity?

lrlna (dismissed):
This looks good! 

The only thing that needs to change is that the new file in `apollo-compiler/test_data/diagnostics` is missing it's matching `.txt` pair. Other comments are clarification questions.

Since this is a breaking change for the compiler, do we need to have a conversation about what should go into the next major version of the compiler (if there is anything)?

crates/apollo-compiler/test_data/serializer/diagnostics/0125_query_shorthand_description.graphql
goto-bus-stop: That's expected: this is the output of reserializing the document, and the serializer should not output invalid GraphQL (so it omits the description)

tninesling (approved):

PR #981, opened 2025-07-29
feat(smith): Implement builder for arbitrary responses

tninesling:
Adds a new `ResponseBuilder` to the public API for `apollo-smith`. The builder takes an instance of `arbitrary::Unstructured`, an `ExecutableDocument`, and a `Schema`. The resulting response object has a top-level `data` key which maps to the response in the expected shape.

This adds a new dependency on `serde_json_bytes`, since that is what Router expects for serialization.

### DIFF
--- a/crates/apollo-compiler/Cargo.toml
+++ b/crates/apollo-compiler/Cargo.toml
@@ -31,7 +31,7 @@ typed-arena = "2.0"
 
 [dev-dependencies]
 anyhow = "1.0"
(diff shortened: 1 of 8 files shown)


duckki (approved):
looks great. Thanks!

lrlna (changes requested):
Could I ask you to add an example of how someone would use this new API to the README and/or to the examples dir?

PR #987, opened 2025-08-08
fix: fix handling of directive arguments with nested types

dariuszkuc:
`FindRecursiveDirective` was only checking whether for recursion chains on directive names (i.e. going from `@custom -> args -> directive on args -> @custom`) but was not handling recursion on the input types.

### DIFF
--- a/crates/apollo-compiler/src/validation/directive.rs
+++ b/crates/apollo-compiler/src/validation/directive.rs
@@ -66,7 +66,12 @@ impl FindRecursiveDirective<'_> {
 
         let type_name = input_value.ty.inner_named_type();
         if let Some(type_def) = self.schema.types.get(type_name) {
-            self.type_definition(seen, type_def)?;
+            if seen.contains(type_def.name()) {
+                // input type was already processed
+                return Ok(());
+            }
+            let mut new_guard = seen.push(type_def.name())?;
+            self.type_definition(&mut new_guard, type_def)?;
         }
 
         Ok(())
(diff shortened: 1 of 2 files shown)


crates/apollo-compiler/src/validation/directive.rs
```diff
@@ -66,7 +66,16 @@ impl FindRecursiveDirective<'_> {
 
         let type_name = input_value.ty.inner_named_type();
         if let Some(type_def) = self.schema.types.get(type_name) {
-            self.type_definition(seen, type_def)?;
+            if seen.contains(type_def.name()) {
```
duckki, line 69: Now, the RecursionGuard holds both directive names and type names. But, since the type namespace and directive namespace are separate, collision is possible. We can mangle names to distinguish them.
```suggestion
            let mangled_type_name = Name::new_unchecked(&format!("!type_name!{}", type_def.name()));
            if seen.contains(&mangled_type_name) {
```

duckki (commented):
Looks good to me. It makes sense to guard against both directive recursion and type recursion.

crates/apollo-compiler/src/validation/directive.rs
goto-bus-stop: I think I'd prefer two separate `RecursionGuard`s over name mangling

crates/apollo-compiler/tests/validation/recursion.rs
```diff
@@ -347,3 +349,28 @@ type Query {
     "#]]
     .assert_eq(&errors.to_string());
 }
+
+#[test]
+fn handles_directive_with_nested_input_types() {
```
lrlna, line 354: I was fixing up the changelog for 1.30.0 and when looking over this PR's description I couldn't figure out whether this test was incorrect, or I don't understand the description. Isn't the point that we safe guard from recursive argument types in directive definitions? If so, shouldn't this test return an error?

PR #990, opened 2025-08-14
feat(compiler): allow SDL to contain built-in scalar definitions

dariuszkuc:
Adds new `ignore_builtin_redefinitions` option on the `SchemaBuilder` to allow SDL to contain built-in scalar re-definitions. Re-definitions are going to be effectively ignored and compiler will continue to use built-in GraphQL spec definitions. While [GraphQL spec](https://spec.graphql.org/October2021/#sec-Scalars) explicitly specifies following

```
When representing a GraphQL schema using the type system definition language, all built-in scalars must be omitted for brevity.
```

`graphql-js` reference implementation does not enforce this and does not report any errors even if built-in scalars are present in SDL. This change allows users to skip this validation to match `graphql-js` behavior.

### DIFF
--- a/crates/apollo-compiler/CHANGELOG.md
+++ b/crates/apollo-compiler/CHANGELOG.md
@@ -21,7 +21,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## Features
 
-- **Adds `allow_builtin_redefinitions` option to `SchemaBuilder` to allow SDL to contain built-in
+- **Adds `ignore_builtin_redefinitions` option to `SchemaBuilder` to allow SDL to contain built-in
   scalar definitions - [dariuszkuc], [pull/990]**
 
 
--- a/crates/apollo-compiler/CHANGELOG.md
+++ b/crates/apollo-compiler/CHANGELOG.md
@@ -17,6 +17,14 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## Maintenance
 ## Documentation-->
 
+# [x.x.x] (unreleased) - 2025-mm-dd
+
+## Features
+
+- **Adds `allow_builtin_redefinitions` option to `SchemaBuilder` to allow SDL to contain built-in
+  scalar definitions- [dariuszkuc], [pull/990]**
+
+
 # [1.29.0](https://crates.io/crates/apollo-compiler/1.29.0) - 2025-08-08
 
 ## Features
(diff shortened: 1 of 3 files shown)


duckki (approved):
looks good to me!

crates/apollo-compiler/CHANGELOG.md
```diff
@@ -17,6 +17,14 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## Maintenance
 ## Documentation-->
 
+# [x.x.x] (unreleased) - 2025-mm-dd
+
+## Features
+
+- **Adds `allow_builtin_redefinitions` option to `SchemaBuilder` to allow SDL to contain built-in
+  scalar definitions- [dariuszkuc], [pull/990]**
```
dariuszkuc, line 25:
```suggestion
  scalar definitions - [dariuszkuc], [pull/990]**
```

crates/apollo-compiler/src/schema/from_ast.rs
```diff
@@ -125,10 +133,14 @@ impl SchemaBuilder {
                         Entry::Occupied(entry) => {
                             let previous = entry.get();
                             if $is_scalar && previous.is_built_in() {
-                                self.errors.push(
-                                    $def.location(),
-                                    BuildError::BuiltInScalarTypeRedefinition,
-                                )
+                                if self.allow_builtin_redefinitions {
+                                    continue;
```
goto-bus-stop, line 137: Skipping them means that, when you write:

```graphql
"All strings in our API are ASCII-only."
scalar String @metadata(argument: "value")
```

you can not then access the custom description and directives:

```rust
let s = schema.types.get("String").unwrap();
assert!(s.directives().get("metadata").is_some());
```

which I think is surprising.

How about replacing the type with the custom one? It would make things a bit trickier because we'd also want to re-serialize custom scalars that override builtin ones, I guess. (because unlike graphql-js, apollo-rs cares about directives)

(Another approach to solve the confusion could just be to call the option `ignore_builtin_redefinitions` 😄 .)

goto-bus-stop (changes requested):

crates/apollo-compiler/src/schema/from_ast.rs
dariuszkuc: renamed `allow_builtin_redefinitions` -> `ignore_builtin_redefinitions` in https://github.com/apollographql/apollo-rs/pull/990/commits/25716eb5276f73da032f4f53c927c34589174892

PR #991, opened 2025-08-14
chore: add missing changelog entry for #987 and #989

dariuszkuc:


dariuszkuc:
Keeping it in draft as there will be merge conflict with #990

PR #992, opened 2025-08-15
apollo-compiler@1.29.1

lrlna:


lrlna:
@duckki yea, i cut this before #990 merged, so it _could_ theoretically be a patch if you need that fix in asap. but if this is for composition, rather than query planning (i.e does not need to be in the next router release), we can bundle all the changes together.

duckki:
Either way works for us. Just one `1.30.0` release may be easier :)

PR #994, opened 2025-08-26
feat(compiler): allow SDL to contain built-in type redefinition

dariuszkuc:
Updates `SchemaBuilder` to allow SDL to contain built-in meta types (e.g. `__Type`) redefinitions. GraphQL spec doesn't really define the expected behavior for handling SDL with redeclared meta fields (see: [#1036](https://github.com/graphql/graphql-spec/pull/1036), [#1049](https://github.com/graphql/graphql-spec/pull/1049)). `graphql-js` [reference implementation](https://github.com/graphql/graphql-js/blob/v16.11.0/src/utilities/extendSchema.ts#L187) simply ignores those redefinitions even if the underlying type definitions don't match.

This PR updates `ignore_builtin_redefinitions` option to also ignore meta types redefinitions (previously it was only allowing built-in scalar redefinitions #990).

### DIFF
--- a/crates/apollo-compiler/tests/validation/ignore_builtin_redefinition.rs
+++ b/crates/apollo-compiler/tests/validation/ignore_builtin_redefinition.rs
@@ -17,7 +17,7 @@ fn handles_built_in_scalar_redefinition() {
   "#;
 
     let errors = Schema::parse_and_validate(schema, "schema.graphql")
-        .expect_err("invalid schema")
+        .expect_err("valid schema")
         .errors;
     let expected = expect![[r#"
         Error: built-in scalar definitions must be omitted
@@ -137,7 +137,7 @@ fn handles_built_in_type_redefinition() {
      "#;
 
     let errors = Schema::parse_and_validate(schema, "schema.graphql")
-        .expect_err("invalid schema")
+        .expect_err("valid schema")
         .errors;
     let expected = expect![[r#"
         Error: the type `__Directive` is defined multiple times in the schema


crates/apollo-compiler/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## Features
 
+- **Update `ignore_builtin_redefinitions` option to allow SDL to contain built-in meta type definitions - [dariuszkuc], [pull/994]**
```
tninesling, line 24: Since this hasn't been released yet, should we combine these two entries?

tninesling (approved):

crates/apollo-compiler/tests/validation/ignore_builtin_redefinition.rs
```diff
@@ -0,0 +1,256 @@
...
+  scalar Boolean
+  scalar ID
+
+  type Query {
+    foo: String
+  }
+  "#;
+
+    let errors = Schema::parse_and_validate(schema, "schema.graphql")
+        .expect_err("valid schema")
```
lrlna, line 20: i changed the message here (before it said "invalid schema"). `expect_err` panics when an `Err` is expected but we got an `Ok`. So here, instead of getting an invalid schema, we actually get a valid one.

crates/apollo-compiler/tests/validation/types.rs
```diff
@@ -2018,34 +2016,3 @@ mod variable_default_values {
         );
     }
 }
-
-#[test]
-fn handles_built_in_type_redefinition() {
```
lrlna, line 2023: This `types.rs` file was for ValuesOfCorrectType Rule tests, so I moved these builtin_type_redefinition tests into their own file.

crates/apollo-compiler/tests/validation/ignore_builtin_redefinition.rs
```diff
@@ -0,0 +1,256 @@
+use apollo_compiler::schema::SchemaBuilder;
+use apollo_compiler::Schema;
+use expect_test::expect;
+
+#[test]
+fn handles_built_in_scalar_redefinition() {
+    let schema = r#"
+  scalar String
```
lrlna, line 8: So we are extra sure, I added all the scalar redefinitions and all the type redefinitions (below).

crates/apollo-compiler/tests/validation/ignore_builtin_redefinition.rs
lrlna: I left it as incorrect, and all the other types I added were also slightly incorrect. A bit harder with scalars, so scalars are all the same.

PR #1000, opened 2025-09-15
chore(compiler): Add benchmark for parsing and validation when a type has many extensions

tninesling:
# Overview

Introduces a new benchmark for query parsing and validation when a type has many extensions.

## Comparing `apollo-compiler@1.28.0` to `apollo-compiler@1.30.0`

We made an update in `apollo-compiler@1.28.0` to expose `.iter_origins()` for AST nodes, and we reimplemented `.extensions()` in terms of `.iter_origins()`. We were concerned that this may have caused a performance regression in parsing, but running this new benchmark against `main` with `1.28.0` as the base indicates no change in performance:

```
many_extensions         time:   [9.4118 ms 9.4420 ms 9.4842 ms]
                        change: [−0.3722% +1.0521% +2.1863%] (p = 0.11 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe
```


lrlna (approved):
Thank you!

PR #1002, opened 2025-09-26
Fix: Accept all valid Float value inputs on `coerce_variable_value`

rb090:
This PR fixes the behaviour of `Float` input values in the `apollo-compiler` crate in `apollo_compiler::request::coerce_variable_values` to align with the [GraphQL specification](https://spec.graphql.org/September2025/#sec-Float.Input-Coercion).

The GraphQL specifications defines the input coercion like this:

> When expected as an input type, both integer and float input values are accepted. Integer input values are coerced to Float by adding an empty fractional part, for example `1.0` for the integer input value `1`.

The issue this PR solves is, that `coerce_variable_values` incorrectly rejects JSON numbers without fractional parts (e.g. 42) when the corresponding input variable type was `Float`.

The changes include accepting any JSON number (int or float) that can be converted to an `f64` for `Float` inputs as well as a small unit test `test_graphql_float_variable_coercion` showing the behaviour.

Big thanks to my colleague [David](https://github.com/DavidDiffusion), who identified where the changes needed to be applied and also reviewed the code updates. Much appreciated! 🙌

### DIFF
--- a/crates/apollo-compiler/src/resolvers/input_coercion.rs
+++ b/crates/apollo-compiler/src/resolvers/input_coercion.rs
@@ -119,8 +119,10 @@ fn coerce_variable_value(
                 }
             }
             "Float" => {
-                // https://spec.graphql.org/October2021/#sec-Float.Input-Coercion
-                if value.is_f64() {
+                // https://spec.graphql.org/September2025/#sec-Float.Input-Coercion
+                // Accept any JSON number (`int` or `float`) that coerces to a finite f64,
+                // rejecting special values (NaN, +∞, -∞) as required by the GraphQL spec.
+                if value.as_f64().is_some_and(f64::is_finite) {
                     return Ok(value.clone());
                 }
             }
--- a/crates/apollo-compiler/tests/introspection.rs
+++ b/crates/apollo-compiler/tests/introspection.rs
@@ -321,3 +321,44 @@ fn mixed() {
         }"#]]
     .assert_eq(&response);
 }
+
+#[test]
+fn test_graphql_float_variable_coercion() {
+    // Small schema with a Float in the input object
+    let sdl = r#"
+      type Car { id: ID! kilometers: Float! }
+      input CarInput { kilometers: Float! }
+      type Query { getCarById(id: ID!): Car }
+      type Mutation { insertACar(car: CarInput!): Car!
+      }
+    "#;
+
+    let parsed_schema = Schema::parse_and_validate(sdl, "sdl").unwrap();
+
+    let executable_mutation = ExecutableDocument::parse_and_validate(
+        &parsed_schema,
+        "mutation MyCarInsertMutation($car: CarInput!){ insertACar(car:$car) { id kilometers } }",
+        "MyCarInsertMutation",
+    ).unwrap();
+    let operation = executable_mutation.operations.get(Some("MyCarInsertMutation")).unwrap();
+
+    let kilometers_value = 3000;
+
+    // Provide an integer for a Float field
+    let input_variables = serde_json_bytes::json!({ "car": { "kilometers": kilometers_value } });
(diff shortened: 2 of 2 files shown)


crates/apollo-compiler/tests/input_coercion.rs
```diff
@@ -0,0 +1,50 @@
+
+#[test]
```
lrlna, line 2: This test won't run as is. It needs to be added to `apollo-compiler/tests/validation/main.rs`. Once added, you will notice rust compiler won't compile this file, and you will need to declare `Schema`, `ExecutableDocument` and `coerce_variable_values`.

crates/apollo-compiler/tests/input_coercion.rs
```diff
@@ -0,0 +1,50 @@
...
+fn test_graphql_float_variable_coercion() {
+    // Small schema with a Float in the input object
+    let sdl = r#"
+      type Car { id: ID! kilometers: Float! }
+      input CarInput { kilometers: Float! }
+      type Query { getCarById(id: ID!): Car }
+      type Mutation { insertACar(car: CarInput!): Car!
+      }
```
lrlna, line 10: Nit: this might be a bit easier to read.
```suggestion
      type Mutation { insertACar(car: CarInput!): Car! }
```

crates/apollo-compiler/tests/input_coercion.rs
```diff
@@ -0,0 +1,50 @@
...
+        .unwrap();
+
+    let kilometers_value = 3000;
+
+    // Provide an integer for a Float field
+    let input_variables = serde_json_bytes::json!({ "car": { "kilometers": kilometers_value } });
+    let map = match input_variables {
+        serde_json_bytes::Value::Object(m) => m,
+        _ => unreachable!(),
+    };
```
lrlna, line 34: Could I ask you to add another test to this file that shows how coercion would fail? 

Since we have changed the coercion code to now be `value.as_f64().is_some_and(f64::is_finite)`, there is a chance for a `None` to be returned from `as_f64`. You might have to have `kilometers_value` be defined as a string, otherwise the compiler won't let this compile. So potentially two variations of the test like this:

```rust
// Infinity; should fail coercion
let kilometers_value = "1e1000";
```
and another one:
```rust
// i128 MAX; should also fail coercion, since it's > i64 MAX
let kilometers_value = "170141183460469231731687303715884105727";
```

lrlna (changes requested):
Thank you so much for your contribution! There are few small things to fix, which I've noted in the review comments. Once those are in, I can merge this in.

tninesling (approved):
This looks good to me

crates/apollo-compiler/tests/input_coercion.rs
```diff
@@ -0,0 +1,149 @@
...
+/// If coercion fails (for example, due to exceeding `f64` limits or type mismatches),
+/// an error message is returned as a [`String`].
+fn build_and_coerce_test_mutation_variables<R, K>(
+    range: R,
+    total_kilometers: K,
+) -> Result<Map<ByteString, Value>, String>
+where
+    R: Into<f64>,
+    K: Into<f64>,
+{
```
lrlna: This helper function is too complex. What you had previously with a single test was more readable and covered the changes more appropriately. If you go back to what you had previously, you can just add additional calls to `coerce_variable_values` with different values and additional asserts to validate .

rb090:
Thank you for the review, @lrlna and @tninesling. I really appreciate the feedback you gave me on this.
I’ve reverted the changes to match the 1st commit I did.

@tninesling I agree with you, let’s please move forward with your PR. Thanks a lot for your help and collaboration. I really appreciate it. 

The key thing is that we get this issue resolved, as it’s an important one.

Is it okay if we close this PR?

PR #1003, opened 2025-09-26
fix: catch missing fragments

### LINKED ISSUE #988: Missing fragment validation on stadalone executable validation.
## Description

With the following query:

``` graphql
query {
    company { 
        user {
          ...UserFragment
        }
        
        ...CompanyFragment
  }
}

fragment UserFragment on User {
    id
    name
}
```

And with something like this to parse and validate:

``` rust
let document = apollo_compiler::ast::Document.parse(query, path).unwrap();

println!("{:?}", document.validate_standalone_executable());
```

It returns `Ok(())`, but I expected to see a fragment missing error.

I don't know if it's an expected behaviour, but missing fragments is something that can be validated without a schema.

For reference, in "graphql-js"  they do have this validation without the schema:

``` javascript
import graphql from 'graphql.js'

var graph = graphql("/graphql")

graph.fragment({
  UserFragment: `on User {
    id,
    name
  }`
})

const allUsers = graph(`query {
    company {
        user {
          ...UserFragment
        }

        ...CompanyFragment
  }
}`)

await allUsers()


// Error: Fragment CompanyFragment not found
```

Abdel-Monaam-Aouini:
Summary

Ensures nested selections are traversed without a schema, so undefined fragments are reported. 

Closes #988

### DIFF
--- a/crates/apollo-compiler/src/validation/field.rs
+++ b/crates/apollo-compiler/src/validation/field.rs
@@ -33,10 +33,18 @@ pub(crate) fn validate_field(
 
     super::argument::validate_arguments(diagnostics, &field.arguments);
 
-    // Return early if we don't know the type--this can happen if we are nested deeply
-    // inside a selection set that has a wrong field, or if we are validating a standalone
-    // operation without a schema.
+    // If we don't know the type (no schema, or invalid parent), we cannot perform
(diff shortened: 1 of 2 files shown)


lrlna (approved):
Thank you so much @Abdel-Monaam-Aouini!

PR #1011, opened 2025-10-28
feat(compiler): Allow coercing Int variables to Float

tninesling:
# Coercing Int variable values to Float

The GraphQL spec allows coercing Int values to Float in input positions (see [input coercion]). There are a couple things to note about this. 

1. Strings are not allowed to be coerced in this position, even if they are numeric.
2. Ints can only be converted to Float when they are "representable by finite IEEE 754" floating point numbers.

Since an IEEE 754 floating point double (f64) has 53 bits of precision, it can safely represent up to the value `2^53 - 1` as a finite value. Beyond that, the round trip from integer to float and back will lose information. I have represented this with a new `MAX_SAFE_INT` constant which is often included in other languages like JavaScript's `Number.MAX_SAFE_INT`. When, we encounter an Int variable in a Float position, we ensure that its value is finitely representable.

There is some nuance in that the spec does not say _all_ floats have to be within this range. So, this implementation allows explicitly passed floats which are greater than that bound, only applying the integer conversion limit when coercing a value.

[input coercion]: https://spec.graphql.org/September2025/#sec-Float.Input-Coercion

### DIFF
--- a/crates/apollo-compiler/CHANGELOG.md
+++ b/crates/apollo-compiler/CHANGELOG.md
@@ -70,6 +70,29 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
         .expect("schema parsed successfully");
   ```
 
+- **Allow coercing Int variables to Float - [tninesling], [pull/1011]**
+
+  The GraphQL spec allows coercing Int values to Float in input positions (see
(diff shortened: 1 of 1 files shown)


lrlna (approved):
thank you!

PR #1012, opened 2025-10-29
Add `iter_orphan_extension_types` method on `SchemaBuilder`

duckki:
This PR adds a public method on `SchemaBuilder` that exports the names of orphan extension types.

This is an alternative solution to https://github.com/apollographql/apollo-rs/pull/1009 without breaking the API.


tninesling (approved):
Awesome workaround while we figure out a long-term solution 🚀

PR #1030, opened 2026-03-27
feat(compiler): implement @oneOf input objects

### LINKED ISSUE #882: Implement `@oneOf` input unions
What will we need, beyond validation?

https://github.com/graphql/graphql-spec/pull/825

abernix:
## Summary

Implements the [``@oneOf`` input objects directive][rfc], which became part of the stable GraphQL spec in September 2025.  ``@oneOf`` lets a schema author say "exactly one of these fields must be provided" — it's a discriminated-union pattern that the type system enforces rather than leaving to resolver logic.

[rfc]: https://github.com/graphql/graphql-spec/pull/825

Closes #882.

## How it works — three validation layers

The interesting design question is *when* each constraint should fire.  We ended up with three complementary layers:

1. *Schema validation* (`validation/input_object.rs`) — when a developer writes a ``@oneOf`` type, every field must already be nullable and carry no default.  These aren't runtime concerns; they're structural invariants the schema has to uphold before any query is written.

2. *Static document validation* (`validation/value.rs`) — when a query is compiled, a variable used to satisfy a ``@oneOf`` field must itself be declared non-null.  A nullable variable is unsafe because it could legally be omitted at runtime, which would make the field-count invariant impossible to enforce without executing the query.  This is the rule that most other implementations miss or leave to runtime.  We validated against Apollo Kotlin, Hot Chocolate, Strawberry, graphql-java, and juniper — our static check is stricter than juniper and graphql-java, but strictly correct per spec.

3. *Runtime coercion* (`request/coerce_variable_values`) — when variables arrive at execution time, exactly one field must be provided and it must be non-null.  This is the canonical spec rule and the last line of defense for literal values or dynamic inputs the static analysis can't see.

## What's new in the public API

- `directive @oneOf on INPUT_OBJECT` is now a built-in.
- `InputObjectType::is_one_of() -> bool` — clean check without digging into directives.
- `__Type.isOneOf: Boolean!` in the introspection schema — returns `true` for ``@oneOf`` types, `null` for everything else (matching the spec's introspection definition).
- Five new diagnostic variants covering both schema- and document-level violations.

## Fuzz coverage

Two complementary fuzz targets:

- `one_of` — structure-aware, using `apollo-smith`.  We extended smith so it generates ``@oneOf`` input types naturally (with all invariants pre-satisfied at generation time), which means the fuzzer spends its budget on schema shape diversity rather than discarding malformed SDL.  Asserts that `is_one_of()` always agrees with directive presence, and all field invariants hold.

- `one_of_exec` — the schema-invariant target can't reach document-level rules, because valid schema SDL and valid executable SDL are different constraint domains.  This second target fixes a rich ``@oneOf`` schema and fuzzes only the executable document.  It confirmed coverage of all five ``@oneOf``-specific code paths in both `validation/value.rs` and `coerce_variable_values`.

## Version

`apollo-compiler` bumps to `1.32.0` — minor because `is_one_of()` and the introspection field are new public API.  `apollo-smith` bumps to `0.15.3` — 0.x minor-y/patch-y because `Ty::as_nullable()` is new public API.

## Notes

The production federation path (Rover + JS composition) preserves ``@oneOf`` through the supergraph correctly.  Rust in-process composition currently strips all directives from input types during merge — that's a separate pre-existing gap and is tracked as a follow-up, not a blocker for this PR.

[ROUTER-454]: https://apollographql.atlassian.net/browse/ROUTER-454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


martinbonnin (approved):
One comment from https://github.com/graphql/graphql-spec/pull/1211 but looks good otherwise!

Nice test suite!

crates/apollo-compiler/src/validation/input_object.rs
```diff
@@ -100,6 +101,38 @@ pub(crate) fn validate_input_object_definition(
...
+                        coordinate: TypeAttributeCoordinate {
+                            ty: input_object.name.clone(),
+                            attribute: field_name.clone(),
+                        },
+                        default_location,
+                    },
+                );
+            }
+        }
+    }
```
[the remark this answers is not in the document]
abernix: You're probably following this better than I am — do you think that's something we should do now?  Or ... later?  Is there any risk of it not landing?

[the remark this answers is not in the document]
martinbonnin: graphql-js is still missing it. You're safe to add it down the road. 

I'd merge this as is and we can revisit later.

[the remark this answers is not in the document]
martinbonnin: It's on the [agenda for the May wg](https://github.com/graphql/graphql-wg/blob/main/agendas/2026/05-May/07-wg-primary.md).

fuzz/fuzz_targets/one_of.rs
```diff
@@ -0,0 +1,65 @@
...
+        };
+
+        let has_directive = input_obj.directives.get("oneOf").is_some();
+        let is_one_of = input_obj.is_one_of();
+
+        // Invariant: is_one_of() must agree with directive presence.
+        assert_eq!(
+            has_directive, is_one_of,
+            "is_one_of() disagrees with @oneOf directive presence for type `{type_name}`"
+        );
```
lrlna: This is a bit of a strange thing to check for: `input_obj.is_one_of()` on line 39 literlally does the same thing as the line above it. So you are asserting that `input_obj.directives.get("oneOf").is_some()` is equal to `input_obj.directives.get("oneOf").is_some()`. This would work if you are going getting the directive in a different way, by walking the entire tree for example, but not for as is.

fuzz/fuzz_targets/one_of_exec.rs
```diff
@@ -0,0 +1,108 @@
...
+fuzz_target!(|data: &str| {
+    let _ = env_logger::try_init();
+    debug!("{data}");
+
+    let schema = schema();
+
+    // Parse and validate the fuzz input as an executable document.
+    // We intentionally discard validation errors — we are looking for panics
+    // and for coverage of the @oneOf validation code paths.
+    let result = ExecutableDocument::parse_and_validate(schema, data, "fuzz.graphql");
```
lrlna, line 90: We should be generating an executable document in fuzz/src/lib.rs in a new function, something like `generate_valid_executable_document`. That function should use `Unstructured` to create the executable document, and you can indeed pass it the schema you have defined above. The router repo sort of does something similar in its fuzz src, albeit using the parser rather than the compiler.

crates/apollo-smith/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "apollo-smith"
-version = "0.15.2" # When bumping, also update README.md
+version = "0.15.3" # When bumping, also update README.md
```
lrlna: This is usually bumped when a release go through and in a separate PR. So let's leave it on 0.15.2 until a release happens.

crates/apollo-compiler/src/validation/diagnostics.rs
```diff
@@ -292,8 +292,39 @@ pub(crate) enum DiagnosticData {
...
+    #[error(
+        "variable `${variable}` is of type `{variable_type}` \
+         but must be non-nullable to be used for @oneOf input object `{name}` field `{field}`"
+    )]
+    OneOfInputObjectNullableVariable {
+        name: Name,
+        field: Name,
+        variable: Name,
+        variable_type: Node<ast::Type>,
+    },
```
lrlna: Can this be an existing `DisallowedVariableUsage` diagnostic?

lrlna: If `must be non-nullable to be used for `@oneOf` input object` is a useful addition, it can be added as a help text.

crates/apollo-compiler/src/validation/diagnostics.rs
```diff
@@ -292,8 +292,39 @@ pub(crate) enum DiagnosticData {
         "{describe} cannot be named `{name}` as names starting with two underscores are reserved"
     )]
     ReservedName { name: Name, describe: &'static str },
+    #[error("`{coordinate}` field of a @oneOf input object must be nullable")]
+    OneOfInputObjectFieldNonNull {
+        coordinate: TypeAttributeCoordinate,
+        definition_location: Option<SourceSpan>,
+    },
```
lrlna, line 367: This sounds like an existing `UnsupportedValueType` rule, which can have a help text specifying this field needs to be nullable.

crates/apollo-compiler/src/validation/diagnostics.rs
```diff
@@ -292,8 +292,39 @@ pub(crate) enum DiagnosticData {
...
+    #[error(
+        "@oneOf input object `{name}` must specify exactly one key, but {provided} {} given",
+        if *provided == 1 { "was" } else { "were" }
+    )]
+    OneOfInputObjectWrongNumberOfFields { name: Name, provided: usize },
+    #[error("`{name}.{field}` value for @oneOf input object must be non-null")]
+    OneOfInputObjectNullField { name: Name, field: Name },
```
lrlna: This also sounds like `UnsupportedValueType` rule.

crates/apollo-compiler/src/validation/value.rs
```diff
@@ -219,6 +219,48 @@ pub(crate) fn value_of_correct_type(
...
+                        );
+                    } else if let Some((field_name, field_val)) = obj.first() {
+                        if field_val.is_null() {
+                            diagnostics.push(
+                                field_val.location(),
+                                DiagnosticData::OneOfInputObjectNullField {
+                                    name: input_obj.name.clone(),
+                                    field: field_name.clone(),
+                                },
+                            );
```
lrlna: Was this diagnostic not being caught already by an `UnsupportedValueType`? Because if something is declared NonNull in the schema, it should be caught. It doesn't matter if it's a oneOf input object or not.

crates/apollo-compiler/src/validation/value.rs
```diff
@@ -219,6 +219,48 @@ pub(crate) fn value_of_correct_type(
...
+                                            field: field_name.clone(),
+                                            variable: var_name.clone(),
+                                            variable_type: var_def.ty.clone(),
+                                        },
+                                    );
+                                }
+                            }
+                        }
+                    }
+                }
```
lrlna: Does this not get validated in `is_variable_usage_allowed`?

crates/apollo-compiler/src/validation/diagnostics.rs
```diff
@@ -292,8 +292,39 @@ pub(crate) enum DiagnosticData {
...
+    #[error("`{coordinate}` field of a @oneOf input object must be nullable")]
+    OneOfInputObjectFieldNonNull {
+        coordinate: TypeAttributeCoordinate,
+        definition_location: Option<SourceSpan>,
+    },
+    #[error(
+        "@oneOf input object `{name}` must specify exactly one key, but {provided} {} given",
+        if *provided == 1 { "was" } else { "were" }
+    )]
+    OneOfInputObjectWrongNumberOfFields { name: Name, provided: usize },
```
lrlna: I'd rather have these be non-specific to oneOf. Perhaps this can be `UniqueField` or `UniqueKey`, similar to `UniqueDirective`?

crates/apollo-compiler/src/validation/diagnostics.rs
```diff
@@ -292,8 +292,39 @@ pub(crate) enum DiagnosticData {
...
+    #[error("`{name}.{field}` value for @oneOf input object must be non-null")]
+    OneOfInputObjectNullField { name: Name, field: Name },
+    #[error("`{coordinate}` field of a @oneOf input object must not have a default value")]
+    OneOfInputObjectFieldHasDefault {
+        coordinate: TypeAttributeCoordinate,
+        default_location: Option<SourceSpan>,
+    },
```
lrlna: Like one of the comments above, this should also be named independantly of oneOf. Perhaps, just `UnsupportedDefault`?

crates/apollo-compiler/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "apollo-compiler"
-version = "1.31.1" # When bumping, also update README.md
```
lrlna: same comment as with apollo-smith: we usually bump these in a release PR

lrlna (dismissed):

lrlna:
The main things to sort out are the diagnostics and whether some of them already run, but otherwise should be good to land.

crates/apollo-compiler/src/validation/diagnostics.rs
abernix: I'm not sure the `Unique*` parallel quite holds.

All the existing `Unique*` variants (`UniqueVariable`, `UniqueArgument`, `UniqueDirective`, `UniqueInputValue`) mean "this thing was provided more than once".  i.e. *at most one.*  The `@oneOf` rule is *exactly one*, and this diagnostic also fires for *zero* fields, which doesn't fit the uniqueness framing.

Two alternatives worth considering:

1. Split into `OneOfInputObjectMissingField` (zero case) and `OneOfInputObjectMultipleFields` (2+ case).  Seems cleanly defined, but the con is it doubles the variant count for what's a single spec rule.
2. Keep the single variant pragmatic given "exactly one of N" is, I think, currently `@oneOf`-"only" in GraphQL.

what's your take?  same as before?

crates/apollo-compiler/src/validation/diagnostics.rs
abernix: There's a clean middle ground here, I think.

Your underlying point that this shouldn't be `@oneOf`-specific is fair.  However folding directly into `UnsupportedValueType` seems awkward because its message reads `"expected value of type {ty}, found {value}"`, plus the declared `ty` is nullable, so "expected `String`, found null" reads as kinda misleading? (`String` does allow null at _most_ positions; `@oneOf` is the special case).

Would you be in favor of a new generic variant like `NullValueNotAllowed` (or `NullValueAtRequiredPosition`?) that describes the actual invariant ("this position is treated as required even though the declared type allows null").  It'd be reusable for any future feature with the same shape; and carries the `@oneOf` specifics in help text.

Want me to take a swing at that?

crates/apollo-compiler/src/validation/diagnostics.rs
abernix: Fair. will swap to `UnsupportedDefault` and move the `@oneOf` specifics to help text.

crates/apollo-compiler/src/validation/diagnostics.rs
abernix: Tried this but the field shape doesn't quite line up.  Notes below!

`DisallowedVariableUsage` carries `argument: Name` and `argument_type: Type` but in this case here the variable's location is an *input object field*, not an argument.  Folding into the other variant would mean either repurposing those fields with mismatched names, or renaming them to something neutral.  The variant's string identifier is externally observable via `mod.rs` publicity so the ripple touches consumers matching on diagnostic kind too, right?

Three(?+) paths:

a. Keep `OneOfInputObjectNullableVariable` as a distinct variant which is most accurate semantically; consumers could match on it cleanly.
b. Rename `DisallowedVariableUsage`'s fields to `location_name` / `location_type` and fold both cases.  It can be an internal-only field rename, but the variant's meaning widens (which arguably has ripple risk for any consumer matching on it).
c. New generic `InvalidVariableUsage` variant covering both 👯 same field shape as today, just less `@oneOf`-specific in the naming of.

Thoughts?  Mildly lean toward "a" since the location semantics seem genuinely different but happy to follow b/c if you'd rather see consolidation.  Which do you prefer?

crates/apollo-compiler/src/validation/value.rs
abernix: So as a double check: I stashed the rule, re-ran `invalid_one_of_nullable_variable`, and document came back `Valid` with no other diagnostic.  So `is_variable_usage_allowed` **doesn't** catch it.

I think for two reasons:

1. `validate_variable_usage` is only invoked at the top-level argument boundary (in `field.rs:58` and in `directive.rs:303`), and it bails immediately unless the *entireeeeee* argument value is a `Value::Variable`.  When the argument **is** `arg: { stringField: $v }`, it sees an `Object` literal and the check is skipped entirely.

2. Even if it did recurse in,`is_variable_usage_allowed`'s strict branch requires `location_ty.is_non_null()`.  However, `@oneOf` fields are declared nullable, so that branch is skipped and the fall-through `is_assignable_to(location_ty)` returns true.

Net net that means: `query Q($v: String) { f(arg: { stringField: $v }) }` passes static validation entirely without the addition of this rule, leaving the violation to surface only at runtime coercion time.

Side note that this digging around is flagging: I think the recursive variable handling in `value.rs:135-158`:

https://github.com/apollographql/apollo-rs/blob/809ef8e9f5fb824bdb850401a1699c15a296b394/crates/apollo-compiler/src/validation/value.rs#L135-L158

...only checks type *names*, not nullability — so spec [rule 5.8.5](https://spec.graphql.org/October2021/#sec-All-Variable-Usages-are-Allowed) **isn't** fully applied to nested input-object positions even *outside* `@oneOf`.

That can be verified locally with a `RegularInput { y: String! }` test, which passes as `Valid`.  If I'm correct, that's a separate latent gap unrelated to this PR.  If you agree, I can file a follow-up issue.

crates/apollo-compiler/Cargo.toml
abernix: Yeah, I can revert that.  I had this for local testing of something.

crates/apollo-smith/Cargo.toml
abernix: sounds good, i'll circle back with a revert on that.

fuzz/fuzz_targets/one_of.rs
abernix: Comparing a value to itself is certainly genius.  🙂   Let remove that.

I think the remaining invariant checks (all fields nullable, no defaults) are the ones actually exercising the schema-validation rule, and they'll stay.

fuzz/fuzz_targets/one_of_exec.rs
abernix: I think I agree in principle.  The mechanical scope is chunkier than I'd hoped though since `apollo-smith` generates schema + operations together on its own internal `****TypeDef` types, and I don't think there's  an option today that takes an existing `apollo_compiler::Schema` and uses it to generate only operations against it.  I think `with_document` is the closest hook but it expects an `apollo_smith::Document`, so we'd need a conversion path?

Two routes:

1. *In this PR*, I can add a `from_compiler_schema(&Schema) -> apollo_smith::Document` conversion in apollo-smith that maps every type kind, then `generate_valid_executable_document` in `fuzz/src/lib.rs` can wrap `with_document` + `operation_definition`.
2. OR, and I think this is my preference: *Follow-up PR*, and instead land this with the existing raw-string target (which genuinely caught coverage of the new code paths during dev: all five `@oneOf` paths in `value.rs` and `coerce_variable_values` get hit), file a follow-up for the structure-aware version.

Happy to do "1" here if you'd prefer.  Which do you prefer?

crates/apollo-compiler/src/validation/diagnostics.rs
abernix: I don't think `UnsupportedValueType` quite fits here since its fields are `{value, ty, definition_location}` and scoped to value coercion.  This variant here fires at *schema* validation, where there's no value only a type definition.  It appears that schema-validation diagnostics in this codebase keep their own variants (e.g., `RecursiveInputObjectDefinition`, `EmptyInputValueSet`, etc.) for the same reason?

If the underlying ask is "make the name less `@oneOf`-specific," happy to rename to `InputObjectFieldNonNull` and let the help text (which is already using the shared `ONE_OF_FIELD_REQUIREMENTS` constant) carry the `@oneOf` context.

Does that get at the root or is it something else??

crates/apollo-compiler/src/validation/value.rs
abernix: I'm relatively sure it is not getting caught: to verify, I stashed the new rule this PR introduces, re-ran `invalid_one_of_null_field`, and the document came back `Valid` ✅  with no other diagnostic.  So the gap seems real.

A reason?: `@oneOf` fields are *declared* nullable and that's currently enforced at schema validation time in `input_object.rs:108-119` here:

https://github.com/apollographql/apollo-rs/blame/00854481b9a3459ace23395f2e185e5d909de555/crates/apollo-compiler/src/validation/input_object.rs#L108-L119

(If that's wrong, we should discuss!)

So at this point in `value.rs`, the field's `ty.is_non_null()` is always false, which means `UnsupportedValueType`'s null branch never fires for the literal-`null` case.  Without this rule, `{ stringField: null }` silently passes static validation and only blows up at runtime coercion.

Where I think you're directionally right though: the *diagnostic* doesn't have to be `@oneOf`-specific.  I replied on `OneOfInputObjectNullField` with a generic-variant proposal that could be right / what you're looking for.

abernix (commented):
Thank you for your review!  I've replied to I think all of them here.

crates/apollo-smith/Cargo.toml
abernix: done: f9a23415

crates/apollo-compiler/Cargo.toml
abernix: done: f9a23415

crates/apollo-compiler/src/validation/diagnostics.rs
abernix: done: 64457a1b

fuzz/fuzz_targets/one_of.rs
abernix: done: 823426bd

fuzz/fuzz_targets/one_of_exec.rs
lrlna: `DocumentBuilder::with_document` is indeed what you are looking for. We do this - generate operations from an existing schema - in the router. You can take a look at an example [here](https://github.com/apollographql/router/blob/dev/fuzz/src/lib.rs).

crates/apollo-compiler/src/validation/value.rs
lrlna: ok right, but then this should call `unsupported_type` and therefore return `UnsupportedValueType` error. An InputValueDefinition still has a `Name` and a `Type`, and if its `Type` is incorrect, it's unsupported.

crates/apollo-compiler/src/validation/value.rs
lrlna: Right, so this means `is_variable_usage_allowed` needs to be patched up. One of the arguments is the `variable_usage`, aka `InputValueDefinition`, which will have `directives` such as `oneOf` defined on it. So this check can be done there.

The part of the code here should be recursively calling to `value_of_correct_ty`, which would land it in the `ast::Value::Variable`, which can then call validate_variable_usage.

crates/apollo-compiler/src/validation/value.rs
```diff
@@ -219,6 +219,48 @@ pub(crate) fn value_of_correct_type(
...
+                if input_obj.is_one_of() {
+                    if obj.len() != 1 {
+                        diagnostics.push(
+                            arg_value.location(),
+                            DiagnosticData::OneOfInputObjectWrongNumberOfFields {
+                                name: input_obj.name.clone(),
+                                provided: obj.len(),
+                            },
+                        );
+                    } else if let Some((field_name, field_val)) = obj.first() {
```
lrlna: Why is this only checking for `obj.first`? Shouldn't this check for all of them? If so, we might have to potentially recurse into `value_of_correct_type` again, so this would fall into the `ast::Value::Null` branch (and trigger `unsupported_type`).

lrlna (commented):
Answering a few of your questions.

crates/apollo-compiler/src/validation/value.rs
abernix: I tried to address this in b3030054 — the null-field case now routes through `unsupported_type`, but it uses a synthetic non-null type derived from the field's declared nullable type.  This means the diagnostic reads differently now (snapshot `invalid_one_of_null_field` shows it).

crates/apollo-compiler/src/validation/diagnostics.rs
abernix: I think b3030054 covers this — the `@oneOf` nullable-variable case now emits `DisallowedVariableUsage`, with the field's declared type coerced to non-null.  This means the message reads as *"variable `$x` of type `String` cannot be used for argument `stringField` of type `String!`"*.

That felt like enough to me — the coerced type seems to carry the "must be non-nullable" invariant on its own — but I could be missing something.  Let me know if you'd still like a `@oneOf`-specific help text on top of that.

crates/apollo-compiler/src/validation/value.rs
abernix: With 86609a19, `value_of_correct_type` now threads a `VariablePosition` through recursion, and the `Value::Variable` arm calls `is_variable_usage_allowed` with the position's type + default, emitting `DisallowedVariableUsage` when the check fails.

Also did the refactor of `is_variable_usage_allowed` so it takes `(variable_def, location_ty, location_has_default)` making it possible for list-value entries to share the same check.  And also made the pass at closing a pre-existing `TODO` which was in the neighborhood.

crates/apollo-compiler/src/validation/diagnostics.rs
abernix: Ok, with b3030054 this routes through `UnsupportedValueType` directly but it uses a synthetic non-null type derived from the field's declared type.  This means the diagnostic reads as *"expected value of type `String!`, found null"*.  (The bespoke `NullValueNotAllowed` variant I'd sketched above is retired.)

crates/apollo-compiler/src/validation/diagnostics.rs
abernix: Landed in baa081a3 as `OneOfInputObjectFieldCount` — kept the `OneOf` prefix for now.

Can you help me understand the `Unique*` fit?  What I see is that the existing `Unique*` variants (`UniqueVariable`, `UniqueArgument`, `UniqueDirective`, `UniqueInputValue`) carry `{ name, original_definition, redefined_definition }` for at-most-one / redefined-name semantics.  The `@oneOf` rule is *exactly-one* and (in my head, at least) needed a payload that carries `Vec<(Name, Option<SourceSpan>)>` for the per-field labels because it also fires for *zero* fields.

Am I missing a way `UniqueField` maps cleanly onto that shape?

crates/apollo-compiler/src/validation/diagnostics.rs
abernix: Kept `OneOfInputObjectFieldNonNull` for now.  Can you help me understand the `UnsupportedValueType` bits? What I see is that `UnsupportedValueType` expects `{value, ty, definition_location}` but the check here runs at schema _definition_ time on `input Foo @oneOf { a: String! }`, where no value is present.

The other schema-only checks (`RecursiveInputObjectDefinition`, `EmptyInputValueSet`) keep dedicated variants for what I"d guess was the same reason. 

Is there a shape here where `UnsupportedValueType` fits that I'm not seeing?

crates/apollo-compiler/src/validation/value.rs
abernix: I think this is done — spread across b3030054 (iterate all provided fields) and 86609a19 (collapse into the recursive path).  Your recursion nudge made the collapse cleaner than I'd expected.

The per-defined-field iteration in `value_of_correct_type` now coerces the field's declared nullable type to non-null when the enclosing input object is `@oneOf`.  So the generic `Value::Null` arm fires `unsupported_type` and the generic `Value::Variable` arm fires `DisallowedVariableUsage`.  (Bespoke per-provided-field iteration retired.)

fuzz/fuzz_targets/one_of_exec.rs
abernix: Thanks for the example.  Will leave this unresolved and land a `DocumentBuilder::with_document` fuzzer as a follow-up.

crates/apollo-compiler/CHANGELOG.md
```diff
@@ -42,6 +29,73 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+
+[dariuszkuc]: https://github.com/dariuszkuc
+[lrlna]: https://github.com/lrlna
+[trevor-scheer]: https://github.com/trevor-scheer
+[pull/1017]: https://github.com/apollographql/apollo-rs/pull/1017
+[pull/1035]: https://github.com/apollographql/apollo-rs/pull/1035
+[pull/1036]: https://github.com/apollographql/apollo-rs/pull/1036
+
+
+>>>>>>> origin/main
```
abernix: Oops — self review!  Fixed this one in d8936f8e.  :)

tninesling:
I think we may have discussed a while back, but is this going to be considered a breaking change? If so, it should probably be retargeted to [the v2 branch](https://github.com/apollographql/apollo-rs/tree/v2). Let me know if this needs more attention from the Rust Platform team, and I can review it.

crates/apollo-compiler/src/built_in_types.graphql
```diff
@@ -32,6 +32,8 @@ type __Type {
   ofType: __Type
   # may be non-null for custom SCALAR, otherwise null.
   specifiedByURL: String
+  # always non-null; true only for @oneOf INPUT_OBJECT types.
+  isOneOf: Boolean!
```
tninesling: The spec says this should be nullable, per https://spec.graphql.org/September2025/#sel-EAXHACuIBCzIBtyP

crates/apollo-compiler/tests/validation/one_of.rs
```diff
@@ -0,0 +1,568 @@
...
+    assert_eq!(
+        data["oneOfType"]["isOneOf"],
+        serde_json::Value::Bool(true),
+        "expected isOneOf=true for @oneOf type, got: {json}"
+    );
+
+    // String is not a @oneOf type → isOneOf must be false.
+    assert_eq!(
+        data["regularType"]["isOneOf"],
+        serde_json::Value::Bool(false),
```
tninesling: It looks like the latest version of the spec has this returning `null` for types which are not InputObjects https://spec.graphql.org/September2025/#sec-The-__Type-Type

There are some corresponding updates to the resolvers which should naturally fall out of updating this assertion.

crates/apollo-compiler/tests/validation/one_of.rs
```diff
@@ -0,0 +1,568 @@
...
+
+#[test]
+fn adding_oneof_via_extension_with_valid_base_type_is_valid() {
+    // A regular input type whose fields are all nullable and have no defaults
+    // can have @oneOf added via extension.
+    Schema::parse_and_validate(
+        r#"
+        type Query { f: String }
+        input Foo { a: String b: Int }
+        extend input Foo @oneOf
```
tninesling, line 715: This will need to be updated to return a validation error. The specification now says

> The oneOf directive must not be provided by an Input Object type extension.

Source: [§3.10.2 Input Object Extensions](https://spec.graphql.org/September2025/#sel-GAHhBbHFDJAACLA_5d)

crates/apollo-compiler/src/validation/input_object.rs
```diff
@@ -100,6 +101,38 @@ pub(crate) fn validate_input_object_definition(
         }
     }
 
+    // @oneOf input objects: all fields must be nullable and must not have default values.
+    // https://spec.graphql.org/draft/#sec-OneOf-Input-Objects
```
tninesling: Now that the specification is stabilized, these comments should be updated to reference https://spec.graphql.org/September2025/#sec-OneOf-Input-Objects instead of the draft

crates/apollo-compiler/src/resolvers/input_coercion.rs
```diff
@@ -407,6 +437,41 @@ fn coerce_argument_value(
...
+                                if provided_count == 1 { "was" } else { "were" }
+                            ),
+                            path,
+                            value.location(),
+                            &ctx.document.sources,
+                        ));
+                        return Err(PropagateNull);
+                    }
+                    if let Some((field_name, field_value)) = object.iter().next() {
+                        if field_value.is_null() {
```
tninesling, line 476: I think this may miss the case where the field is a null variable value. The specification says we should do this check on the coerced values, which may warrant shifting this below the coercion block below

Source: https://spec.graphql.org/September2025/#sec-OneOf-Input-Objects.Input-Coercion

tninesling (changes requested):
This is looking good! I think the main thing we need to do is to bring this in line with some changes which seem to have been made between the draft and official spec release. Once we're compliant with the rules in https://spec.graphql.org/September2025/, I think we're good to go.

tninesling (approved):

PR #1031, opened 2026-03-31
feat(compiler): add CustomScalarCoercer trait for custom scalar validation and coercion

Abdel-Monaam-Aouini:
Description:
Adds a CustomScalarCoercer trait that allows users to register custom
validation and coercion logic for custom scalar types. Previously, custom
scalars accepted any value without validation.

The trait provides two extension points:

coerce_variable_value: validates/transforms JSON variable values during input coercion
validate_ast_value: validates AST literal values (e.g., default values) during schema/document validation
Coercers are registered on the Schema via add_custom_scalar_coercer().
When no coercer is registered, the existing permissive behavior is preserved.

### DIFF
--- a/crates/apollo-compiler/src/schema/mod.rs
+++ b/crates/apollo-compiler/src/schema/mod.rs
@@ -73,11 +73,11 @@ use crate::response::JsonValue;
 use crate::ty;
 use crate::validation::DiagnosticList;
 use crate::validation::Valid;
-use std::sync::Arc;
 use crate::validation::WithErrors;
 pub use crate::Name;
 use crate::Node;
 use std::path::Path;
+use std::sync::Arc;
 use std::sync::OnceLock;
 
 mod component;
--- a/crates/apollo-compiler/src/schema/mod.rs
+++ b/crates/apollo-compiler/src/schema/mod.rs
@@ -69,9 +69,11 @@ use crate::name;
 use crate::parser::FileId;
 use crate::parser::Parser;
 use crate::parser::SourceSpan;
+use crate::response::JsonValue;
 use crate::ty;
 use crate::validation::DiagnosticList;
 use crate::validation::Valid;
+use std::sync::Arc;
 use crate::validation::WithErrors;
 pub use crate::Name;
 use crate::Node;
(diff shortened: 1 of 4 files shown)


crates/apollo-compiler/src/schema/mod.rs
```diff
@@ -125,6 +179,11 @@ pub struct Schema {
     ///   built-in scalar definitions for scalars that are used in the schema.
     ///   We reflect in this Rust API the behavior of `__Schema.types` in GraphQL introspection.
     pub types: IndexMap<NamedType, ExtendedType>,
+
+    /// Registered custom scalar coercers, keyed by scalar type name.
+    ///
+    /// Use [`Schema::add_custom_scalar_coercer`] to register a coercer.
+    custom_scalar_coercers: HashMap<Name, Arc<dyn CustomScalarCoercer>>,
```
duckki: IMO, why don't we allow to set a blanket coercer that handles all customer scalars, instead of adding a map here. So, that the user can decide how to organize this -- either using a map or using a match statement, etc.

Abdel-Monaam-Aouini: @duckki check now :)

crates/apollo-compiler/src/schema/mod.rs
```diff
@@ -98,6 +100,78 @@ pub use crate::ast::NamedType;
...
+    ///
+    /// Called during variable coercion when a variable of a custom scalar type is provided.
+    /// `scalar_name` is the name of the scalar type being coerced.
+    /// Return `Ok(value)` with the (possibly transformed) value if valid,
+    /// or `Err(message)` with an error message if the value cannot be coerced.
+    ///
+    /// The default implementation accepts any value as-is.
+    fn coerce_variable_value(
+        &self,
+        scalar_name: &str,
```
TylerBloom, line 154: I think this ought to be a `&ScalarType` rather than a `&str`. To go along with that, `scalar_name` should just be `scalar`.

crates/apollo-compiler/src/schema/mod.rs
```diff
@@ -98,6 +100,78 @@ pub use crate::ast::NamedType;
...
+        value: &JsonValue,
+    ) -> Result<JsonValue, String> {
+        let _ = scalar_name;
+        Ok(value.clone())
+    }
```
TylerBloom, line 159: This should not be a default implementation.

crates/apollo-compiler/src/schema/mod.rs
```diff
@@ -98,6 +100,78 @@ pub use crate::ast::NamedType;
...
+    /// Called during schema and document validation when a literal value
+    /// (e.g., a default value) is provided for a field of a custom scalar type.
+    /// `scalar_name` is the name of the scalar type being validated.
+    /// Return `true` if the value is acceptable, `false` otherwise.
+    ///
+    /// The default implementation accepts any value.
+    fn validate_ast_value(&self, scalar_name: &str, value: &ast::Value) -> bool {
+        let _ = (scalar_name, value);
+        true
+    }
```
TylerBloom, line 172: Similar thoughts here: No default impl and use `&ScalarType` instead of `&str`.

crates/apollo-compiler/src/schema/mod.rs
```diff
@@ -98,6 +100,78 @@ pub use crate::ast::NamedType;
...
+///
+///     fn validate_ast_value(&self, scalar_name: &str, value: &Value) -> bool {
+///         match scalar_name {
+///             "DateTime" => matches!(value, Value::String(_)),
+///             _ => true,
+///         }
+///     }
+/// }
+/// ```
+pub trait CustomScalarCoercer: Send + Sync {
```
TylerBloom, line 143: Since the `set_custom_scalar_coercer` method takes a `'static` coercer, its probably best to add that bound on the trait itself.

crates/apollo-compiler/src/schema/mod.rs
```diff
@@ -447,6 +526,58 @@ impl Schema {
...
+    ///
+    /// The coercer is called for every custom scalar during validation and input coercion.
+    /// The scalar type name is passed to each method so the implementation can dispatch
+    /// per-scalar logic using a `match` statement, a `HashMap`, or any other approach.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use apollo_compiler::Schema;
+    /// use apollo_compiler::schema::CustomScalarCoercer;
```
TylerBloom, line 539: This example seems a bit verbose. Better to defer to the trait's docs for the example impl.

crates/apollo-compiler/src/validation/value.rs
```diff
@@ -20,6 +20,22 @@ fn unsupported_type(
...
+    scalar_name: &str,
+    arg_value: &Node<ast::Value>,
+    ty: &Node<ast::Type>,
+) {
+    if let Some(coercer) = schema.custom_scalar_coercer() {
+        if !coercer.validate_ast_value(scalar_name, arg_value) {
+            unsupported_type(diagnostics, arg_value, ty);
+        }
+    }
+}
```
TylerBloom, line 37: nit:
```suggestion
    if schema
        .custom_scalar_coercer()
        .is_some_and(|coercer| !coercer.is_validate_ast_value(scalar, arg_value))
    {
        unsupported_type(diagnostics, arg_value, ty);
    }
```

TylerBloom (commented):
Thanks for your contribution! I have a couple on the design of the new trait's API.

TylerBloom:
@Abdel-Monaam-Aouini, after closer inspection, this would be a breaking change. Since `Schema` has only public fields, any new fields would be a breaking change as a user can (de)construct a `Schema` directly. These changes will be considered in the discussions for a 2.0 release. Given that we do not have immediate plans for this feature, I'm going to close this PR.

PR #1033, opened 2026-04-06
feat!(smith) Backport configurable response generation

SharkBaitDLS:
When we built the Rust subgraph mock server, we started with apollo-smith's response generation code as our foundation. Over time, we've added more configuration flexibility and GraphQL spec-compliant behavior to that code, diverging from this source. This change backports those improvements and fixes and exposes configuration options that will allow the subgraph mock to just take a dependency on this code rather than remaining a fork of it.


crates/apollo-smith/examples/generate_response.rs
```diff
@@ -11,7 +11,7 @@ use std::fs;
 pub fn generate_valid_response(
     doc: &Valid<ExecutableDocument>,
     schema: &Valid<Schema>,
-) -> Result<Value> {
+) -> Result<Value, ResponseError> {
```
tninesling, line 14: This is a breaking change in the public API, so we should mark the PR tag as `feat!`, and we'll want to make a breaking change note in the changelog when we do the release.

crates/apollo-smith/src/random.rs
```diff
@@ -0,0 +1,230 @@
...
+        let raw: u32 = self.arbitrary().map_err(|_| ResponseError::Exhausted)?;
+        let fraction = (raw as f64) / (u32::MAX as f64); // [0.0, 1.0]
+        Ok(min + fraction * (max - min))
+    }
+
+    fn gen_alphanumeric_char(&mut self) -> Result<char, ResponseError> {
+        let idx: u8 = self
+            .int_in_range(0..=61)
+            .map_err(|_| ResponseError::Exhausted)?;
+        Ok(match idx {
```
tninesling: Did you explore any other implementations for this? I would expect to do something simple like `self.choose(ALPHANUM_CHARS)` where we just enumerate the possible values.

crates/apollo-smith/src/random.rs
```diff
@@ -0,0 +1,230 @@
...
+            0..=9 => (b'0' + idx) as char,
+            10..=35 => (b'a' + idx - 10) as char,
+            _ => (b'A' + idx - 36) as char,
+        })
+    }
+
+    fn choose_index(&mut self, len: usize) -> Result<usize, ResponseError> {
```
tninesling, line 82: Why not delegate to `self.choose_index` for this?

crates/apollo-smith/src/random.rs
```diff
@@ -0,0 +1,230 @@
...
+            return Err(ResponseError::InvalidFormat(
+                "cannot choose from empty collection".into(),
+            ));
+        }
+        self.int_in_range(0..=(len - 1))
+            .map_err(|_| ResponseError::Exhausted)
+    }
+
+    fn ratio(&mut self, numerator: u32, denominator: u32) -> Result<bool, ResponseError> {
+        // Unstructured::ratio takes u8, so we scale down if needed.
```
tninesling: Not sure if this is just in recent versions, but `Unstructured::ratio` can handle any `T: Int`, so `u32` should work fine.

crates/apollo-smith/src/random.rs
```diff
@@ -0,0 +1,230 @@
...
+        let (n, d) = if numerator <= u8::MAX as u32 && denominator <= u8::MAX as u32 {
+            (numerator as u8, denominator as u8)
+        } else {
+            // Scale to fit in u8
+            let scale = denominator.max(1) as f64 / 255.0;
+            let n = (numerator as f64 / scale).round() as u8;
+            let d = ((denominator as f64 / scale).round() as u8).max(1);
+            (n, d)
+        };
+        self.ratio(n, d).map_err(|_| ResponseError::Exhausted)
```
tninesling: `Unstructured::ratio` will panic if `n == 0` or `n > d`. It might be worth either documenting that or guarding against it.

crates/apollo-smith/src/random.rs
```diff
@@ -0,0 +1,230 @@
...
+    }
+}
+
+/// Configuration for generating scalar values.
+///
+/// Each variant describes how to generate a value of a particular type using
+/// a [`RandomProvider`]. Register custom scalar configs via
+/// [`ResponseBuilder::with_scalar_config`][crate::ResponseBuilder::with_scalar_config].
+#[derive(Debug, Clone)]
+pub enum ScalarConfig {
```
tninesling: Have you given any thought to how this might be extended for custom scalars?

crates/apollo-smith/src/response.rs
```diff
@@ -80,139 +101,510 @@ impl<'a, 'doc, 'schema> ResponseBuilder<'a, 'doc, 'schema> {
         self
     }
 
-    /// Builds a `Value` matching the shape of `self.doc`
-    pub fn build(mut self) -> Result<Value> {
+    /// Overrides the schema SDL string for federation `_service { sdl }` responses. If not set, `_service`
+    /// will respond with the full SDL for `self.schema` which may include implicit federation types.
```
tninesling: I'm wondering if this belongs here, or if we need a more general mechanism. Generally, we try to keep `apollo-rs` crates narrowly focused on the GraphQL spec without specializing to Federation-specific concepts. Is there a more general hook we could expose without specifying the `_service` field?

crates/apollo-smith/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "apollo-smith"
-version = "0.15.2" # When bumping, also update README.md
+version = "0.16.0" # When bumping, also update README.md
```
tninesling: I think we usually wait to bump this until we cut the release.

crates/apollo-smith/src/response.rs
SharkBaitDLS: How does exposing a configuration option to have a map of arbitrary field overrides to static values sound?

crates/apollo-smith/src/random.rs
SharkBaitDLS: This is defining the generation rules for all the concrete scalar types that GraphQL can support. The interface for ResponseGenerator takes in a `HashMap<Name, ScalarConfig>` exactly so that we can support custom scalars, by mapping them to the concrete scalar type they should generate.

crates/apollo-smith/src/random.rs
SharkBaitDLS: No, most of this (and other similar comments here) are just me being relatively unfamiliar with `Unstructured` compared to `Random`.

crates/apollo-smith/src/random.rs
SharkBaitDLS: This is true for Rand as well so I'll document it at the trait level.

crates/apollo-smith/src/random.rs
tninesling: What if I have some custom scalar which has structure, like a JSON object? I think we'd have to map it to the fixed String generator here, but we'd end up generating some random String with the wrong structure.

I think the consumer of the library needs to be able to write the generation logic, so it might need some `pub trait ScalarGenerator` and that registry signature changes slightly to `HashMap<Name, Box<dyn ScalarGenerator>>` (or similar).

crates/apollo-smith/src/response.rs
tninesling: That sounds like it would fit better. There's already a structured representation of schema coordinates, which might work well for specifying what should be overidden.

crates/apollo-smith/src/random.rs
SharkBaitDLS: That makes sense. I've definitely been thinking about this purely from the angle of generating schema-correct data rather than logically correct data. I'll go down that track.

crates/apollo-smith/src/response.rs
SharkBaitDLS: As I went down the path of putting in custom scalar generators, I realized we can accomplish this same goal with custom object generation instead of overriding specific coordinate paths. I felt like it was a more consistent API so I went that route instead.

crates/apollo-smith/src/response.rs
```diff
@@ -80,139 +122,581 @@ impl<'a, 'doc, 'schema> ResponseBuilder<'a, 'doc, 'schema> {
...
+                .get("content")
+                .unwrap()
+                .as_array()
+                .unwrap();
+
+            for item in content {
+                let typename = item.get("__typename").unwrap().as_str().unwrap();
+                assert_ne!(typename, "Content", "__typename must not be the union name");
+                assert!(
+                    typename == "Post" || typename == "Article",
```
tninesling, line 607: It might be good to check that this only generates the fields corresponding to the selected fragment (i.e. Posts do not have `citations` and Articles do not have `views`)

crates/apollo-smith/src/response.rs
```diff
@@ -80,139 +122,581 @@ impl<'a, 'doc, 'schema> ResponseBuilder<'a, 'doc, 'schema> {
...
+        }
+
+        let mut result = Map::new();
+
+        for (key, fields) in grouped_fields {
+            // The first field is representative for schema-defined metadata (type, nullability, etc.)
+            let meta_field = &fields[0];
+
+            let val = if meta_field.name == TYPENAME {
+                let type_name = if let Some(union_ty) = self
```
tninesling: Do we need something similar for interfaces? I believe those should always resolve to a concrete type when returned from subgraphs

crates/apollo-smith/src/generators/scalar.rs
```diff
@@ -0,0 +1,128 @@
...
+            Name::new_unchecked("Boolean"),
+            DefaultScalarGenerator::Bool.boxed(),
+        ),
+        (
+            Name::new_unchecked("Int"),
+            DefaultScalarGenerator::Int { min: 0, max: 100 }.boxed(),
+        ),
+        (
+            Name::new_unchecked("ID"),
+            DefaultScalarGenerator::Int { min: 0, max: 100 }.boxed(),
```
tninesling: The spec says

> The ID type is serialized in the same way as a String; however, it is not intended to be human-readable. While it is often numeric, it must always serialize as a String.

So I think we should generate a string here, since it's representing a serialized response coming back from a remote server.

crates/apollo-smith/src/random.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    fn gen_alphanumeric_char(&mut self) -> Result<char, ResponseError> {
+        self.choose(ALPHANUM_CHARS)
+            .map_err(|_| ResponseError::Exhausted)
+            .copied()
+    }
+
+    fn choose_index(&mut self, len: usize) -> Result<usize, ResponseError> {
+        self.choose_index(len)
+            .map_err(|_| ResponseError::InvalidFormat("cannot choose from empty collection".into()))
```
tninesling: I think this could either be `EmptyChoose` or `Exhausted`. Would it be worth including an `EmptyChoose` variant and mapping the inner error accordingly?

crates/apollo-smith/src/response.rs
SharkBaitDLS: Good call, this caught that the current implementation actually *doesn't* respect that since the union decision happens after field collection.

crates/apollo-smith/src/response.rs
SharkBaitDLS: Yeah, we didn't have interface support at all for this. It naturally fits alongside the fix for the union issue.

tninesling:
I think this is looking good! The one thing I'm not sure about is the ergonomics of the generator traits. I was having a hard time describing it, so I attempted to show how I was thinking about consolidating things in #1042. It allows a single interface for generating scalars and composite types, plus it removes some of the manual boxing and casting when registering custom generators. Let me know what you think.

tninesling (approved):

PR #1036, opened 2026-05-06
fix(compiler): validate interface field type covariance per GraphQL spec

dariuszkuc:
Add the missing GraphQL spec rule `IsValidImplementationFieldType` to validate that object and interface types implementing an interface have fields whose return types are proper subtypes of the interface fields.

Previously, apollo-compiler only checked that implementing types had all fields required by their interfaces but did not verify type compatibility. This meant schemas like `interface Foo { x: String! }` with `type Bar implements Foo { x: String }` (nullable where non-null is required) would pass validation silently.

The validation covers:
- Non-null covariance (non-null implementation satisfies nullable interface)
- List item type covariance (recursive check on list item types)
- Named type subtyping via `Schema::is_subtype()` (object/interface hierarchy)

GraphQL spec reference:
https://spec.graphql.org/draft/#IsValidImplementationFieldType()

### DIFF
--- a/crates/apollo-compiler/src/validation/diagnostics.rs
+++ b/crates/apollo-compiler/src/validation/diagnostics.rs
@@ -85,6 +85,20 @@ pub(crate) enum DiagnosticData {
         /// Location of the definition of the field in the interface
         field_location: Option<SourceSpan>,
     },
+    #[error(
+        "Interface field {interface}.{field} expects type {interface_type} but {name}.{field} of type {actual_type} is not a proper subtype."
+    )]
+    InvalidImplementationFieldType {
+        name: Name,
+        interface: Name,
+        field: Name,
+        interface_type: Type,
+        actual_type: Type,
+        /// Location of the field in the implementing type
+        field_location: Option<SourceSpan>,
+        /// Location of the field in the interface
+        interface_field_location: Option<SourceSpan>,
+    },
     #[error("the required argument `{coordinate}` is not provided")]
     RequiredArgument {
         name: Name,
@@ -472,6 +486,24 @@ impl DiagnosticData {
                 report.with_help(
                     "An object or interface must declare all fields required by the interfaces it implements",
                 )
+            }
+            DiagnosticData::InvalidImplementationFieldType {
+                name: _,
+                interface,
+                field,
+                interface_type: _,
+                actual_type: _,
+                field_location,
+                interface_field_location,
+            } => {
+                report.with_label_opt(
+                    *field_location,
+                    format_args!("field type is not a proper subtype of `{interface}.{field}`"),
+                );
+                report.with_label_opt(
+                    *interface_field_location,
+                    format_args!("`{interface}.{field}` originally defined here"),
+                );
             }
             DiagnosticData::TransitiveImplementedInterfaces {
                 interface: _,
--- a/crates/apollo-compiler/src/validation/diagnostics.rs
+++ b/crates/apollo-compiler/src/validation/diagnostics.rs
@@ -99,6 +99,47 @@ pub(crate) enum DiagnosticData {
         /// Location of the field in the interface
         interface_field_location: Option<SourceSpan>,
     },
+    #[error(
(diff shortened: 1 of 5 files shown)


duckki (approved):
Looks good to me. Here're some minor suggestions.

- **Tests are missing some important spec cases.** Worth adding to lock down behavior:
  - Interface field returns a **union**; impl returns a member object (the load-bearing case for `is_subtype`'s union branch).
  - Interface field returns interface `I`; impl returns interface `J` that implements `I` (interface-to-interface covariance).
  - Interface implementing another interface (`interface B implements A` with non-covariant field) — currently your first failing test only exercises `type … implements`, not `interface … implements`. Both code paths are wired, but only one is tested.
  - Recursive list case, e.g. `[[Item]!]!` vs `[[Item]]`.

- **Doc comment is slightly off.** [[interface.rs](https://claude.ai/epitaxy/crates/apollo-compiler/src/validation/interface.rs)](crates/apollo-compiler/src/validation/interface.rs) line `Validate that fields in the implementing interface have return types…` — the function is also called from objects, so "implementing type" reads better.

- **Verbose inline types in `validate_implementation_field_types`.** `crate::collections::IndexMap<Name, crate::schema::Component<crate::ast::FieldDefinition>>` is heavy. The file already imports several of these names; pulling `IndexMap`, `Component`, `FieldDefinition` into `use` would match surrounding style.

- **Out of scope, but worth noting:** Some part of the spec's sister rule [`IsValidImplementation`](https://spec.graphql.org/September2025/#IsValidImplementation()) (argument-name parity, argument type *invariance*, no extra non-null args on impl) is still not enforced as far as I can see. Worth a follow-up issue if not tracked already.

duckki:
I tried to follow-up with my own PR: https://github.com/apollographql/apollo-rs/pull/1037
😄

duckki:
The follow-up commit looks good!

crates/apollo-compiler/src/validation/diagnostics.rs
```diff
@@ -85,6 +85,61 @@ pub(crate) enum DiagnosticData {
         /// Location of the definition of the field in the interface
         field_location: Option<SourceSpan>,
     },
+    #[error(
+        "Interface field {interface}.{field} expects type {interface_type} but {name}.{field} of type {actual_type} is not a proper subtype."
```
tninesling: It looks like these have a small style mismatch from the existing error formats, which all start with lowercase. Not a huge deal, but it might be worth it for consistency. Also, the coordinates, field names, and type names should probably be wrapped with backticks.

crates/apollo-compiler/tests/validation/interface.rs
```diff
@@ -210,3 +210,369 @@ scalar Url @specifiedBy(url: "https://tools.ietf.org/html/rfc3986")
...
+type DigitalProduct implements Product {
+  details: ProductDetails
+}
+"#;
+    let errors = Schema::parse_and_validate(input, "schema.graphql")
+        .unwrap_err()
+        .errors
+        .to_string();
+    assert!(
+        errors.contains("Interface field Product.details expects type ProductDetails! but DigitalProduct.details of type ProductDetails is not a proper subtype"),
```
tninesling: Would it be worth showing the full rendering for at least one of these? The statement ending with "is not a proper subtype" makes me ask "subtype of what?", but I suspect the help messages appended under the main diagnostic explain that better.

crates/apollo-compiler/src/validation/diagnostics.rs
dariuszkuc: [265ff4f](https://github.com/apollographql/apollo-rs/pull/1036/commits/265ff4f23725d6bba69ef2b839addf17c9cf8317)

crates/apollo-compiler/tests/validation/interface.rs
dariuszkuc: [265ff4f](https://github.com/apollographql/apollo-rs/pull/1036/commits/265ff4f23725d6bba69ef2b839addf17c9cf8317)

tninesling (approved):

PR #1037, opened 2026-05-06
fix(compiler): validate interface field arguments per GraphQL spec

duckki:
## Summary

Stacks on top of #1036. Implements the argument-related sub-rules of the GraphQL spec rule [`IsValidImplementation`](https://spec.graphql.org/draft/#IsValidImplementation()) in apollo-compiler — raised alongside the existing `IsValidImplementationFieldType` return-type check that #1036 introduces.

The spec requires that, for every argument declared on the implemented (interface) field, the implementing field must declare an argument with the same name and exactly the same type (invariant). The implementing field may declare additional arguments, but each such extra argument must be nullable so callers conforming only to the interface contract are not required to supply it.

Three new diagnostic variants:
- `MissingImplementationFieldArgument` — implementor lacks an argument the interface declares
- `InvalidImplementationFieldArgumentType` — same-named argument has a different type
- `ExtraImplementationFieldArgumentMustBeNullable` — extra argument on the implementing field is non-null

The branch is structured as two commits:
1. A test-only commit that adds `#[ignore]`'d regression tests demonstrating the gap (CI-safe checkpoint).
2. The fix that implements the validation and removes the `#[ignore]` attributes.

The deprecation-propagation sub-rule of `IsValidImplementation` ("if `field` is deprecated then `implementedField` must also be deprecated") is **not** addressed here and remains a separate gap.

## Test plan

- [x] `cargo test -p apollo-compiler` — 282 integration tests, 0 failures
- [x] `cargo clippy -p apollo-compiler --tests -- -D warnings` clean
- [x] Existing `test_data/ok/*.graphql` golden files still validate (no false positives on previously-valid schemas)
- [x] Two negative tests + one positive test for the extra-argument nullability rule

### DIFF
--- a/crates/apollo-compiler/src/validation/diagnostics.rs
+++ b/crates/apollo-compiler/src/validation/diagnostics.rs
@@ -99,6 +99,46 @@ pub(crate) enum DiagnosticData {
         /// Location of the field in the interface
         interface_field_location: Option<SourceSpan>,
     },
+    #[error(
+        "Field `{name}.{field}` is missing the argument `{argument}` declared by interface `{interface}.{field}`."
+    )]
+    MissingImplementationFieldArgument {
+        name: Name,
+        interface: Name,
+        field: Name,
+        argument: Name,
+        /// Location of the field in the implementing type
+        field_location: Option<SourceSpan>,
+        /// Location of the argument in the interface
+        interface_argument_location: Option<SourceSpan>,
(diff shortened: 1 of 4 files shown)


duckki:
Closing in favor of the follow-up commits on the https://github.com/apollographql/apollo-rs/pull/1036.

PR #1041, opened 2026-05-15
fix(compiler): preserve source location for adopted orphan type extensions

dariuszkuc:
When `adopt_orphan_extensions` created a type from orphan extensions (e.g., `extend type T { ... }` without a base `type T` definition), the synthesized `Node` was created via `def.into()` → `Node::new()`, which sets `location: None`. Downstream consumers that rely on node source locations (such as apollo-federation's `HasLocations` trait for composition hints) would silently drop these types from location metadata.

Use the first extension's source location for the synthesized type node via `first_ext.same_location(def)`, matching the pattern used by `from_ast` for base type definitions (`definition.same_location(ty)`).

### DIFF
--- a/crates/apollo-compiler/tests/extensions.rs
+++ b/crates/apollo-compiler/tests/extensions.rs
@@ -189,6 +189,31 @@ fn test_orphan_extensions_kind_mismatch() {
     );
 }
 
+#[test]
+fn test_orphan_extension_preserves_source_location() {
+    let input = r#"
+        extend type T { foo: String }
+        type Query { x: Int }
+    "#;
+
+    let schema = Schema::builder()
+        .adopt_orphan_extensions()
+        .parse(input, "schema.graphql")
+        .build()
+        .unwrap();
+
+    let type_def = &schema.types["T"];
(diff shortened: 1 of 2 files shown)


crates/apollo-compiler/tests/extensions.rs
```diff
@@ -189,6 +189,31 @@ fn test_orphan_extensions_kind_mismatch() {
...
+        .adopt_orphan_extensions()
+        .parse(input, "schema.graphql")
+        .build()
+        .unwrap();
+
+    let type_def = &schema.types["T"];
+    let obj = match type_def {
+        apollo_compiler::schema::ExtendedType::Object(obj) => obj,
+        _ => panic!("expected object type"),
+    };
```
goto-bus-stop:
```suggestion
    let apollo_compiler::schema::ExtendedType::Object(obj) = type_def else {
        panic!("expected object type");
    };
```

nbd but we should be on a rust version that supports this

goto-bus-stop (approved):
makes sense to me 👍🏻

PR #1042, opened 2026-05-15
refactor(smith): consolidate generators for composite types and scalars into a unified Generator trait

tninesling:
Proposed changes for #1033


crates/apollo-smith/src/generators.rs
```diff
@@ -0,0 +1,230 @@
...
+            min_len: 1,
+            max_len: 10,
+        };
+        fallback.generate(rng, self, &empty)
+    }
+}
+
+/// Generates a random boolean.
+#[derive(Debug, Clone)]
+pub struct BooleanGenerator;
```
SharkBaitDLS, line 153: I think I prefer these as an enum from a contract standpoint for discoverability, or perhaps if we keep the structs they should be in a `default` submodule? This is very much just my personal preference though and if this organization is more consistent with the rest of this repo I'm totally fine to keep it.

SharkBaitDLS (approved):

crates/apollo-smith/src/generators.rs
tninesling: Could you explain what you mean by discoverability? As in, you want to communicate what the defaults are for each built-in type?

I think it's more idiomatic to represent defaults through an `impl Default` instead of using a module name. I did that for `Generators`, but we could also do it for each of these individual type generators if that's what you're looking for.

PR #1043, opened 2026-05-18
apollo-smith@0.16.0

SharkBaitDLS:


crates/apollo-smith/CHANGELOG.md
```diff
@@ -18,6 +18,31 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+
+- **Improve byte efficiency for type and field name generation - [tninesling], [pull/1040]**
+- **Backport configurable response generation - [SharkBaitDLS], [tninesling], [pull/1033]**
+  When we built the Rust subgraph mock server, we started with
+  apollo-smith's response generation code as our foundation. Over time,
+  we've added more configuration flexibility and GraphQL spec-compliant
+  behavior to that code, diverging from this source. This change backports
+  those improvements and fixes and exposes configuration options that will
+  allow the subgraph mock to just take a dependency on this code rather
+  than remaining a fork of it.
```
goto-bus-stop: This reads like an internal note, not a user-facing changelog. Suggest summarising what users can now do that they couldn't before and not mentioning the subgraph mock server at all (which i think is not public?)

SharkBaitDLS: It is public. I'm fine with summarizing more though.

SharkBaitDLS:
Moving this back to a draft as Taylor has some bugfixes in-flight that we want to include as part of the 0.16 release.

PR #1045, opened 2026-05-19
fix(smith): ensure unique type names across all type kinds

tninesling:
## Summary
- `DocumentBuilder::type_name` previously could return the same identifier for two definitions of different kinds (e.g. `type A1` and `interface A1`, `union A2` and `input A2`). The GraphQL spec puts all type definitions in a single namespace, so these were validation errors.
- Build a `HashSet` of existing type-like names once and increment a numeric suffix until the candidate is free. Byte-stream layout is preserved; only the chosen suffix shifts (snapshot expectations regenerated).
- Follow-up for #1044 (validate fuzz target). With the validate target running ~60 s, "type defined multiple times" errors drop from ~79 k to ~9 k — an 88% reduction. The remaining count is from extension paths, addressed separately.

### DIFF
--- a/crates/apollo-smith/src/name.rs
+++ b/crates/apollo-smith/src/name.rs
@@ -1,6 +1,5 @@
 use crate::DocumentBuilder;
 use arbitrary::Result as ArbitraryResult;
-use std::collections::HashSet;
 use std::fmt::Write as _;
 
 // First char in a GraphQL name can't be a digit and we don't want it to be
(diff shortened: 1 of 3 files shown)


crates/apollo-smith/src/snapshot_tests.rs
```diff
@@ -17,19 +17,19 @@ fn snapshot_tests() {
           A0
         }
 
-        fragment A2 on A1 {
+        fragment A5 on A2 {
```
goto-bus-stop: I'm surprised fragment and directive names are affected by this change. They don't share the same namespace. Are we using `.type_name()` to generate their names today? We should change that in a separate PR if so.

crates/apollo-smith/src/name.rs
```diff
@@ -80,17 +81,23 @@ impl DocumentBuilder<'_> {
...
-            let _ = write!(
-                new_name,
-                "{}",
-                self.object_type_defs.len() + self.enum_type_defs.len() + self.directive_defs.len()
-            );
+        let base = self.limited_string(30)?;
+        let existing: HashSet<&str> = self
+            .list_existing_type_names()
+            .map(|n| n.name.as_str())
+            .collect();
```
goto-bus-stop: It does a lot more work than before. I guess it would get slow when we have very many types. Maybe we should refactor the type storage to use indexmaps in future so we can look up name collisions directly instead of building a temporary set each time.

crates/apollo-smith/src/name.rs
```diff
@@ -80,17 +81,23 @@ impl DocumentBuilder<'_> {
...
+    /// definition the builder has produced so far. Per the GraphQL spec, all
+    /// type definitions share a single namespace, so a numeric suffix is
+    /// appended and incremented until the candidate is free.
     pub fn type_name(&mut self) -> ArbitraryResult<Name> {
-        let mut new_name = self.limited_string(30)?;
-        if self.list_existing_type_names().any(|n| n.name == new_name) {
-            let _ = write!(
-                new_name,
-                "{}",
-                self.object_type_defs.len() + self.enum_type_defs.len() + self.directive_defs.len()
```
goto-bus-stop, line 90: Honestly, it seems like _this_ is the line that's the root cause. If it just used all the type kinds that share a namespace to compute the suffix, this would be a one-line fix.

It would save a lot of work on every `type_name()` call at the expense of having more arbitrary suffixes.

crates/apollo-smith/src/name.rs
tninesling: That makes sense. I switched back to using the counts of generated types, including the remaining kinds we were missing.

crates/apollo-smith/src/snapshot_tests.rs
tninesling: Yes, that's currently used for fragment names https://github.com/apollographql/apollo-rs/blob/main/crates/apollo-smith/src/fragment.rs#L159

crates/apollo-smith/src/name.rs
```diff
@@ -80,15 +80,25 @@ impl DocumentBuilder<'_> {
         Ok(Name::new(self.limited_string(30)?))
     }
 
-    /// Create an arbitrary type `Name`
+    /// Create an arbitrary type `Name`. Per the GraphQL spec, all type
+    /// definitions share a single namespace; if the randomly drawn base name
+    /// collides with anything already generated, append a numeric suffix
+    /// derived from the running count of every kind of definition the
+    /// builder has produced. The suffix grows monotonically across calls, so
+    /// disambiguation stays a constant-time addition rather than a search.
```
goto-bus-stop:
```suggestion
    /// Create an arbitrary type `Name` that does not yet exist in the document.
```

Documenting intent of the function not the exact implementation

goto-bus-stop (approved):

PR #1046, opened 2026-05-19
fix(smith): valid type kinds in input positions

tninesling:
## Summary
- Argument types, input-object fields, and variable definitions previously drew from `list_existing_types`, which only returns object + enum types. Object types are not valid in input positions, so most input definitions failed validation.
- New `list_existing_input_types` returns custom scalars, enums, and input objects (built-in scalars are still added by `choose_ty`). Used at the three input-position call sites: `input_value_def`, `input_values_def`, and `variable_definition`.
- `input_value_for_type` gains branches for custom scalars (produces a string literal) and input objects (recurse over declared fields). The existing object-type branch stays as a defensive fallback.
- Stacked on #1045. With the validate fuzz target running ~60 s, `must be of an input type` errors drop from 26,032 to 0; no new categories surfaced.

## Test plan
- [x] New `crates/apollo-smith/tests/input_type_kinds.rs` covers the empty seed and a range of small sequential seeds; both fail without the fix and pass with it.
- [x] `cargo test -p apollo-smith` — all existing tests pass (snapshot expectations regenerated for the shifted type pool).
- [x] `cargo +nightly fuzz run -s none validate` — no `must be of an input type` diagnostics across 134 k iterations.

### DIFF
--- a/crates/apollo-smith/src/input_value.rs
+++ b/crates/apollo-smith/src/input_value.rs
@@ -236,12 +236,42 @@ impl DocumentBuilder<'_> {
                 Ok(InputValue::Enum(
                     doc_builder.arbitrary_variant(&enum_)?.clone(),
                 ))
+            } else if let Some(input_object_ty) = doc_builder
+                .input_object_type_defs
+                .iter()
+                .find(|io| &io.name == ty.name())
+                .cloned()
+            {
+                Ok(InputValue::Object(
+                    input_object_ty
+                        .fields
+                        .iter()
+                        .map(|field_def| {
+                            Ok((
+                                field_def.name.clone(),
+                                doc_builder.input_value_for_type(&field_def.ty)?,
+                            ))
+                        })
+                        .collect::<ArbitraryResult<Vec<_>>>()?,
+                ))
+            } else if doc_builder
+                .scalar_type_defs
+                .iter()
+                .any(|s| &s.name == ty.name())
+            {
+                // Custom scalars accept any literal value; a String is always
+                // syntactically valid.
+                Ok(InputValue::String(doc_builder.limited_string(40)?))
             } else if let Some(object_ty) = doc_builder
                 .object_type_defs
                 .iter()
                 .find(|o| &o.name == ty.name())
                 .cloned()
             {
+                // Object types are not valid input types, but keep this branch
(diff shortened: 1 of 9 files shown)


crates/apollo-smith/src/input_value.rs
```diff
@@ -236,12 +236,42 @@ impl DocumentBuilder<'_> {
...
+                        .collect::<ArbitraryResult<Vec<_>>>()?,
+                ))
+            } else if doc_builder
+                .scalar_type_defs
+                .iter()
+                .any(|s| &s.name == ty.name())
+            {
+                // Custom scalars accept any literal value; a String is always
+                // syntactically valid.
+                Ok(InputValue::String(doc_builder.limited_string(40)?))
```
goto-bus-stop: Maybe it's worth using an integer for scalars to consume less entropy? Or use a random value of a random type.

crates/apollo-smith/src/input_value.rs
```diff
@@ -236,12 +236,42 @@ impl DocumentBuilder<'_> {
...
             } else if let Some(object_ty) = doc_builder
                 .object_type_defs
                 .iter()
                 .find(|o| &o.name == ty.name())
                 .cloned()
             {
+                // Object types are not valid input types, but keep this branch
+                // as a defensive fallback in case one slips into an input
+                // position via a path that doesn't yet use
+                // `list_existing_input_types`.
```
goto-bus-stop: Seems more appropriate to panic here?

crates/apollo-smith/src/input_value.rs
tninesling: Updated in https://github.com/apollographql/apollo-rs/pull/1046/commits/34444693c7810bd0317421e00bf6337eca7b6ea7

crates/apollo-smith/src/input_value.rs
tninesling: Updated to panic. This was keeping a unit test passing below, which was explicitly creating a Object type as an input. The test now uses input types correctly (per https://github.com/apollographql/apollo-rs/pull/1046/commits/34444693c7810bd0317421e00bf6337eca7b6ea7)

goto-bus-stop (approved):
Thanks!

PR #1050, opened 2026-05-28
fix(smith): generate valid interface inheritance

tninesling:
##  Summary

`apollo-smith`'s interface generation produced documents that failed validation in four ways the fuzzer kept hitting:
- A type that implements interface B where B implements A did not also declare implements A (transitive implements rule).
- An implementer's field signature could disagree with the parent interface's (covariance / matching argument types).
- An implementer could omit an interface field entirely, or add a required argument the interface didn't declare.
- An extend interface block could never pick new implements clauses, so transitive inheritance through extensions was unreachable.

This PR fixes all four. 

## What changed

- New `ImplementsGraph` using `petgraph::DiGraph<Name, ()>` with one edge per `X implements Y`. This is used to find the closure of interfaces that need to be implemented
- When picking interfaces to implement for an in-progress definition, we check the entire closure of implied interfaces to ensure it will not cause cycles or field conflicts
- After all definitions are generated, fields are backfilled from any implemented interfaces since a type `X implements Y` must have all fields from `Y` (or fields compatible with `Y`, based on covariance rules). The previous step ensures all implemented interfaces have compatible fields that can be merged in.
- There is also logic to account for interfaces being implemented by different extensions to a base definition, which involves some iteration of definitions with the same name and tracking indices.


crates/apollo-smith/src/interface.rs
```diff
@@ -168,23 +174,281 @@ impl DocumentBuilder<'_> {
...
+                &fields_by_type_name,
+                &mut accepted,
+                &mut accumulated_signatures,
+                self_name,
+            );
+        }
+        Ok(accepted
+            .into_iter()
+            .filter(|n| !already_implemented_parents.contains(n))
+            .collect())
```
TylerBloom: nit: use `retain()`
```suggestion
        accepted.retain(|n| !already_implemented_parents.contains(n));
        Ok(accepted)
```

crates/apollo-smith/src/implements_graph.rs
```diff
@@ -0,0 +1,78 @@
...
+    by_name: HashMap<Name, NodeIndex>,
+}
+
+impl ImplementsGraph {
+    pub(crate) fn new() -> Self {
+        Self::default()
+    }
+
+    /// Record `from implements to`. Idempotent.
+    pub(crate) fn add_edge(&mut self, from: &Name, to: &Name) {
```
TylerBloom, line 24: Is there a reason we don't try to up hold the acyclic invariant in this method? Doing so should help confine the cycle-checking logic to this module. Functions like `try_accept_candidate` wouldn't be needed.

crates/apollo-smith/src/implements_graph.rs
```diff
@@ -0,0 +1,78 @@
...
+use petgraph::visit::Bfs;
+use petgraph::visit::Reversed;
+use std::collections::HashMap;
+
+/// Directed graph of `implements` edges (`X implements Y` ⇒ edge `X →
+/// Y`). Holds both interface→interface and object→interface edges.
+#[derive(Debug, Clone, Default)]
+pub(crate) struct ImplementsGraph {
+    graph: DiGraph<Name, ()>,
+    by_name: HashMap<Name, NodeIndex>,
```
TylerBloom, line 15: This there a reason you are using a `DiGraph` + `HashMap` over something like a `HashMap<Name, IndexSet<Name>>`. Provided you maintain the acyclic invariant, this becomes, IMO, easier to parse while keeping `closure`s and topo. ordering easy to calculate.

TylerBloom (commented):
This might not be an issue, but it is worth noting that these changes introduce a fair number of ephemeral allocations.

crates/apollo-smith/src/implements_graph.rs
tninesling: We could do the cycle check in here. I'll have to give it a bit more thought. Although, that would not totally remove the need for `try_accept_candidate` because there are other invariants we need to uphold. That function also ensures we do not inherit conflicting interfaces.

That would be something like
```graphql
interface I1 {
  conflictingField: String!
}

interface I2 {
  conflictingField: Int!
}

type A implements I1 {
  conflictingField: String!
}
```
where the generator tries to add `I2` to the interface list of `A`.

crates/apollo-smith/src/implements_graph.rs
tninesling: The main reason is that I didn't want to implement the algorithms for `toposort` and `closure` when I know they come out of the box in `petgraph`. I can certainly try to do that if you feel strongly about not having the dependency.

TylerBloom: I don't have a strong feeling. Adding `petgraph` isn't too bad. That said, `closure` and `toposort` (given that we maintain the acyclic invariant) are the same algorithm with the multi-map using `IndexMap` approach. You would do a recursive, depth-first fetch + `shift_insert` on the `IndexSet` that is being returned.

crates/apollo-smith/src/interface.rs
tninesling: Update in https://github.com/apollographql/apollo-rs/pull/1050/commits/d63cdfcc0e16506c8c129582b404932f67912f75

TylerBloom (approved):

PR #1052, opened 2026-05-29
fix(smith): restrict union members to object types

tninesling:
## Summary

`apollo-smith` generated unions whose members weren't Object base types, producing documents the validator rejected with errors like `union member Int must be an object type` (and the same for `Float`, `String`, `Boolean`, `ID`, and other union names).

The generator built its member candidate pool from three sources:

1. `list_existing_object_types()` — valid.
2. `.extend(self.union_type_defs ...)` — added other union names, which can't be union members.
3. `choose_named_ty()`'s implicit fallback to `BUILTIN_SCALAR_NAMES` — added `Int`/`Float`/`String`/`Boolean`/`ID`, none of which are Object types.

Sources (2) and (3) both produce invalid documents. Pick directly from `list_existing_object_types()` instead.

See <https://spec.graphql.org/October2021/#sec-Unions>.

### DIFF
--- a/crates/apollo-smith/src/union.rs
+++ b/crates/apollo-smith/src/union.rs
@@ -2,7 +2,6 @@ use crate::description::Description;
 use crate::directive::Directive;
 use crate::directive::DirectiveLocation;
 use crate::name::Name;
-use crate::ty::Ty;
 use crate::DocumentBuilder;
 use apollo_compiler::ast;
 use arbitrary::Result as ArbitraryResult;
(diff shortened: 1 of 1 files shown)


crates/apollo-smith/src/union.rs
```diff
@@ -134,15 +133,14 @@ impl DocumentBuilder<'_> {
...
-
+        // Union members must be Object base types — built-in scalars,
+        // other unions, interfaces, enums, and input objects are all
+        // invalid as members.
+        //
+        // See <https://spec.graphql.org/October2021/#sec-Unions>.
+        let object_types = self.list_existing_object_types();
         let members = (0..self.u.int_in_range(2..=10)?)
-            .map(|_| Ok(self.choose_named_ty(&existing_types)?.name().clone()))
+            .map(|_| Ok(self.u.choose(&object_types)?.name().clone()))
```
TylerBloom, line 143: In the cases where `choose_named_ty` fallback on built in types, there were no object types in the schema. We can still reach that point, but now that fallback is an error. Do we need to do anything to mitigate that?

tninesling: The top-level generation logic always generates between 1 and 50 object types in the schema, so we will always have at least one https://github.com/apollographql/apollo-rs/blob/main/crates/apollo-smith/src/lib.rs#L164

If there were no object types, then this would propagate the empty choose error and the fuzzer would move on to a different seed.

TylerBloom: Ah, I see. the `choose_named_ty` doesn't "fallback" on the built-in. It just adds them to the set that can be choosen.

TylerBloom (approved):

PR #1057, opened 2026-06-24
fix(smith)!: prevent extensions from duplicating existing values

tninesling:
When we generate a new extension for an existing enum, we now gather the existing values from all previous definitions and exclude them from the in-progress extension. We apply equivalent logic for preventing duplicate union members and duplicate fields in objects, interfaces, and input definitions.

After we deduplicate, it's possible that we have produced an empty definition. In that case, we return an `IncorrectFormat` error so the fuzzer moves on to another byte string.

### DIFF
--- a/crates/apollo-smith/src/input_value.rs
+++ b/crates/apollo-smith/src/input_value.rs
@@ -290,6 +290,7 @@ impl DocumentBuilder<'_> {
     pub fn input_values_def(
         &mut self,
         directive_location: DirectiveLocation,
+        exclude: &[&Name],
     ) -> ArbitraryResult<Vec<InputValueDef>> {
         let arbitrary_iv_num = self.u.int_in_range(2..=5usize)?;
         let mut input_values = Vec::with_capacity(arbitrary_iv_num - 1);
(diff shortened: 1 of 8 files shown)


crates/apollo-smith/src/input_value.rs
```diff
@@ -290,6 +290,7 @@ impl DocumentBuilder<'_> {
     pub fn input_values_def(
         &mut self,
         directive_location: DirectiveLocation,
+        exclude: &[&Name],
```
TylerBloom: While this follows the pattern established with methods like `DocumentBuilder::fields_definition`, I think `exclude` should be a `&[Name]`. At the call sites of this method (and `fields_definition`), this being a slice of references to names forces two allocations.

Alternatively, this could be an `IndexSet`, like what was done with `DocumentBuilder::enum_values_definition`.

tninesling: I updated those call sites with `exclude` to be `IndexSet<Name>` in https://github.com/apollographql/apollo-rs/pull/1057/commits/a3df5896f94b5879b95ffe93d59e941b813541fb. I also updated the `fields_definition` one as well so we can try to keep the API consistent

TylerBloom (approved):

PR #1059, opened 2026-06-24
fix(smith): disable alias generation to avoid field selection merging conflicts

tninesling:
Aliases could produce response-name collisions where two fields share the same response name but select different schema fields, violating the Field Selection Merging rule. Generating a valid alias must account for aliases on other fields as well as fragments which could be spread over the current type. I attempted to resolve the conflicting aliases, but the logic seemed more complicated than it was worth.

In the spirit of quickly ensuring we output valid schemas, this disables alias generation entirely for now rather than adding complex deduplication logic.

https://spec.graphql.org/October2021/#sec-Field-Selection-Merging

### DIFF
--- a/crates/apollo-smith/src/field.rs
+++ b/crates/apollo-smith/src/field.rs
@@ -162,7 +162,7 @@ impl DocumentBuilder<'_> {
     }
 
     /// Create an arbitrary `Field` given an object type
-    pub fn field(&mut self, index: usize) -> ArbitraryResult<Field> {
+    pub fn field(&mut self, _index: usize) -> ArbitraryResult<Field> {
         let fields_defs = self
             .stack
             .last()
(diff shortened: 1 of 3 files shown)


TylerBloom (commented):
Do we have an idea of how much work will be required to reintroduce this feature?

tninesling:
> Do we have an idea of how much work will be required to reintroduce this feature?

I'm not totally sure how much effort it would be. I think part of the fix is maintaining the persistent global map of which aliases are assigned to which fields, similar to what I did for arguments in the [fragment fix PR](https://github.com/apollographql/apollo-rs/pull/1060). That way, we would at least be able to know if a fragment we generated implies we need some alias on the corresponding object field.

I think the tricky part is handling abstract types, and this is where my GraphQL knowledge is lacking. I think that we would need to pull generated aliases in from implemented interfaces as we go, but I'm not sure of the nuance there.

If you think it's better, we could try the fix for fragments on concrete types and skip the abstract type case for now.

TylerBloom:
> If you think it's better, we could try the fix for fragments on concrete types and skip the abstract type case for now.

No, I think this will work for now. We just need to track this future work and any the potential pitfalls and fixes.

TylerBloom (approved):

PR #1062, opened 2026-06-25
fix(smith)!: prevent non-null self-references in input objects

tninesling:
## Summary
- When generating input object fields, if the chosen type is a non-null self-reference (e.g. `MyInput!` as a field on `MyInput`), strip the `NonNull` wrapper to make it nullable. Non-null self-references are invalid GraphQL because the type becomes impossible to construct.
- Adds a `self_name` parameter to `input_values_def` so callers can indicate which input object is being defined. Arguments pass `None` since they aren't subject to self-referential constraints.

Note: this prevents the case where we generate `input A { field: A! }` by removing the non-null requirement to break the cycle. There is still an unhandled possible case where we end up with `input A { field: B!}` and `input B { field: A! }`.

### DIFF
--- a/crates/apollo-smith/src/input_value.rs
+++ b/crates/apollo-smith/src/input_value.rs
@@ -292,6 +292,7 @@ impl DocumentBuilder<'_> {
         &mut self,
         directive_location: DirectiveLocation,
         exclude: &IndexSet<Name>,
+        self_name: Option<&Name>,
     ) -> ArbitraryResult<Vec<InputValueDef>> {
         let arbitrary_iv_num = self.u.int_in_range(2..=5usize)?;
         let mut input_values = Vec::with_capacity(arbitrary_iv_num - 1);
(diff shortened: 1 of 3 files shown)


crates/apollo-smith/src/input_value.rs
```diff
@@ -292,6 +292,7 @@ impl DocumentBuilder<'_> {
         &mut self,
         directive_location: DirectiveLocation,
         exclude: &IndexSet<Name>,
+        self_name: Option<&Name>,
```
TylerBloom, line 295: For the "`A` needs `B!` needs `A!`" problem, could we just pass a set of names here rather than a single name?

TylerBloom (approved):

crates/apollo-smith/src/input_value.rs
tninesling: Effectively, yes. Although there is more bookkeeping to be done when an extension mutates an existing definition. I'll draft that change as a separate branch/ticket from this one.

PR #1063, opened 2026-06-25
perf(smith): cache used type names to avoid O(n²) rebuilds

tninesling:
## Summary
- `type_name()` previously rebuilt a `HashSet` of all existing type names from 9 definition collections on every call, making total name generation O(n²)
- Maintain a persistent `HashSet<String>` on `DocumentBuilder` that is updated incrementally as names are created
- Reuse the `String` allocation in the collision-resolution loop via `write!` instead of `format!`

In a small local benchmark, this showed a 42% improvement, generating ~24k docs/sec compared to ~14k docs/sec before the change.

### DIFF
--- a/crates/apollo-smith/src/lib.rs
+++ b/crates/apollo-smith/src/lib.rs
@@ -27,6 +27,7 @@ pub(crate) mod union;
 pub(crate) mod variable;
 
 use indexmap::IndexMap;
+use std::collections::HashSet;
 use std::fmt::Debug;
 
 #[derive(Debug, Clone, thiserror::Error)]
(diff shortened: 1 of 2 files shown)


TylerBloom (approved):
Excellent! We should look into using `Arc<str>` for names that are static after they are created.

tninesling:
> Excellent! We should look into using `Arc<str>` for names that are static after they are created.

Definitely! Although I don't quite understand why we don't just use `apollo_compiler::Name` which is already doing that under the hood. Perhaps that's the way to go

TylerBloom:
> I don't quite understand why we don't just use apollo_compiler::Name

I was wondering that myself. I only recently realized that we are using a different `Name` type here.

PR #1065, opened 2026-07-13
feat(compiler): allow parsing FieldSet from a source span in an existing document

tninesling:
# Parsing FieldSet from within an already-parsed document

We support parsing "free-standing" FieldSet instances for Federation, to support selections which are embedded in String-type directive arguments. So far, using `FieldSet::parse_and_validate()` cannot produce the rich diagnostic messages we get from the initial parser pass because that entrypoint creates a fake source file document from the string passed into the parser.

This change adds support for parsing a FieldSet by giving the parser a source span inside an existing schema, resulting in rich diagnostics about that selection within the context of the actual schema. To support this, this adds `FieldSet::parse_and_validate_at_span()`. The main difference from the existing `parse_and_validate()` is that we pass the source span instead of a string, and that span is used to locate the string content in the schema. Previously, the `Convert` API for turning the CST into an AST only accepted a `FileId` for source information. The bulk of this change is a mechanical substitution of those parameters for a new `SourceContext` which carries the `FileId` and an offset. These are internal APIs, so the change is contained to this project. The only public-facing change is the addition of `FieldSet::parse_and_validate_at_span()`.

### DIFF
--- a/crates/apollo-compiler/src/parser.rs
+++ b/crates/apollo-compiler/src/parser.rs
@@ -23,7 +23,6 @@ use crate::validation::Valid;
 use crate::validation::WithErrors;
 use crate::ExecutableDocument;
 use crate::Schema;
-use apollo_parser::SyntaxNode;
 use rowan::TextRange;
 use serde::Deserialize;
 use serde::Serialize;
@@ -162,7 +161,23 @@ impl Parser {
         errors: &mut DiagnosticList,
         parse: impl FnOnce(apollo_parser::Parser) -> apollo_parser::SyntaxTree<T>,
     ) -> apollo_parser::SyntaxTree<T> {
-        let mut parser = apollo_parser::Parser::new(&source_text);
+        let tree = self.parse_syntax(&source_text, parse);
+        let source_file = Arc::new(SourceFile {
+            path,
+            source_text,
+            source: OnceLock::new(),
+        });
+        Arc::make_mut(&mut errors.sources).insert(file_id, source_file);
+        Self::collect_parse_errors(&tree, file_id, 0.into(), errors);
+        tree
+    }
+
+    fn parse_syntax<T: apollo_parser::cst::CstNode>(
+        &mut self,
+        source_text: &str,
+        parse: impl FnOnce(apollo_parser::Parser) -> apollo_parser::SyntaxTree<T>,
+    ) -> apollo_parser::SyntaxTree<T> {
+        let mut parser = apollo_parser::Parser::new(source_text);
         if let Some(value) = self.recursion_limit {
             parser = parser.recursion_limit(value)
         }
@@ -172,25 +187,28 @@ impl Parser {
         let tree = parse(parser);
         self.recursion_reached = tree.recursion_limit().high;
         self.tokens_reached = tree.token_limit().high;
-        let source_file = Arc::new(SourceFile {
-            path,
-            source_text,
-            source: OnceLock::new(),
-        });
-        Arc::make_mut(&mut errors.sources).insert(file_id, source_file);
+        tree
+    }
+
+    fn collect_parse_errors<T: apollo_parser::cst::CstNode>(
+        tree: &apollo_parser::SyntaxTree<T>,
+        file_id: FileId,
+        base_offset: rowan::TextSize,
+        errors: &mut DiagnosticList,
+    ) {
         for parser_error in tree.errors() {
             // Silently skip parse errors at index beyond 4 GiB.
             // Rowan in apollo-parser might complain about files that large
             // before we get here anyway.
-            let Ok(index) = parser_error.index().try_into() else {
+            let Ok(index): Result<rowan::TextSize, _> = parser_error.index().try_into() else {
                 continue;
             };
-            let Ok(len) = parser_error.data().len().try_into() else {
+            let Ok(len): Result<rowan::TextSize, _> = parser_error.data().len().try_into() else {
                 continue;
             };
             let location = Some(SourceSpan {
                 file_id,
-                text_range: rowan::TextRange::at(index, len),
+                text_range: rowan::TextRange::at(index + base_offset, len),
             });
             let details = if parser_error.is_limit() {
                 Details::ParserLimit {
@@ -203,7 +221,6 @@ impl Parser {
             };
             errors.push(location, details)
         }
-        tree
     }
 
     /// Parse the given source text as the sole input file of a schema.
@@ -350,7 +367,8 @@ impl Parser {
         path: impl AsRef<Path>,
     ) -> (executable::FieldSet, DiagnosticList) {
         let file_id = FileId::new();
-        let mut errors = DiagnosticList::new(Default::default());
+        let sources = IndexMap::clone(&schema.sources);
+        let mut errors = DiagnosticList::new(Arc::new(sources));
         let tree = self.parse_common(
             source_text.into(),
             path.as_ref().to_owned(),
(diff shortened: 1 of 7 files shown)


TylerBloom (commented):
We might want to add a couple more tests for this behavior. Otherwise, the logic looks good.

TylerBloom (approved):

goto-bus-stop:
Note in the past I've been a bit resistant to this because a simple "start offset" does not cover all cases. for example, the string might contain escape sequences, and then the indices reported in node metadata and diagnostics will be wrong (hopefully not in a way where an index is not on a UTF-8 character boundary!).

I'm not opposed to a half-solution along these lines, if we want to _improve_ UX and accept that escape sequences will cause wrong outputs, but we should be very careful about it. Especially we should make sure that we don't introduce any DOS vectors with that UTF-8 problem^.

crates/apollo-compiler/src/parser.rs
```diff
@@ -359,25 +377,66 @@ impl Parser {
...
         let field_set = executable::FieldSet {
             sources: errors.sources.clone(),
             selection_set,
         };
         (field_set, errors)
     }
 
+    /// Parses the field set at `source_span` within an already-parsed source
+    /// file. The text is extracted from the schema's sources and errors point
+    /// to the original location. If the span covers a quoted string literal,
```
goto-bus-stop: I see it does this for single quotes. What about """block strings"""?

crates/apollo-compiler/src/parser.rs
```diff
@@ -359,25 +377,66 @@ impl Parser {
...
+    /// the quotes are stripped automatically.
+    pub(crate) fn parse_field_set_at_span(
+        &mut self,
+        schema: &Valid<Schema>,
+        type_name: ast::NamedType,
+        source_span: SourceSpan,
```
goto-bus-stop, line 414: I didn't see before commenting, but this is a good API shape: this lets us add support for escape sequences down the line because it doesn't take an input string.

crates/apollo-compiler/src/parser.rs
tninesling: Good call-out. I added support for block strings in https://github.com/apollographql/apollo-rs/pull/1065/commits/75589f95f283a5b7407b0232c692ba7e551122ff, and instead of falling back to just parsing whatever's in the span, it now rejects spans which aren't enclosed by single or triple quotes.

I also added a test with some utf-8 characters. Let me know if that assuages the concerns from your top-level comment.

goto-bus-stop:
I was thinking of a string like this:

```
@sel(fields: "id\u0020한글")
```
Here we have a escape sequence for the space character followed by two multi-byte characters (3 bytes each). If we naively treat the escape sequence as length 1, because it's a single space, then the error for the character 한 will be off by 5 bytes (`\u0020` is 6 bytes but ` ` is 1), meaning it's in the middle of the multi-byte character 글. (e; it might go the other way around...but point is as long as you have enough multi-byte characters following the escape sequence you can get into this situation)

I tried to write a test for this but this solution doesn't support escape sequences in strings. This means we definitely don't get a DOS but it also means that we can't parse some strings that are valid. I don't know if that would be an acceptable compromise for us--I wouldn't be surprised if a customer has written a \n in their `@requires` for example.

(This is why I was a bit resistant to a simple solution in the past: to support all GraphQL strings, we need to use the already-parsed value as input to the parser, but that means that a simple base offset + offset-in-string is not always going to point to the right place.)

tninesling:
> I was thinking of a string like this:
> 
> ```
> @sel(fields: "id\u0020한글")
> ```
> 
> Here we have a escape sequence for the space character followed by two multi-byte characters (3 bytes each). If we naively treat the escape sequence as length 1, because it's a single space, then the error for the character 한 will be off by 5 bytes (`\u0020` is 6 bytes but ` ` is 1), meaning it's in the middle of the multi-byte character 글. (e; it might go the other way around...but point is as long as you have enough multi-byte characters following the escape sequence you can get into this situation)
> 
> I tried to write a test for this but this solution doesn't support escape sequences in strings. This means we definitely don't get a DOS but it also means that we can't parse some strings that are valid. I don't know if that would be an acceptable compromise for us--I wouldn't be surprised if a customer has written a \n in their `@requires` for example.
> 
> (This is why I was a bit resistant to a simple solution in the past: to support all GraphQL strings, we need to use the already-parsed value as input to the parser, but that means that a simple base offset + offset-in-string is not always going to point to the right place.)

We're just using the existing parse logic for the content of the string. The only byte offset we're considering is to add 1 when we see `"`, so we get the content of the string itself. I tried out your example as well, and it fails to parse both because we do not decode the escape sequence and the actual utf-8 characters are invalid (although I believe they will become valid when we start supporting the latest version of the spec).

I'm failing to see why it would be a DOS vector, but in the context of Federation, this will run at composition time to validate the selections are valid. I'm inclined to merge this as-is, but let me know if there are any concrete improvements I can make before then.

goto-bus-stop:
> We're just using the existing parse logic for the content of the string.

Not exactly...Compare it to what you would be doing before: you'd have a parsed document, and pass a `ast::Value::String`'s _contents_ to the `parse_field_set` function. The contents of a string in the AST doesn't contain escape sequences, because those are parsed when we build the AST.

If you replace `parse_field_set(value.to_string())` by `parse_field_set_at_span(source, value.span())`, then escape sequences cause parse errors, where they didn't in the past. Previously you were parsing the contents of the string, now you're parsing the source code form of the string.

Again I don't know if it is a big deal to Federation, but it _is_ incorrect.

> I'm failing to see why it would be a DOS vector

If we do handle escape sequences correctly, but don't do full escape-sequence-aware translation of source offsets, then you can get diagnostics that point to the middle of a UTF-8 multi-byte sequence, which can panic (we had another case of this in #1023). Arguably not a DOS vector for federation (or for the router in general) as we don't parse selections from end-user input, only from schemas provided by the customer. But in general I don't think apollo-compiler should panic on regular use of its APIs, that's why we have fuzzing and all that.

Here's an example of a panic on user input if we support escape sequences but don't do full source mapping: https://github.com/apollographql/apollo-rs/compare/tninesling/field-set-source-merge...renee/field-set-parse-panic#diff-db12797cd603794eace3e3351c1cff57d959ab79c25f60fb263f64ee5b793124R236
(first commit parses the string correctly (ignoring some detail around block strings), second commit shows the panic)

goto-bus-stop:
> I'm inclined to merge this as-is, but let me know if there are any concrete improvements I can make before then.

I think this is a judgement call for federation to make, if it is an acceptable limitation (i.e. we _know_ that nobody is using escape sequences and we're willing to deal with it if a customer tries to do `\n` and is surprised that it doesn't work).

I think at the very least the limitations of the method should be documented, maybe it should even be `#[doc(hidden)]` if we don't support escape sequences.

It's certainly _better_ to not support escape sequences at all than to support them incorrectly :)

tninesling:
> Here's an example of a panic on user input if we support escape sequences but don't do full source mapping

That's good to know. I tried it on my branch without any extra parsing changes, and it just reports the escape sequence as a syntax error, which is what we want for now. This will probably come back up when updating to the latest GQL spec, since it adds support for escaped unicode. I have added a note on the internal tracking ticket and linked your comment/example accordingly so we can revisit then.

goto-bus-stop:
> This will probably come back up when updating to the latest GQL spec, since it adds support for escaped unicode.

Escape sequences are already supported everywhere else in apollo-rs and in older versions of the GraphQL spec (eg in the [2021 version](https://spec.graphql.org/October2021/#sec-String-Value)). That's why I suggested documenting this or even doc-hiding the `parse_field_set_at_span` method: it's incorrect and incomplete as written, in a way that might be acceptable to Federation, but not in general

Concretely if a user writes `@key(fields: "a\nb")`, it would have worked in all previous versions of apollo-rs and graphql-js, and will start reporting a syntax error if federation adopts this method. That's the tradeoff.

tninesling:
>  Concretely if a user writes @key(fields: "a\nb"), it would have worked in all previous versions of apollo-rs and graphql-js, and will start reporting a syntax error if federation adopts this method.

I see. I thought it was only to do with unicode escape sequences, but that makes sense. I confirmed that's the case, and while we would rule it out at composition time for new versions of composition, it would potentially land in a planner running graphs from older composition versions.

I'm backing this change out for now, so I can revisit it with proper parsing fixes.

PR #1069, opened 2026-07-24
feat(compiler): validate @defer directive usage

duckki:
Add validation for the `@defer` directive rules from the GraphQL Defer & Stream
draft ([graphql-spec#1110](https://github.com/graphql/graphql-spec/pull/1110)). apollo-compiler previously had no `@defer`
validation, so operations that the spec forbids passed silently.

The feature is still draft-only (not in the released spec), but it is enabled by
default in Apollo Router via `incremental_delivery.enable_defer`, so invalid
operations reach the planner in practice.

Three rules are validated, matching graphql-js's `specifiedRules`:

- **Labels are unique** — every `@defer(label:)` value must be a static string
  (not a variable) and unique across the document.
- **Not on root fields** — `@defer` is not allowed on root selections of
  `mutation` or `subscription` operations, recursing through fragment spreads
  and inline fragments at the root level.
- **Disable-able in subscriptions** — inside a `subscription`, any `@defer` not
  statically removed by `@skip`/`@include` must be disable-able via an `if`
  argument set to `false` or a variable. A `@skip`/`@include` with a missing or
  variable `if` is conservatively treated as "may be excluded", so we do not
  flag `@defer` beneath it.

Note: the spec PR's formal step for the third rule reads *"`argumentValue` must
not be the boolean value `{false}`"*, which contradicts its own explanatory text
and example. We follow graphql-js (and the explanatory text): forbid a literal
`if: true` or a missing `if`.

Because `@defer` is not a built-in directive in apollo-compiler, the new test
schemas declare it explicitly. `@stream` is out of scope here.

graphql-js reference rules:
- https://github.com/graphql/graphql-js/blob/main/src/validation/rules/DeferStreamDirectiveLabelRule.ts
- https://github.com/graphql/graphql-js/blob/main/src/validation/rules/DeferStreamDirectiveOnRootFieldRule.ts
- https://github.com/graphql/graphql-js/blob/main/src/validation/rules/DeferStreamDirectiveOnValidOperationsRule.ts


duckki:
Currently, Lint step is failing as summarized as following:

## Summary

The break is a **rustfmt ICE in the `2026-07-23` nightly** that CI pulls fresh each run.

## Options to unblock (none involve changing the `@defer` code)

1. **Re-run Lint in a day or two** — rustfmt ICEs are usually reverted/fixed in the next nightly; CI does `rustup install nightly` fresh each run, so a later run picks up the fix automatically. Lowest-effort, nothing to change.
2. **Pin nightly rustfmt to a known-good date in a separate infra PR** — edit `.circleci/config.yml` (`rustup install nightly-2026-04-22` or similar). Unblocks immediately and unblocks the whole repo, but it's a repo-wide change that maintainers should own, not something to smuggle into this feature PR.

tninesling:
Hey @duckki! I merged a small mitigation for the `rustfmt` panic. Merging `main` should unblock CI for you.

crates/apollo-compiler/src/validation/operation.rs
```diff
@@ -174,6 +177,291 @@ pub(crate) fn validate_operation(
...
+    for operation in document.operations.iter() {
+        let op_kind = match operation.operation_type {
+            ast::OperationType::Query => continue,
+            ast::OperationType::Mutation => "mutation",
+            ast::OperationType::Subscription => "subscription",
+        };
+        let _ = forbid_defer_on_root(
+            document,
+            &operation.selection_set,
+            op_kind,
```
tninesling: I wonder if we could pass the `ast::OperationType` in here to avoid a stringly-typed API

crates/apollo-compiler/src/validation/operation.rs
```diff
@@ -174,6 +177,291 @@ pub(crate) fn validate_operation(
...
+    operation_type: &'static str,
+    diagnostics: &mut DiagnosticList,
+    visited_fragments: &mut HashSet<&'doc Name>,
+    mut guard: DepthGuard<'_>,
+) -> Result<(), RecursionLimitError> {
+    for selection in &selection_set.selections {
+        match selection {
+            // Spec ForbidDeferStream only checks @stream on Fields; @stream is not
+            // covered by this @defer-focused validation.
+            executable::Selection::Field(_) => {}
```
tninesling, line 324: This being empty was counterintuitive to me at first, since I don't deal with `@defer` much. It may be worth a note that the reason we don't do anything with `@defer` in the field case is that it's only applicable on inline fragments and fragment spreads.

crates/apollo-compiler/src/validation/operation.rs
```diff
@@ -174,6 +177,291 @@ pub(crate) fn validate_operation(
...
+    Ok(())
+}
+
+fn report_root_defer(
+    directives: &executable::DirectiveList,
+    operation_type: &'static str,
+    diagnostics: &mut DiagnosticList,
+) {
+    for directive in directives.iter() {
+        if directive.name == "defer" {
```
tninesling: Do we have any constants for the names of `@defer`, `@skip`, and `@includes`? I don't think it matters as much for built-ins, but I'm used to sharing a common name in the context of Federation directives that can be renamed.

tninesling (commented):
Overall, I think this looks good. I just left a few small suggestions, but nothing I consider blocking. I'd like to see if @lrlna or @goto-bus-stop have any feedback on whether there's more required for this type of PR as well as whether this kind of change needs to be opt-in.

tninesling (approved):

PR #1071, opened 2026-07-28
fix(compiler): validate that root operations use different types

tninesling:
Per the GraphQL spec [section on operation definitions](https://spec.graphql.org/October2021/#sel-EAHTRDLAACNA4pD): 

> The query, mutation, and subscription root types must all be different types if provided.

We had a comment claiming we validated this, when we really didn't. This now enforces the rule.


tninesling:
Coordinate with #1069 on diagnostic test data numbering

goto-bus-stop (approved):

PR #1072, opened 2026-07-28
fix(smith): choose different types for schema roots

tninesling:
Per the GraphQL spec [section on operation definitions](https://spec.graphql.org/October2021/#sel-EAHTRDLAACNA4pD): 

> The query, mutation, and subscription root types must all be different types if provided.

This ensures smith chooses a different type for the different schema roots. The corresponding validation is implemented in #1071.

### DIFF
--- a/crates/apollo-smith/src/schema.rs
+++ b/crates/apollo-smith/src/schema.rs
@@ -132,13 +132,11 @@ impl DocumentBuilder<'_> {
             .then(|| self.description())
             .transpose()?;
         let directives = self.directives(DirectiveLocation::Schema)?;
-        let named_types: Vec<Ty> = self
+        let mut named_types: Vec<Ty> = self
             .list_existing_object_types()
             .into_iter()
             .filter(|ty| ty.is_named() && !ty.is_builtin())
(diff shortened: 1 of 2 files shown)


goto-bus-stop (approved):

crates/apollo-smith/src/schema.rs
```diff
@@ -132,30 +132,37 @@ impl DocumentBuilder<'_> {
...
+            let m = self.u.choose(&named_types)?.clone();
+            named_types.retain(|ty| *ty != m);
+            Some(m)
+        } else {
+            None
+        };
+
+        let include_subscription = !named_types.is_empty() && self.u.arbitrary().unwrap_or(false);
+        let subscription = if include_subscription {
+            Some(self.u.choose(&named_types)?.clone())
```
TylerBloom, line 162: Shouldn't we remove this type like we are in the `mutation` case?

tninesling: This is the last one, so there's no reason to filter it out. We just need this one to be unique from the previous two.

PR #1074, opened 2026-07-29
Revert "fix(ci): generate macro in a way that works with nightly rustfmt"

tninesling:
Reverts apollographql/apollo-rs#1070

The underlying panic was resolved by https://github.com/rust-lang/rustfmt/pull/6990, so we can revert this once that is published in a new nightly.


goto-bus-stop (approved):

tninesling:
Note: waiting on the Rust folks to sync the rustfmt tree into the main repository; then it will go out in a nightly.

PR #1075, opened 2026-07-30
fix(smith): generate correctly shaped default values based on type being filled

tninesling:
Fixes a longstanding issue where default values were generated without knowledge of the type they needed to be.

[FED-1080]: https://apollographql.atlassian.net/browse/FED-1080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


goto-bus-stop (approved):
Not sure what our testing story is for apollo-smith if anything, is this something that could use a test?

PR #1080, opened 2026-08-05
feat!: September 2025 spec terminology, links, and built-in schema

tninesling:
Mechanical portion of the September 2025 GraphQL specification update. No new validation or execution behavior beyond the built-in schema tweaks below.

## BREAKING

- `executable::Field::response_key()` → `response_name()` (spec terminology: "response key" → "response name")
- `resolvers::FieldError` → `resolvers::ExecutionError` (spec terminology: "field error" → "execution error")
- CST `VariableDefinitions` → `VariablesDefinition` (spec grammar rename): `cst::VariablesDefinition`, `SyntaxKind::VARIABLES_DEFINITION`, `OperationDefinition::variables_definition()`

## Built-in schema

- `includeDeprecated` introspection arguments are now `Boolean! = false`
- `@deprecated` is declared with `reason: String!`, so explicit `reason: null` is no longer valid

## Documentation

- All spec links updated from October2021/draft to September2025

Most of the file count is one-line doc-link updates; the substantive changes are in `built_in_types.graphql`, the CST rename in apollo-parser, and the two type renames in apollo-compiler.

Part of a stack; two follow-up PRs cover Unicode escapes (apollo-parser) and the new validation/execution rules.


goto-bus-stop:
Before I look at this, what's the upgrade story for a customer like the Router? If there's supergraph schemas that contain `@deprecated(reason: null)` then we should continue supporting it.

tninesling:
> Before I look at this, what's the upgrade story for a customer like the Router? If there's supergraph schemas that contain `@deprecated(reason: null)` then we should continue supporting it.

If users have explicitly set `reason: null` in their subgraph, then it will carry forward into the supergraph schema. I don't know how many people are doing that, but it would be a fairly mechanical fix. We're going to have a similar conversation for other breaking changes like default value validation.

We'll have two different major versions of the package once we release v2. Should router just import them both and allow the user to toggle between them? That could be a simple escape hatch for users who don't want `oneOf` and need time to update their graphs.

goto-bus-stop:
I think it makes sense to have a Router configuration along the lines of `graphql_spec: 2025`. I think using multiple apollo-compiler versions would be incredibly invasive, since the types are widely used and would be different between the versions. Or we'd have to convert an already-parsed tree from v1 to v2.

I think it may be simpler if apollo-compiler validation took a spec version parameter.

tninesling:
> I think it may be simpler if apollo-compiler validation took a spec version parameter.

I don't really like the idea of supporting multiple GQL spec versions in the same library. I think we can provide opt-outs for things we think will be very onerous to fix, but I think it's cleaner for us to be able to say "v2 uses the 2025 spec".

With that said, our corpus of test schemas surfaced 3 graphs which have this `@deprecated(reason: null)` pattern out of many thousands. I think a better path forward is to resolve this during composition. We already have a notion of "upgrading" a schema, which currently goes from fed1 to fed2 subgraph formats. I'd like to add a fed3 upgrader which can remove `reason: null`. Then our resolution can be that customers recompose with the latest version, and it'll turn this into a hint.

I expect default value validation to be a tougher hurdle to get over, so I've provided an opt-out mechanism in my PR for that one.

TylerBloom:
I agree with Taylor here. The fix on the part of the schema authors is either just deleting the `null` (so they use the default argument) or replacing it with an empty string. "v2 uses the 2025 spec" should be our goal.

goto-bus-stop:
As long as we have a feasible, backwards-compatible path to adopt this in the Router I'm okay with it.

If it's part of composition, then users would have to recompose, I think? Probably fine if we can adopt this in Router 3.0.0 but otherwise we'd have to have a different solution

TylerBloom (approved):

TylerBloom:
> Probably fine if we can adopt this in Router 3.0.0

I believe that is the plan. Router 2.X would stay on apollo-compiler v1.X, and router 3 would move the v2

tninesling:
> > Probably fine if we can adopt this in Router 3.0.0
> 
> I believe that is the plan. Router 2.X would stay on apollo-compiler v1.X, and router 3 would move the v2

Yes, I am only planning to upgrade to v2 in Router 3. That release is coming up fairly soon though, so I'd like to get this out a bit ahead of it. We'll add some small fixup logic to composition for the breaking changes to `@_deprecated` to smooth the transition for the few folks who would break on that. Then, we'll have Federation v3 enforce the rules faithfully.

PR #1084, opened 2026-08-12
feat(compiler): parse FieldSet from a source span, with escape-aware offsets

tninesling:
# Re-land #1065 with correct escape sequence handling

Stacked on #1083 (apollo-parser half). This re-lands #1065 (reverted in #1067) plus the fix for the escape-sequence problems @goto-bus-stop raised there:

1. **Correctness**: the at-span parser previously parsed the *source form* of the string literal, so `@key(fields: "a\nb")` — accepted by `FieldSet::parse_and_validate` since escapes are decoded at AST build — produced a spurious syntax error.
2. **Panic vector**: naively unescaping before parsing while keeping a flat base offset shifts every diagnostic offset after an escape (`\u0020` is 6 source bytes but 1 parsed byte), which can land spans in the middle of a multi-byte UTF-8 character and panic when rendering diagnostics (same class as #1023).

## How it works

`"..."` literals are unescaped with `unescape_string_with_offsets()` (from #1083) before parsing. Its `(unescaped_offset, source_offset)` checkpoints ride along in the internal `SourceContext`, and every AST node location and parse error is mapped from parsed-text offsets back to source offsets via binary search over the checkpoints + 1:1 advance. Since bytes between checkpoints are identical to the source, mapped offsets always land on character boundaries — the panic is ruled out structurally rather than per-escape-kind.

Block string (`"""..."""`) content is still parsed as written: block strings do not process escape sequences per the spec, and their indentation is irrelevant to whitespace-insensitive field sets. This is documented on `FieldSet::parse_and_validate_at_span()`.

## Test plan

- Regression test for the panic scenario: unicode escape followed by multi-byte characters (`"id\u0020한글"`), asserting diagnostics point exactly at the offending characters.
- Parity tests: escaped selections (`"id\naltId"`, `"id\u0020altId"`) accepted by both `parse_and_validate` and `parse_and_validate_at_span`; block strings containing `\n` rejected by both.
- Error-rendering snapshot showing diagnostics past an escape pointing at the right source columns.
- All 375 workspace tests pass.

### DIFF
--- a/crates/apollo-compiler/src/ast/from_cst.rs
+++ b/crates/apollo-compiler/src/ast/from_cst.rs
@@ -10,48 +10,88 @@ use apollo_parser::SyntaxNode;
 use apollo_parser::S;
 
 #[derive(Copy, Clone)]
-pub(crate) struct SourceContext {
+pub(crate) struct SourceContext<'a> {
     file_id: FileId,
     base_offset: rowan::TextSize,
+    /// `(unescaped_offset, source_offset)` checkpoints recorded after each
+    /// escape sequence when the parsed text was unescaped from a string
+    /// literal in the source. Empty when parsed text matches the source 1:1.
+    /// Both offsets are relative: to the parsed text and to `base_offset`
+    /// respectively.
+    escape_checkpoints: &'a [(rowan::TextSize, rowan::TextSize)],
 }
 
-impl SourceContext {
+impl<'a> SourceContext<'a> {
     pub(crate) fn new(file_id: FileId) -> Self {
-        Self {
-            file_id,
-            base_offset: 0.into(),
-        }
+        Self::with_offset(file_id, 0.into())
     }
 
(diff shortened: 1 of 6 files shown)


goto-bus-stop:
It would also be great to have a fuzz test for the new method, to make sure that parsing a field set from an arbitrary string doesn't panic. Doesn't have to be in this PR tho.

> Block string ("""...""") content is still parsed as written: block strings do not process escape sequences per the spec, and their indentation is irrelevant to whitespace-insensitive field sets. This is documented on FieldSet::parse_and_validate_at_span().

Note block strings support exactly one kind of escape sequence: `\"""` to insert a literal triple-quote as the content of the string. This would cause the same problems as `\n` does in regular strings, either a spurious parse error or a panic. Scroll down a bit from https://spec.graphql.org/September2025/#sec-String-Value.Static-Semantics to find the item `BlockStringCharacter :: \"""`.

PR #1087, opened 2026-08-12
feat!: September 2025 validation and execution rules

tninesling:
## Summary

Implements the remaining normative changes from the September 2025 GraphQL specification:

- **Required field deprecation:** required (non-null, no default) arguments and input object fields must not be `@deprecated`
- **IsValidImplementation:** an implementing field must not be deprecated when the interface field it implements is not deprecated
- **Execution coercion:** default values are coerced (applying nested defaults) instead of passed through verbatim
- **Execution serialization:** numeric ID results serialize as String

**Stack:** #1080 → #1086 → **this PR**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1089, opened 2026-08-13
fix(compiler): align resolver execution and input coercion with the GraphQL spec

duckki:
## Summary

Fixes six spec conformance issues in `apollo_compiler::resolvers` execution and input
coercion. Expected behaviors follow the September 2025 Edition and were verified
against graphql-js v17.

[Detailed walkthrough](https://gist.github.com/duckki/6379d510226bf2f6913596bbb7c5884d)

## Fixes

- An input object field whose value is a variable with no runtime value now uses the
  field's default value, or is omitted — previously it became explicit `null`.
- A resolver error for a list item with a nullable item type now nullifies only that
  item (`[1, null, 2]`), not the whole list.
- Variables nested inside custom scalar literals (`arg: {id: $var}`) are now replaced
  with their runtime values instead of raising a spurious execution error.
- `@skip` / `@include` with an `if` value of `null` now raises an execution error,
  like graphql-js, instead of being silently ignored.
- Float input coercion now accepts integer values up to and including 2⁵³ − 1
  (off-by-one at the boundary).

Stacked on #1087, which independently covers numeric-`ID`-as-string result coercion —
this PR keeps a regression test for it.

All changes are response-observable and move outputs toward graphql-js parity. Also
documents the existing sibling-cancellation behavior on non-null error propagation
(unchanged, matches graphql-js) in the module docs.

## Tests

New `crates/apollo-compiler/tests/resolvers.rs`: one reproducer per fix plus
regression pins for the cancellation semantics; expected outputs cross-checked
against graphql-js v17.0.1.

### DIFF
--- a/crates/apollo-compiler/CHANGELOG.md
+++ b/crates/apollo-compiler/CHANGELOG.md
@@ -120,12 +120,45 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   `@defer` is not a built-in directive in apollo-compiler, so a schema using it
   must declare the directive. `@stream` is out of scope.
 
+## Fixes
+
+- **Align resolver-based execution and input coercion with the GraphQL spec - [duckki], [pull/1089]**
+
+  Fixes several spec conformance issues in `apollo_compiler::resolvers`
+  execution, found by auditing it against the GraphQL specification and the
+  graphql-js reference implementation:
+
+  - An input object field whose value is a variable with no runtime value now
+    uses the field’s default value, or is omitted when it has no default.
+    Previously the field was set to explicit `null`, which is semantically
+    different and skipped defaults.
+  - A resolver error for a list item with a nullable item type now nullifies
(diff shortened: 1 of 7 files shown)


tninesling:
Thanks @duckki! I'm trying to have all 2025 spec compatibility changes target the `v2` branch. Would you be able to retarget there, or preferably stack on top of #1087 ? I'm hoping to have the v1 -> v2 upgrade be a clean swap from the 2021 spec to 2025.

duckki:
> Thanks @duckki! I'm trying to have all 2025 spec compatibility changes target the `v2` branch. Would you be able to retarget there, or preferably stack on top of #1087 ? I'm hoping to have the v1 -> v2 upgrade be a clean swap from the 2021 spec to 2025.

Done — rebased onto #1087 and retargeted the base to tninesling/spec-validation-rules.
Also dropped my numeric-ID fix in favor of yours and kept a regression test for it.

tninesling (approved):

PR #1091, opened 2026-08-21
refactor!: replace ComponentName with Node<Name>

tninesling:
## Summary
- Replaces `ComponentName` wrapper with `Node<Name>`, using Node's built-in extension tracking
- `Name::to_component()` → `Name::to_node()`, taking `Option<ExtensionId>` instead of `ComponentOrigin`
- Updates `extensions()` methods to use `.flatten()` on the new `Option<&ExtensionId>` iterators

## Test plan
- [x] All 307 tests pass
- [x] No clippy warnings
- [x] Formatted with nightly rustfmt

Part 2 of a stacked PR series (#1090 → #1091 → #1092)

### DIFF
--- a/crates/apollo-compiler/src/name.rs
+++ b/crates/apollo-compiler/src/name.rs
@@ -1,12 +1,12 @@
 use crate::diagnostic::CliReport;
 use crate::diagnostic::ToCliReport;
+use crate::node::ExtensionId;
 use crate::parser::FileId;
 use crate::parser::LineColumn;
 use crate::parser::SourceMap;
 use crate::parser::SourceSpan;
 use crate::parser::TaggedFileId;
-use crate::schema::ComponentName;
-use crate::schema::ComponentOrigin;
+use crate::Node;
 use rowan::TextRange;
 use std::fmt;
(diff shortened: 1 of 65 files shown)


crates/apollo-compiler/src/name.rs
```diff
@@ -383,6 +386,30 @@ impl PartialOrd for Name {
...
+    fn eq(&self, other: &str) -> bool {
+        self.as_str() == other
+    }
+}
+
+impl PartialEq<&'_ str> for Node<Name> {
+    fn eq(&self, other: &&str) -> bool {
+        self.as_str() == *other
+    }
+}
```
TylerBloom: We can generalize this. We can do the same for `Name`.
```suggestion
impl PartialEq<str> for Node<Name> {
    fn eq(&self, other: &str) -> bool {
        self.as_str() == other
    }
}

impl<T: AsRef<str>> PartialEq<T> for Node<Name> {
    fn eq(&self, other: &T) -> bool {
        self.as_str() == other.as_ref()
    }
}
```

PR #1107, opened 2026-09-03
test(compiler): deferred fragment spread multiple times is considered unique

tninesling:
Adds an `ok/` validation fixture confirming that the "Defer And Stream Directive Labels Are Unique" rule ([graphql-spec#1110](https://github.com/graphql/graphql-spec/pull/1110)) is evaluated over the document as written: a labeled `@defer` inside a fragment definition is a single directive, so spreading that fragment in multiple places must pass validation.


tninesling:
This is a follow-up test for the validation added in #1069

sachindshinde (approved):

PR #1111, opened 2026-09-17
Support for the 2026 Spec

TylerBloom:
Implements the only substantive change in the [September 2026 changelog](https://github.com/graphql/graphql-spec/blob/sept-2026-release/changelogs/September2026.md): directives may be applied to directive definitions, and directive definitions may be extended ([graphql-spec#1206](https://github.com/graphql/graphql-spec/pull/1206)). The other five changelog entries are editorial or already satisfied.

```graphql
directive @internal on DIRECTIVE_DEFINITION

directive @example @deprecated(reason: "use @other instead") on FIELD

extend directive @example @internal
```

## BREAKING

- `ast::DirectiveDefinition` gains a `directives: DirectiveList` field — struct literals must add it
- `ast::DirectiveLocation` gains `DirectiveDefinition`, and `ast::Definition` gains `DirectiveExtension(Node<DirectiveExtension>)` — exhaustive matches must be updated
- `cst::Definition` gains a `DirectiveExtension` variant — likewise

None of these enums are `#[non_exhaustive]`.

## apollo-parser

- `DirectiveDefinition` accepts `Directives[Const]?` between the arguments definition and `repeatable`, per the amended grammar; `cst::DirectiveDefinition::directives()` exposes them
- `DIRECTIVE_DEFINITION` is accepted as a directive location (`SyntaxKind::DIRECTIVE_DEFINITION_KW`)
- New `DirectiveExtension` production and CST node (`SyntaxKind::DIRECTIVE_EXTENSION`), dispatched from `extend`

## apollo-compiler

- Directives on a directive definition are validated against the `DIRECTIVE_DEFINITION` location and participate in the recursive-directive-definition cycle check
- A directive extension merges its directives into the matching definition, tracking origin through `Node`'s `ExtensionId` the same way type extensions do, and may appear before the definition it extends
- Extending a never-defined directive is a new build error. Unlike other extension kinds this is an error even under `adopt_orphan_extensions()`: a directive definition's locations are mandatory, so there is no empty-but-valid definition to adopt into — one could never be serialized back to parseable SDL
- Fixes `ast::Definition::directives()`, which previously returned an empty list for directive definitions

## Built-in schema

Per Appendix D:

- `__Directive` gains `isDeprecated: Boolean!` and `deprecationReason: String`
- `__Schema.directives` gains `includeDeprecated: Boolean! = false`, and now excludes deprecated directives unless it is set
- `__DirectiveLocation` gains `DIRECTIVE_DEFINITION`
- `@deprecated` is declared on `DIRECTIVE_DEFINITION`

## Not included

- apollo-smith generates neither construct and skips `extend directive` on input; its own `DirectiveLocation` still lacks `DIRECTIVE_DEFINITION` and will hit `unreachable!()` on a schema using it
- `Schema::to_ast()` flattens extension-sourced directives into the definition rather than re-emitting `extend directive` blocks, unlike other extension kinds. Output is valid and revalidates; only round-trip structure differs
- Spec doc links still point at September2025, pending the mass update once the spec is released

Much of the diff is tests, regenerated snapshots (line-number shifts from the additions to `built_in_types.graphql`), and generated CST code; the substantive changes are in `graphql.ungram`, `grammar/directive.rs`, `ast/`, and `schema/from_ast.rs`.


crates/apollo-compiler/src/ast/mod.rs
```diff
@@ -168,6 +171,7 @@ impl Hash for DirectiveDefinition {
         hash_unordered(self.arguments.iter(), state, self.arguments.len());
         self.repeatable.hash(state);
         hash_unordered(self.locations.iter(), state, self.locations.len());
+        hash_unordered(self.directives.iter(), state, self.directives.len());
```
tninesling, line 174: Unfortunately directive order can matter, so I think this should be a regular ordered hash.

crates/apollo-compiler/src/schema/from_ast.rs
```diff
@@ -76,6 +78,10 @@ impl SchemaBuilder {
     /// Configure the builder so that “orphan” schema extensions and type extensions
     /// (without a corresponding definition) are “adopted”:
     /// accepted as if extending an empty definition instead of being rejected as errors.
+    ///
+    /// This does not apply to directive extensions: a directive definition's locations are
```
tninesling, line 82: Is this saying that the directive definition case can never happen, but then we have code below to handle that particular case anyway?

tninesling (changes requested):
I think we're missing a round-trip serialization test to make sure parsing and serializing maintains the `extend` keyword on the definition. The `test_schema_reserialize` test in `crates/apollo-compiler/tests/schema` might be a good spot for that.

crates/apollo-compiler/src/ast/mod.rs
TylerBloom: That's a good point. Looking at it a more closely, it was always inconsistent with `DirectiveList`'s `PartialEq` `impl`, which assumed ordering mattered. Addressed in e366dd93.

crates/apollo-compiler/src/schema/serialize.rs
```diff
@@ -106,6 +106,28 @@ impl Node<SchemaDefinition> {
     }
 }
 
+impl Node<DirectiveDefinition> {
+    fn to_ast(&self) -> impl Iterator<Item = ast::Definition> + use<'_> {
```
tninesling, line 110: This is neat! I haven't seen the `use<'_>` syntax before

tninesling (approved):

crates/apollo-compiler/src/schema/serialize.rs
TylerBloom: I think it was introduced right before the 2024 edition, where they fixed the semantics of `-> impl Trait + 'a`. Its to get around the weird `trait Capture<'a>` thing you used to have to do.

